PR #1, opened 2019-02-07
Add lineage link

danwom:
- Added a lineage section in the table details side bar. This link can be enabled and customized by overwriting the default config values.
- Cleaned up some global styles by nesting them inside their component classes.
- Removed deprecated 'nav_link' file.
- Added comments for static configs.


amundsen_application/static/js/components/common/Feedback/styles.scss
```diff
@@ -5,67 +5,58 @@
...
+  }
+
+  .feedback-header {
+    margin-bottom: 8px;
+  }
+
+  .btn-group {
+    margin: 8px auto 16px;
+  }
 }
```
danwom: These css rules were being applied to all images so I nested the rules and deleted a few that were unused.

danwom:
👀 @ttannis @feng-tao

feng-tao:
Given we are open source now, I would prefer our PR description is more descriptive as not everyone has the same context on the pr.

And I wonder how external people could use the feature. Does it require to have a lineage service which visualizes the data somewhere else? cc @jinhyukchang . I wonder whether Apache Atlas supported this feature.

And have we decided not using the column linage graph?

feng-tao:
It seems that you are removing some existing nav_links which I think we should mention in the pr description as well.

amundsen_application/static/config/config-default.ts
```diff
@@ -15,6 +15,13 @@ const configDefault: AppConfig = {
     key: 'default-key',
     sampleRate: 100,
   },
+  lineage: {
```
feng-tao: it would be good if we could add a comment for what the config is used for. In fact, it would be good to have a comment for all the config here(could be a different pr)

danwom: I'll add some comments to the type definitions in config.types.ts.

PR #2, opened 2019-02-08
Added Travis-CI configuration file

danwom:
Travis will execute pre-commit checks by building and testing our code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ matrix:
         - pip3 install -r requirements3.txt
       script:
         - make test
-        - make lint
 
     - language: node_js
       node_js:
@@ -18,5 +17,8 @@ matrix:
         - npm install
       script:
         - npm run build
-#        TODO - Fix and enable tests
+#        TODO - replace build with `tsc` and 'lint`
+#        - npm run tsc
+#        - npm run lint
+#        TODO - Fix and enable 'test'
 #        - npm run test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+matrix:
+  include:
+    - language: python
+      python:
+      - "3.6"
+      install:
+        - pip3 install -r requirements3.txt
+      script:
+        - make test
+        - make lint
+        - make mypy
+
+    - language: node_js
+      node_js:
+        - "8"
+      before_install:
+        - cd amundsen_application/static
+      install:
+        - npm install
+      script:
+        - npm run build
+        - npm run test
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ matrix:
       script:
         - make test
         - make lint
-        - make mypy
 
     - language: node_js
       node_js:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,22 @@
...
+    - language: node_js
+      node_js:
+        - "8"
+      before_install:
+        - cd amundsen_application/static
+      install:
+        - npm install
+      script:
+        - npm run build
+        - npm run test
```
danwom: As-is, the jest tests will fail because the snapshots are not in sync with the currently build code.

ttannis: Because our unit tests are still TBD, `npm run test` will fail. In addition, `npm run build` builds our source so I'm wondering if we want to run that on Travis. In the event we don't run `npm run build`, we'd instead want to run `npm run tsc` for the type checks. However for some reason `npm run tsc` fails locally so I assume it would fail on Travis as well -- I have a ticket somewhere to investigate.

What do you think of only configuring `npm run lint` for now? Later we can add `npm run test` when unit tests are completed/fixed.

.travis.yml
```diff
@@ -0,0 +1,22 @@
...
+  include:
+    - language: python
+      python:
+      - "3.6"
+      install:
+        - pip3 install -r requirements3.txt
+      script:
+        - make test
+        - make lint
+        - make mypy
```
jinhyukchang: I don't think "amundsenfrontendlibrary" Make file has mypy target. You may need to add entry like this.
https://github.com/lyft/amundsensearchlibrary/blob/master/Makefile#L16

.travis.yml
jinhyukchang: I think we need to either fix the test or disable this as it will block the deployment step that will be added shortly.

danwom:
https://travis-ci.com/lyft/amundsenfrontendlibrary/builds/100291238

.travis.yml
```diff
@@ -0,0 +1,22 @@
...
+  include:
+    - language: python
+      python:
+      - "3.6"
+      install:
+        - pip3 install -r requirements3.txt
+      script:
+        - make test
+        - make lint
+
```
ttannis: `make test` also runs the linter, so we don't need `make lint` here. We also have to add mypy checks:
https://github.com/lyft/amundsensearchlibrary/pull/4

.travis.yml
```diff
@@ -0,0 +1,22 @@
...
+    - language: node_js
+      node_js:
+        - "8"
+      before_install:
+        - cd amundsen_application/static
+      install:
+        - npm install
+      script:
+        - npm run build
```
ttannis: I'm still not sure about running `npm run build` which is a script written to run tests and build the source, as opposed to explicitly running the scripts we have for explicitly running tests. Thoughts?

danwom: As far as I can tell we will some combination of `tsc, lint, and test`. `build` is a fairly nice placeholder until we can fix `test` and `tsc`.

ttannis: Resolving, we will modify after fixing our `npm run tsc` script.

PR #9, opened 2019-02-12
Add support for a logo in the NavBar

ttannis:
This PR adds support for customizable logo in the `NavBar`.  

**Summary of Changes**
1. A `logoPath` option was added to the application configuration, which is `null` by default until we have the design resources for an official Amundsen logo. If a `logoPath` exists, it will render it on the left hand side of the `NavBar`. We do not have any checks/fallbacks for a bad path -- this implementation expects that if a `logoPath` is configured, it is valid.

2. CSS has also been updated accordingly to space the items in `.nav-bar-left`  in the same way they are spaced in `.nav-bar-right`

### DIFF
--- a/amundsen_application/static/config/config-default.ts
+++ b/amundsen_application/static/config/config-default.ts
@@ -22,6 +22,7 @@ const configDefault: AppConfig = {
     },
     iconPath: 'PATH_TO_ICON'
   },
+  logoPath: null,
   navLinks: [
     {
       label: "Announcements",
--- a/amundsen_application/static/js/components/NavBar/index.tsx
+++ b/amundsen_application/static/js/components/NavBar/index.tsx
@@ -53,7 +53,7 @@ class NavBar extends React.Component<NavBarProps, NavBarState> {
             <div className="nav-bar-left">
               {
                 AppConfig.logoPath &&
-                <img src={AppConfig.logoPath} />
+                <img className="logo-icon" src={AppConfig.logoPath} />
               }
               <Link to={`/`}>
                 AMUNDSEN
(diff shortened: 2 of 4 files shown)


amundsen_application/static/config/config-default.ts
```diff
@@ -22,6 +22,7 @@ const configDefault: AppConfig = {
     },
     iconPath: 'PATH_TO_ICON'
   },
+  logoPath: null,
```
feng-tao: It would be good if we could comment about the configs for external users(e.g example value, what it is used for)?

ttannis: We currently mention it in our README: https://github.com/lyft/amundsenfrontendlibrary#application-config, and we also put some lengthier descriptions in this file for the less straightforward cases: https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/config/config.types.ts

amundsen_application/static/js/components/NavBar/index.tsx
```diff
@@ -51,6 +51,10 @@ class NavBar extends React.Component<NavBarProps, NavBarState> {
         <div className="row">
           <div className="nav-bar">
             <div className="nav-bar-left">
+              {
+                AppConfig.logoPath &&
+                <img src={AppConfig.logoPath} />
```
danwom: Might be nice to give this a class like "logo-icon" and give it some max width / height.

ttannis: No problem. Will add something that makes sense for now and we can tweak it later as our NavBar evolves.
```
.logo-icon {
  max-height: 48px;
  max-width: 144px;
}
```

PR #10, opened 2019-02-12
Changed editable-text interaction

danwom:
- Use "edit" link instead of icon
- "edit" is visible for empty text values and on hover
- textarea appears flush with blackground
-- textarea gets light gradient on de-focus

### DIFF
--- a/amundsen_application/static/css/_popovers.scss
+++ b/amundsen_application/static/css/_popovers.scss
@@ -31,7 +31,6 @@
   border-left-color: $gray-darker;
 }
 
-
 .tooltip-inner {
   padding: 0px;
   border-radius: 3px;
--- a/amundsen_application/static/css/_popovers.scss
+++ b/amundsen_application/static/css/_popovers.scss
@@ -30,3 +30,38 @@
 .popover.left .arrow:after {
   border-left-color: $gray-darker;
 }
+
+
+.tooltip-inner {
+  padding: 0px;
+  border-radius: 3px;
+  background-color: $gray-darker;
+}
+
+.tooltip-inner button {
+  height: 36px;
+  width: 96px;
+  font-size: 14px;
+  border-radius: 3px;
+  border: none;
+  background-color: $gray-darker;
+  color: $body-bg;
+  font-family: $font-family-sans-serif-bold;
+  font-weight: $font-weight-sans-serif-bold;
+  outline: none;
+}
+.tooltip-inner button:hover {
+  color: $gray-light;
+}
+
+.error-tooltip {
+    display: flex;
+    padding: 5px;
+}
+.error-tooltip button {
+  height: 24px;
+  width: 24px;
+  font-size: 16px;
+  margin: auto;
+  font-family: 'Glyphicons Halflings';
+}
--- a/amundsen_application/static/js/components/common/EditableText/index.tsx
+++ b/amundsen_application/static/js/components/common/EditableText/index.tsx
@@ -32,7 +32,7 @@ interface EditableTextState {
 
 class EditableText extends React.Component<EditableTextProps, EditableTextState> {
   private textAreaTarget: HTMLTextAreaElement;
-  private editButtonTarget: HTMLButtonElement;
+  private editAnchorTarget: HTMLAnchorElement;
 
   public static defaultProps: EditableTextProps = {
     editable: true,
@@ -77,7 +77,7 @@ class EditableText extends React.Component<EditableTextProps, EditableTextState>
 
   exitEditMode = () => {
     this.setState({ isDisabled: false, inEditMode: false, refreshValue: '' });
-  }
+  };
 
   enterEditMode = () => {
     if (this.props.getLatestValue) {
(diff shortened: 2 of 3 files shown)


amundsen_application/static/js/components/common/EditableText/index.tsx
```diff
@@ -136,9 +127,17 @@ class EditableText extends React.Component<EditableTextProps, EditableTextState>
                   </div>
                 </Tooltip>
               </Overlay>
-            </div>
           }
-          <div className='editable-text'>{ this.state.value }</div>
+          <div className={"editable-text " + (this.state.value ? "" : "no-value")}>
+            { this.state.value }
+            <a className={ "edit-link " + (this.state.isDisabled ? "hidden" : "") }
```
ttannis: For what reason do we want to have the edit target rendered here? It looks like the edit target will end up getting rendered regardless of whether or not the component is editable.

nit: Visually, does the anchor tag happen to take on some default `visited` style as well? (which i think would be undesirable)

amundsen_application/static/css/_popovers.scss
```diff
@@ -30,3 +30,38 @@
 .popover.left .arrow:after {
   border-left-color: $gray-darker;
 }
+
+
+.tooltip-inner {
```
ttannis: I'm concerned about how this will change every other tooltip in the application, and would rather keep this in `EditableText` unless we know this would be acceptable for Design, so that we don't create more work for ourselves having to go back in and fix other tooltips. 

Since we're meeting first thing tomorrow to talk about these exact tickets, can we run this by them?

danwom: The Tooltip component is only used by the EditableText component and the original styles were already global. I did consider creating a separate `_tooltips.scss` file but given how they are so closely related I figured I could put them in the same file.

I also have the idea of using bootstrap variables to configure tooltips and popovers and deleting most of these rules, but that's out of scope for this task.

amundsen_application/static/js/components/common/EditableText/index.tsx
danwom: The current idea is to put the edit link inline with the description and hide/show on hover using opacity so it remains in the DOM flow. When editing is disabled, the `hidden` class will use `display: none` and remove this element from DOM flow. Alternatively we could just not render this link altogether and there might be a very minor performance improvement.

As for `:visited` I believe this uses browser history and the `href` attribute to determine if an anchor is visited or not. Since the href is `JavaScript:void(0)` I believe we can ignore this case.

amundsen_application/static/css/_popovers.scss
ttannis: That's new to me! I had been under the impression `Popover` was an extension of `Tooltip` and therefore shared classes.

danwom: `Popover` does extend `Tooltip` but the classes are different.

amundsen_application/static/js/components/common/EditableText/index.tsx
```diff
@@ -136,9 +127,21 @@ class EditableText extends React.Component<EditableTextProps, EditableTextState>
...
+            { this.state.value }
+            <a className={ "edit-link " + (this.state.value ? "" : "no-value") }
+               href="JavaScript:void(0)"
+               onClick={ this.enterEditMode }
+               ref={ anchor => {
+                  this.editAnchorTarget = anchor;
+                }}
+            >
+              {
+                this.state.value ? "edit" : "Add Description"
```
ttannis: nit: Is the word "edit" also supposed to be capitalized?

amundsen_application/static/js/components/common/EditableText/index.tsx
```diff
@@ -136,9 +127,21 @@ class EditableText extends React.Component<EditableTextProps, EditableTextState>
                   </div>
                 </Tooltip>
               </Overlay>
-            </div>
           }
-          <div className='editable-text'>{ this.state.value }</div>
+          <div className={"editable-text"}>
+            { this.state.value }
+            <a className={ "edit-link " + (this.state.value ? "" : "no-value") }
```
ttannis: How does this get hidden when `this.state.editable` is `false`? You had mentioned a `hidden` class would be applied.

amundsen_application/static/js/components/common/EditableText/index.tsx
danwom: Personally I would like to use consistent capitalization, but I brought this up with Matt and he specifically asked for this behavior.

amundsen_application/static/js/components/common/EditableText/index.tsx
danwom: Originally I was using the `hidden` class but encountered a problem with the empty state since `display: none` takes elements out of the DOM flow it became impossible to hover over the div. This problem went away when we decided to have "Add Description" always show, but I had already switched to using `opacity: 0` and I think that works somewhat better since it doesn't affect the flow.

danwom: I misread the question. I need to cover this case.

PR #13, opened 2019-02-14
WIP fix for footer covering browser search result

danwom:


amundsen_application/static/js/components/common/EditableText/index.tsx
```diff
@@ -162,7 +162,7 @@ class EditableText extends React.Component<EditableTextProps, EditableTextState>
           show={true}
           target={this.getTarget.bind(this,'textArea')}
         >
-          <Tooltip>
+          <Tooltip container={this}>
```
danwom: I didn't get this quite working in the 5 minutes I played with it, but this is how we can fix the floating tooltips issue.

danwom:
@ttannis This is the WIP for footer fix.

amundsen_application/static/js/components/common/EditableText/index.tsx
danwom: `container` should actually go on the Overlay component.

danwom:
Honestly I don't think this issue is worth fixing since it forces us to attach every tooltip/popover to the element instead of the body. I can foresee this causing bigger issues down the road than hiding elements behind the footer for browser ctrl+f.

danwom:
I've sunk some more time into this and I was able to fix the `<Tooltip>`s with `<Overlay>` but not the `<Popover>`s with `<OverlayTrigger>`. The `<OverlayTrigger>` doesn't support a `container` prop.


I think we should move away from the `position: fixed` footer. The standard "sticky footer" sticks to the bottom of the browser when there is little content, but never blocks content. 

https://philipwalton.github.io/solved-by-flexbox/demos/sticky-footer/
(try "Toggle Contents" button)

https://getbootstrap.com/docs/3.3/examples/sticky-footer/

ttannis:
@danwom Kudos to how deeply you explored this. I'm on board with closing our internal ticket due to the fact that the workaround for the issue that would get introduced by the fix won't work for `OverlayTrigger`, which we  use so frequently.

PR #15, opened 2019-02-15
Renamed configs for the table profile and table lineage sections

danwom:
- Added 'isBeta' configs which adds a "(BETA)" label in each section

### DIFF
--- a/amundsen_application/static/config/config-default.ts
+++ b/amundsen_application/static/config/config-default.ts
@@ -5,22 +5,9 @@ const configDefault: AppConfig = {
     curatedTags: [],
     showAllTags: true,
   },
-  exploreSql: {
-    enabled: false,
-    generateUrl: (database: string, cluster: string, schema: string, table: string, partitionKey?: string, partitionValue?: string) => {
-      return `https://DEFAULT_EXPLORE_URL?schema=${schema}&cluster=${cluster}&db=${database}&table=${table}`;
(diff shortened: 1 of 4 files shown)


amundsen_application/static/config/config-default.ts
```diff
@@ -34,7 +21,22 @@ const configDefault: AppConfig = {
...
+      return `https://DEFAULT_LINEAGE_URL?schema=${schema}&cluster=${cluster}&db=${database}&table=${table}`;
+    },
+  },
+  tableProfile: {
+    isBeta: false,
+    isExploreEnabled: false,
+    exploreUrlGenerator: (database: string, cluster: string, schema: string, table: string, partitionKey?: string, partitionValue?: string) => {
+      return `https://DEFAULT_EXPLORE_URL?schema=${schema}&cluster=${cluster}&db=${database}&table=${table}`;
+    }
+  },
```
danwom: Any configs related to the Data Preview button should also fall under `tableProfile` if and when we add them.

feng-tao: the only concern is that external user may have confused what it means by isBeta(not sure if everyone ring a bell for beta, GA term). Otherwise LGTM

danwom: There are some comments about these configs in config.types.ts.

PR #16, opened 2019-02-20
Preview feature fixes

ttannis:
This PR fixes two issues with the state management and action dispatching of the `DataPreviewButton`.

**Summary of Changes**
1. The `preview` state is now associated with the `tableMetadata` state and reducer. This is a more appropriate place for it since updates to the `preview` state are tied to dispatch of `GetTableData` actions. i.e. - clearing the state when visiting a new `TableDetail` page.
2. The `TableDetail` page will also dispatch `GetPreviewData` actions. This is because child components mount before their parents and as a result the `DataPreviewButton` was dispatching two `GetPreviewData` actions when visiting a new `TableDetail` page -- once before it's parent mounted with the old table parameters, and once after with the new and correct table parameters. 

**Future Considerations**
The `tableMetadata` reducer could start using the nested reducers pattern for the preview and other pieces of state like column descriptions, but I did not want this PR for a bug fix to increase in scope or introduce a new pattern.


amundsen_application/static/js/components/TableDetail/DataPreviewButton/index.tsx
```diff
@@ -22,15 +22,11 @@ export interface StateFromProps {
   status: LoadingStatus;
 }
 
-export interface DispatchFromProps {
-  getPreviewData: (queryParams: PreviewQueryParams) => void;
-}
-
 export interface ComponentProps {
   queryParams: PreviewQueryParams;
```
danwom: Now that the `DataPreviewButton` no longer makes the request to fetch data, `queryParams` is only used to generate the modal title. It might make sense to pass in `"title"` or `"modalTitle"` instead.

amundsen_application/static/js/ducks/api/metadata/v0.ts
```diff
@@ -140,9 +142,22 @@ export function metadataUpdateColumnDescription(description, columnIndex, tableD
...
 }
 
-
 export function metadataGetLastIndexed() {
   return axios.get(`${API_PATH}/get_last_indexed`).then((response) => {
     return response.data.timestamp;
   });
 }
+
+export function metadataGetPreviewData(action: GetPreviewDataRequest) {
```
danwom: I've noticed the parameters are sometimes reducer actions and other times they are picked out from picked out of the reducer action.
E.g.
```
export function metadataUpdateColumnDescription(description, columnIndex, tableData) { ... }
yield call(metadataUpdateColumnDescription, action.newValue, action.columnIndex, state.tableMetadata.tableData);
```
vs 
```
export function metadataUpdateColumnDescription(action: updateColumnDescriptionRequest) { ... }
yield call(metadataUpdateColumnDescription, action);
```

Functionally these are equivalent but we should pick a pattern and try to be consistent. I have a slight preference on the latter but could go either way. Thoughts?

amundsen_application/static/js/components/TableDetail/DataPreviewButton/index.tsx
ttannis: Agreed, will pass a `modalTitle`.

amundsen_application/static/js/ducks/api/metadata/v0.ts
ttannis: Agreed. I actually have a preference right now for the first option because the second option assumes a one to one mapping between actions and these helper methods, which is currently the case but I don't know if it will always be. But perhaps it's also a better pattern to keep the one to one mapping. Can you file an internal ticket in our ever-growing improvements epic? 

Open to discussing offline this week so we can decrease tech debt by sticking to one pattern for all upcoming work.

PR #21, opened 2019-02-22
Refactor of Search Results and Search List Item

danwom:


amundsen_application/static/js/ducks/search/types.ts
```diff
@@ -0,0 +1,44 @@
...
+  email: string;
+}
+
+// Placeholder until the schema is defined.
+export interface DashboardSearchResult extends SearchResult  {
+  type: SearchResultType.dashboard;
+  title: string;
+}
+
+interface SearchResults<T> {
```
danwom: `SearchResults` might not be the right name for this. I also considered names like `SearchResultList`, `SearchTypeResults` and the like.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -6,44 +6,63 @@ import Pagination from 'react-js-pagination';
...
+  executeSearch: (term: string, page_index: number) => ExecuteSearchRequest;
   getPopularTables: () => GetPopularTablesRequest;
 }
 
 type SearchPageProps = StateFromProps & DispatchFromProps;
 
 interface SearchPageState {
-  pageIndex: number;
+  page_index: number;
   searchTerm: string;
```
danwom: We will need to add a selected tab here, but that's for another PR.

amundsen_application/api/metadata/v0.py
```diff
@@ -81,6 +81,8 @@ def _map_results(result: Dict) -> Dict:
             response_list = response.json().get('popular_tables')
             top4 = response_list[0:min(len(response_list), POPULAR_TABLE_COUNT)]
             popular_tables = [_map_results(result) for result in top4]
+            for table in popular_tables:
+                table['type'] = 'table'
```
verdan: can't we simply add this information directly in the `_map_results()` function instead of introducing a separate loop?

amundsen_application/api/search/v0.py
```diff
@@ -152,7 +161,15 @@ def _search(*, search_term: str, page_index: int) -> Dict[str, Any]:
...
             ]
             results = response.json().get('results')
-            results_dict['results'] = [{key: result.get(key, None) for key in params} for result in results]
+            for result in results:
+                result['type'] = 'table'
+
+            tables = {
+                'total_results': response.json().get('total_results'),
+                'page_index': int(page_index),
+                'results': [{key: result.get(key, None) for key in params} for result in results],
```
verdan: No need to have the extra loop here as we are already looping over the `results` on line 164. This can go under that loop. 
Or even, if the results are in `results_dict['tables']`, that means that the type is "table". Why do you need to specify that in each item instead? 
Same goes for the popular_tables API, you know these are tables, and can simply add the type "table" in the top level response instead of adding the type in each item. 

Not sure if I am missing something here.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -110,22 +131,25 @@ class SearchPage extends React.Component<SearchPageProps, SearchPageState> {
...
     if (showResultsList) {
-      const startIndex = (RESULTS_PER_PAGE * pageIndex) + 1;
-      const endIndex = RESULTS_PER_PAGE * ( pageIndex + 1);
-      const hasSearchResults = totalResults > 0;
+      const startIndex = (RESULTS_PER_PAGE * page_index) + 1;
+      const endIndex = RESULTS_PER_PAGE * ( page_index + 1);
+      const hasSearchResults = total_results > 0;
       const listTitle = hasSearchResults ?
-        `${startIndex}-${Math.min(endIndex, totalResults)} of ${totalResults} results` :
+        `${startIndex}-${Math.min(endIndex, total_results)} of ${total_results} results` :
```
verdan: This code block has at least 3 times a same condition `hasSearchResults`, and would be cleaner if you make a separate block for that condition instead of inline conditions in multiple places. 
And based on that condition you can also identify what are the results to be passed to `SearchList` in the `return()`

amundsen_application/static/js/components/common/ListItem/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+
+export default class ListItem extends React.Component<ListItemProps, {}> {
+
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    switch(this.props.item.type) {
```
verdan: Since this code block is specific to the SearchResults and won't be used anywhere else in the application as a generic `ListItem`, I think we should not put this in common, or maybe rename to `SearchListItem` if you believe that search would be in multiple places in the application.

amundsen_application/static/js/ducks/search/types.ts
verdan: I like `SearchResults` 👍 It's pretty clear I guess.

amundsen_application/static/js/ducks/search/reducer.ts
```diff
@@ -30,17 +34,29 @@ export function executeSearch(term: string, pageIndex: number): ExecuteSearchReq
 export type SearchReducerAction = ExecuteSearchRequest | ExecuteSearchResponse;
 
 export interface SearchReducerState {
-  pageIndex: number;
-  searchResults: SearchListResult[];
   searchTerm: string;
-  totalResults: number;
+  dashboards?: DashboardSearchResults;
```
ttannis: Is the intent to make `dashboards` and `users` required once we have them?

amundsen_application/static/js/ducks/search/sagas.ts
```diff
@@ -8,7 +8,7 @@ import {
 
 import {
   searchExecuteSearch,
-} from '../api/search/v0';
+} from './api';
```
ttannis: The intent behind this folder naming convention was in anticipation of  supporting different versions of apis, and don't want to remove that concept. However I think I have to take this though offline for us to discuss. One thing we haven't discussed is what our search changes will mean for our API versioning. For example does the frontend service need that concept, and if so how does that affect how we should be implementing new features.

amundsen_application/api/search/v0.py
danwom: I will move the `type` assignment into the loop. I'm a complete python newbie so I'm not 100% sure how line 170 works but will find out.

We have plans for search to have an "all" category that will return ordered results of different types. Ranking among different search types will take time to get right, so our initial version will simply separate each type into different tabs. We also have a bookmark feature that will let users save tables, dashboards, and users. In these cases it will be very useful for each result to indicate its type.

amundsen_application/static/js/components/common/ListItem/index.tsx
danwom: We are also building a bookmark feature. The user profile page will display a list of bookmarks and use the same data format as `SearchResult`. I am trying to figure out a better name that fits both use cases and isn't too confusing or awkward. `Result` or `Item` aren't really descriptive enough, and `SearchResult` is too narrow.

Completely open to suggestions.

amundsen_application/static/js/ducks/search/sagas.ts
danwom: We can rename to '/search/api/v0.ts`. I think that maintains the versioning as well as keeping the APIs and ducks in the same folder.

amundsen_application/static/js/ducks/search/reducer.ts
danwom: That was the intent, but technically we can make these required already since we have the initial state that doesn't get changed.

amundsen_application/api/search/v0.py
```diff
@@ -109,11 +122,8 @@ def _search(*, search_term: str, page_index: int) -> Dict[str, Any]:
 
     TODO: Define an interface for envoy_client
     """
-    results_dict = {
-        'results': [],
+    results_dict: Dict[str, Any] = {
```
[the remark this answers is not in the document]
danwom: The moment I nested the objects, Mypy started complaining. I had tried initializing the way you described but there were some errors when I tried to assign the results. Maybe it was because I was trying to modify results after assignment. I will try again.

amundsen_application/static/js/components/SearchPage/SearchList/index.tsx
```diff
@@ -1,9 +1,10 @@
 import * as React from 'react';
-import SearchListItem from './SearchListItem';
-import { SearchListResult } from '../types';
+import ListItem from '../../common/ListItem';
+import { SearchResult } from '../../../ducks/search/types';
```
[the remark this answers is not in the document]
danwom: I think this will need further discussion to come up with a good pattern. I'll keep this in mind.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -6,44 +6,63 @@ import Pagination from 'react-js-pagination';
 import SearchBar from './SearchBar';
```
ttannis: nit nit: Our mix of having variables in camelCase (JS standard) vs snake_case (Python standard) in our React code is something I'd like to avoid making worse. In the future can we move closer to camelCase or at least prevent mixing in more snake_case? 

Not sure what the best approach is though. I recognize it's a pretty tedious/low-priority task to parse all of our responses to create objects with all camelCase variables, but the consistency will be nice and may help when we want to start enforcing stricter lint rules.

amundsen_application/static/js/components/common/ListItem/index.tsx
ttannis: Maybe `ResourceListItem`.

amundsen_application/static/js/components/common/ListItem/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface SearchIndexParams {
```
ttannis: nit: since these values are used for our logging on the backend at some point we can perhaps rename this to `LoggingParams`, sort of to help decouple concepts/names in the `ListItem` with search since they will be used for things other than search.

ttannis: Hm...and to follow that up, the `ListItem` maybe be used for things other that search, but the logging params might not be. Keeping search in the name would be appropriate then.

amundsen_application/static/js/components/SearchPage/index.tsx
danwom: I too, dislike the mixing of different standards, but converting nested fields for every data payload seems onerous.

One convention that we could follow is to use camelCase for all functions and variables defined in JS, but preserve snake_case for any data that we receive.

amundsen_application/static/js/components/common/ListItem/index.tsx
danwom: `SearchResult => Resource` and `ListItem => ResourceListItem`? On both the search and bookmark sections we would show `Resource[]`. Sounds like this might work.

amundsen_application/static/js/components/SearchPage/index.tsx
danwom: I just noticed that my find-and-replace also hit JS defined variables too. Will fix.

PR #24, opened 2019-02-25
Add the setup line for amundsen data builder

verdan:
It is important to install the amundsen data builder before running the test script locally, 
otherwise the imports won't work in the test script.


feng-tao:
thanks @verdan

PR #26, opened 2019-02-26
Update Owner Editing UI

ttannis:
This PR updates the UI for the owner editing feature.

**Summary of Changes**
1. `EditableList` was re-written in favor of a more specialized `OwnerEditor`. This simplifies the component and its types and was done because we do not  have a feature on our roadmap that justifies managing the complexities of a generic `EditableList`.
    * `OwnerEditor` renders an `AvatarLabel` for each of its `itemProps`.
    *  Selecting the `+ Add` button at the bottom of the list opens up a modal for editing.
        <image src ="https://user-images.githubusercontent.com/1790900/53437696-035f0980-39b3-11e9-81ca-6ad9aab1c761.png" width="50%">
        oe2
    * This component maintains a copy of its `itemProps` as `tempItemProps`. `tempItemProps` is modified when adding/deleting items in the editing modal. Cancelling the edit will reset `tempItemProps`. Saving the edit will dispatch an action with an array of  `UpdatePayload` items. In the Redux layer, this is used to generate an array of `PUT`/`DELETE` requests that is executed with `yield all`.
    * A successful update will update `itemProps` from the Redux store. 
       <image src ="https://user-images.githubusercontent.com/1790900/53437544-a1060900-39b2-11e9-9299-4ef5018ae882.png" width="50%">
    * A failed update will cause the component to show an error text to refresh the page. This is currently our least painful option to deal with any errors with any of the updates.
      

2. `TableOwnerEditableList` renamed `OwnerEditor`, and updated accordingly now that it wraps the new `OwnerEditor` component.
3. Introduced new icons and colors to best match the new UI design. We have a darker purple for button hover, and also needed lighter grey. `$gray-lighter` was darkened for contrast with the new `$gray-lightest`.

**Future Considerations**
Potentially introducing `redux-batched-actions` to make it such that `OwnerEditor` will be triggered to render only once, after the final update to the owners state.


amundsen_application/static/js/components/AvatarLabelList/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+           className='btn add-list-item'
+           onClick={this.handleShow}>
+             <img className='icon icon-plus-circle'/>
+             <span>Add</span>
+          </button>
+        }
+
+        <Modal className='editable-list-modal' show={this.state.showModal} onHide={this.cancelEdit}>
+          <Modal.Header className="text-center" closeButton={false}>
+            <Modal.Title>Owned By</Modal.Title>
```
danwom: Can "Owned By" and "Enter an email address" be customized? It seems like this component should be generic enough to handle owner editing as well as other editing (select users to share ?). 

If the intent is to only be used for owner editing, then this component should be called "OwnerEditor" or the like.

amundsen_application/static/js/components/AvatarLabelList/styles.scss
```diff
@@ -0,0 +1,152 @@
+@import 'variables';
+
+.editable-list-component {
```
danwom: Can we change class names to reflect the new component name?

amundsen_application/static/js/components/AvatarLabelList/styles.scss
```diff
@@ -0,0 +1,152 @@
...
+    &:not([disabled]):hover {
+      color: $gradient-4;
+      background-color: transparent;
+      border-color: transparent;
+
+      .icon {
+        background-color: $gradient-4;
+      }
+    }
+  }
```
danwom: I think we should define a new type of button where the button is essentially an icon. We only had one example of this button with the old `edit-button`, but it looks like we have a few more added here. 

Perhaps we could define a `icon-button` or `flat-icon-button` in `_buttons-default.scss` and give it the transparent background/border/shadow/etc. Then wherever we want to use it it can be as simple as 

```
<button className="btn icon-button">
  <img className="icon icon-circle-plus"/>
</button>
```

Same applies for delete

amundsen_application/static/js/components/AvatarLabelList/index.tsx
```diff
@@ -0,0 +1,215 @@
+
```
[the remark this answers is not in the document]
ttannis: I don't know how much it will simplify the logic, but it can still only be done if we leverage Redux correctly, since that's the only way to have one component edit a piece of state that another component is also displaying.

I think I may consider doing the rework of the tableMetadata state, and then update this PR.

amundsen_application/static/js/components/AvatarLabelList/index.tsx
ttannis: I'll rename it. My intent  for this re-write to worry about making a decent owner editing component, and worry about re-useability when we cross that bridge. That being said, no harm in having these pieces of text also be props.

amundsen_application/static/js/components/AvatarLabelList/styles.scss
ttannis: Oversight on my part

amundsen_application/static/js/components/AvatarLabelList/styles.scss
ttannis: Mind if I do that in a separate button-focused PR, where we take the 3 button styles that will get reused from this component and define them for the whole app? Notice the overrides I had to add for the `Cancel` and `Save` button here. These will also get reused in all of our existing and future editing modals.

amundsen_application/static/js/components/AvatarLabelList/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+    }
+    else {
+      this.state.itemProps.forEach((value, key) => {
+        const item = (
+          <li key={`list-item:${key}`}>
+            { React.createElement(AvatarLabel, value) }
+          </li>
+        );
+        items.push(item);
+      });
```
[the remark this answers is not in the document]
ttannis: es6 = :trollface: ?

amundsen_application/static/js/components/AvatarLabelList/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+  PUT = 'PUT',
+  DELETE = 'DELETE',
+}
+
+export interface DispatchFromProps {
+  onUpdateList: (updateArray: { method: UpdateMethod; id: string; }[], onSuccess?: () => any, onFailure?: () => any) => void;
+}
+
+export interface ComponentProps {
+  errorText?: string | null;
```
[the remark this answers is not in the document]
ttannis: I think the absence of a value is not treated the same as it being null, so if I want to allow `{ readOnly: true }` -- in which case one doesn't need to pass the `errorText` since that is only used if the component is editable -- I would have to make `errorText` optional. But as with most things Typescript, let me re-run the type checker and see what it complains about.

amundsen_application/static/js/components/AvatarLabelList/index.tsx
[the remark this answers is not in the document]
danwom: Yeah I wasn't sure how much simpler the split components would be given our current setup. Okay to revisit this later.

amundsen_application/static/js/components/AvatarLabelList/styles.scss
danwom: Let's put a TODO here and file a ticket.

ttannis:
Note: This PR will be rebased after https://github.com/lyft/amundsenfrontendlibrary/pull/30 is merged. After which I will address review comments and improve how the component responds to `PUT`/`DELETE` success/failures.

amundsen_application/static/js/components/AvatarLabelList/index.tsx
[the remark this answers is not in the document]
ttannis: I'll keep as is. A second use case for a  "+ Add" button that opens a modal will justify separating these out.

amundsen_application/static/js/components/AvatarLabelList/index.tsx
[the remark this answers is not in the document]
ttannis: Confirming I need the `errorText` to be marked as optional since we don't want to require it as the component only needs it if it is not `readOnly`.

amundsen_application/static/js/components/AvatarLabelList/index.tsx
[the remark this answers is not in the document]
ttannis: FYI I removed the use of `Map()` and switched back to regular objects. Our use cases right now really only support mapping a string to an object, so `Map` might be overkill to start using everywhere.

amundsen_application/static/js/components/common/LoadingSpinner/styles.scss
```diff
@@ -5,5 +5,6 @@
   height: 100px;
   margin: auto;
   margin-top: 20%;
+  margin-bottom: 20%;
```
ttannis: Added this so it looks slightly better when used in the modal body. Current styles temporarily increase the height of the owner editing section but I didn't invest effort into custom styling the LoadingSpinner because we are supposed to implement the shimmer effect for all loading states.

ttannis:
@danwom 
1. Addressed review comments
2. Made some changes to the owners state, and store it as a key/value object instead of an array

The owner editing now uses the same approach as tags for batch edits, with the intent still being to look into `redux-batched-actions` to see if it can help better manage the logic updating logic. Ideally we want to be able to send each `PUT`/`DELETE`, update the owner/tag state based on whether or not each request succeeds or fails, and only then should the connected component re-render.

amundsen_application/static/css/_variables-default.scss
```diff
@@ -13,13 +13,15 @@ $gradient-1: #f2f5fe;
 $gradient-2: #cad6ff;
 $gradient-3: #5679ff;
 $gradient-4: #3250c8;
+$gradient-5: #28409f;
```
danwom: Somewhat off topic, but how do you feel about renaming this to something like `$brand-color-1` or `$brand-light`? `$gradient` isn't really obvious what it is.

amundsen_application/static/js/ducks/tableMetadata/api/helpers.ts
```diff
@@ -21,7 +21,12 @@ export function getTableDataFromResponseData(responseData) {
  * Parses the response for table metadata to return the array of table owners
  */
 export function getTableOwnersFromResponseData(responseData) {
-  return responseData.owners;
+  // TODO: owner needs proper id, until then we have to remember that we are using display_name
+  const ownerObj = responseData.owners.reduce((obj, owner) => {
```
danwom: I would rename these as `accObj` and `currentOwner`. It's easy to forget what order these come.

Also I wonder if changing from an array to object will affect render order. This may or may not be an issue.

amundsen_application/static/css/_variables-default.scss
ttannis: I'll go with `$brand-color` since that allows us to have as many as we need instead of trying to come up with variations of `light/lighter/lightest`. Will file a ticket.

PR #27, opened 2019-02-26
Add re-useable Flag component

ttannis:
This PR adds a `Flag` component, which will be needed for upcoming features. The default color is dark text on light gray text. Colors can be customized with css by overriding `color` and `background-color` on the component.

**Examples**
screen shot 2019-02-26 at 12 07 50 pm

Also noticed the need for css tweak for `.detail-list-header` when testing this.

### DIFF
--- a/amundsen_application/static/js/components/common/Flag/index.tsx
+++ b/amundsen_application/static/js/components/common/Flag/index.tsx
@@ -5,20 +5,18 @@ import './styles.scss';
 
 interface FlagProps {
   text: string;
+  style?: string;
 }
 
-const Flag: React.SFC<FlagProps> = ({ text }) => {
-
+const Flag: React.SFC<FlagProps> = ({ text, style }) => {
   return (
-    <div className='flag-component'>
-      <label>{text.toUpperCase()}</label>
-    </div>
+    <span className={`flag label ${style}`}>{text.toUpperCase()}</span>
   );
-
 };
 
 Flag.defaultProps = {
   text: '',
+  style: 'label-default',
 };
 
 export default Flag;
--- a/amundsen_application/static/js/components/common/Flag/index.tsx
+++ b/amundsen_application/static/js/components/common/Flag/index.tsx
@@ -1,0 +1,24 @@
+import * as React from 'react';
+
+// TODO: Use css-modules instead of 'import'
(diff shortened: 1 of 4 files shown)


danwom:
I think we should leverage the bootstrap `label` (merged into `badge` in 4.0) component. The variables ($label-color, $label-color-hover, etc) can be found in `node_modules/bootstrap-sass/assets/stylesheets/bootstrap/labels.scss`.

https://getbootstrap.com/docs/3.3/components/#labels
https://getbootstrap.com/docs/4.1/components/badge/

ttannis:
@danwom I took a look and it seems like we'd still need css to override the label color if we want to match our designs. But I personally don't mind if the text is white for all cases. Made `$label-primary-bg` map to one of the gradient colors for some theme consistency.

**New Examples**

amundsen_application/static/js/components/common/Flag/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+const Flag: React.SFC<FlagProps> = ({ text, style }) => {
+  return (
+    <span className={`flag label ${style}`}>{text.toUpperCase()}</span>
+  );
+};
+
+Flag.defaultProps = {
+  text: '',
+  style: 'label-default',
```
danwom: nit: Can we call this something other than "style"? One might assume it takes a css property instead of a className.

amundsen_application/static/js/components/common/Flag/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+import './styles.scss';
+
+interface FlagProps {
+  text: string;
+  style?: string;
+}
+
+const Flag: React.SFC<FlagProps> = ({ text, style }) => {
+  return (
+    <span className={`flag label ${style}`}>{text.toUpperCase()}</span>
```
danwom: Maybe put a TODO here when we upgrade to bootstrap 4. 'label' should be changed to 'badge'

amundsen_application/static/js/components/common/Flag/index.tsx
ttannis: I can see that. I'll go with `labelStyle`.

PR #31, opened 2019-03-01
Minor visual changes

danwom:
- Fixed panel borders/shadow/padding
- Removed double margin from .table-details > .row

### DIFF
--- a/amundsen_application/static/js/components/TableDetail/styles.scss
+++ b/amundsen_application/static/js/components/TableDetail/styles.scss
@@ -1,53 +1,42 @@
 @import 'variables';
 
 .table-detail {
-  margin: 40px auto 48px;
-  display: flex;
-  flex-direction: row;
-}
+  margin: 64px auto 48px;
 
-/*
-  white-space: nowrap breaks without this width on the row
-*/
-.table-detail .row {
-  width: 100%;
-  margin: 0;
-}
-
-
-.error-label {
-  margin: 48px auto;
-  width: fit-content;
-}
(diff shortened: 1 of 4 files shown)


danwom:
Before: 
screen shot 2019-02-28 at 2 21 55 pm


After: 
screen shot 2019-02-28 at 4 08 49 pm

amundsen_application/static/js/components/TableDetail/styles.scss
```diff
@@ -1,53 +1,42 @@
...
-  flex-direction: row;
-}
-
-/*
-  white-space: nowrap breaks without this width on the row
-*/
-.table-detail .row {
-  width: 100%;
-  margin: 0;
-}
```
danwom: The `margin: 0` was adding an extra 30px of margin since by default bootstrap sets row as `margin: -15px` and add it back as padding in each `col-x` element. I also did some testing and found no issues with column descriptions overflowing. There must have been another cause that was fixed.

amundsen_application/static/js/components/TableDetail/styles.scss
```diff
@@ -1,53 +1,42 @@
...
-
+  margin: 64px auto 48px;
```
ttannis: LGTM but why 40px --> 64px?

danwom: Every other page has 64px of margin-top. I don't think this page needs a shorter margin than everything else.

PR #32, opened 2019-03-01
Rearrange import order of bootstrap, custom, and default variables

danwom:


amundsen_application/static/css/_variables.scss
```diff
@@ -1,6 +1,7 @@
-// Bootstrap Default Values
-@import '~bootstrap-sass/assets/stylesheets/bootstrap/variables';
 // Amundsen Default Values
 @import 'variables-default';
 // Per-Client Custom Values
 @import 'variables-custom';
+// Bootstrap Default Values
+@import '~bootstrap-sass/assets/stylesheets/bootstrap/variables';
+
```
danwom: Learned a bit about the nuances of `!default` values in SASS. 

Regular variables take the last definition.
```
$color: red;
$color: blue

$color => blue
```

Default variables take the first value.
```
$color: red !default;
$color: blue !default;

$color => red
```


This most closely mimics our current setup.
```
$color: red !default; // variables-default
$color: black; // variables-custom
$color: blue !default; // bootstrap/variables

$color => black
```

ttannis:
I’m currently not understanding the difference between these two configurations, can you explain what problem this addresses and in what case are the final variable settings not the same?

danwom:
As written, the resultant styles _should_ remain the same. In the future this will enable us to explicitly define fewer variables while obtaining the same level of customization.

The main use case is for derived values defined in `bootstrap/variables`. Consider the following rules for the Navbar.
```
$navbar-height:                    50px !default;
$navbar-padding-vertical:          (($navbar-height - $line-height-computed) / 2) !default;
```

If we decide to change `$navbar-height` in the old structure, we would also have to recalculate `$navbar-padding-vertical`. In the new structure`$navbar-padding-vertical` gets calculated for the first time using the custom value since `$navbar-height: 50px !default` has an existing value and is ignored.

ttannis:
> As written, the resultant styles _should_ remain the same. In the future this will enable us to explicitly define fewer variables while obtaining the same level of customization.
> 
> The main use case is for derived values defined in `bootstrap/variables`. Consider the following rules for the Navbar.
> 
> ```
> $navbar-height:                    50px !default;
> $navbar-padding-vertical:          (($navbar-height - $line-height-computed) / 2) !default;
> ```
> If we decide to change `$navbar-height` in the old structure, we would also have to recalculate `$navbar-padding-vertical`. In the new structure`$navbar-padding-vertical` gets calculated for the first time using the custom value since `$navbar-height: 50px !default` has an existing value and is ignored.

Awesome. Makes sense.

verdan:
Nice, really useful to make amundsen whitelabel application. 👍

PR #33, opened 2019-03-01
Make Flag text sentence case by default

ttannis:


amundsen_application/static/js/components/common/Flag/index.tsx
```diff
@@ -8,11 +8,15 @@ interface FlagProps {
...
+function toSentenceCase(str: string): string {
+  return `${str.charAt(0).toUpperCase()}${str.slice(1).toLowerCase()}`;
+}
+
 const Flag: React.SFC<FlagProps> = ({ text, labelStyle }) => {
   // TODO: After upgrading to Bootstrap 4, this component should leverage badges
   // https://getbootstrap.com/docs/4.1/components/badge/
   return (
-    <span className={`flag label ${labelStyle}`}>{text.toUpperCase()}</span>
+    <span className={`flag label ${labelStyle}`}>{toSentenceCase(text)}</span>
```
danwom: I also read the feedback from Matt, but I'm not entirely convinced we should bake this into the component itself.

Alternatives:
1) Leave this component alone and let the consumer of this component determine capitalization.
2) Provide some options for customization (all-caps, all-lower, mixed). This might be overengineering however.

ttannis: How about I give it a prop `caseStyle`. The default flag will take the text as is, and the user can optionally set `sentenceCase`, `upperCase`, `lowerCase`. And then eventually if need be we can support others.

I like the idea of having this kind of support baked into the component in some way so that we don't have to write one-off pieces of logic to parse/check our text, and would rather be able to at least set a prop.

ttannis:
@danwom added option to customize the case. Default will just render the text as is and we currently support sentencecase, uppercase, and lowercase. 
```
<Flag text='ALUmni' labelStyle = 'label-danger' />
<Flag text='ALUmni' labelStyle = 'label-danger' caseType='upperCase' />
<Flag text='ALUmni' labelStyle = 'label-danger' caseType='lowerCase' />
<Flag text='ALUmni' labelStyle = 'label-danger' caseType='sentenceCase' />
```

PR #34, opened 2019-03-01
Add shared button styles + modify tag modal to match owner modal

ttannis:
In this PR I added 3 new specifications to the button configurations:
1. `btn-icon` for the "flat" buttons that render icons and are gray by default and a brand color on hover.
2. `btn-cancel` and `btn-save` which are currently used in our modal designs.

I've also update the modal in the `TagInput` to match the modal for `OwnerEditor`.

### DIFF
--- a/amundsen_application/static/css/_buttons-default.scss
+++ b/amundsen_application/static/css/_buttons-default.scss
@@ -41,11 +41,12 @@
     }
   }
 
-  &.btn-icon {
+  &.btn-flat-icon {
     border: none;
     color: $gray-light;
-    -webkit-box-shadow: none !important;
     box-shadow: none !important;
+    padding: 0px;
+    text-align: left;
 
     &:focus,
     &:not(.disabled):hover,
@@ -58,7 +59,7 @@
       }
     }
   }
-  
+
   &.btn-primary {
     border-width: 2px;
     -webkit-box-shadow: 8px 9px 20px -15px rgba(0,0,0,0.46);
--- a/amundsen_application/static/css/_buttons-default.scss
+++ b/amundsen_application/static/css/_buttons-default.scss
@@ -28,6 +28,37 @@
     margin-bottom: 4px;
   }
 
+  &.btn-cancel {
+    border-color: $gray-lighter;
+    font-family: $font-family-sans-serif-bold;
+    font-weight: $font-weight-sans-serif-bold;
+
+    &:focus,
+    &:not(.disabled):hover,
+    &:not([disabled]):hover {
+      background-color: $gray-lighter;
+      border-color: $gray-lighter;
+    }
+  }
+
+  &.btn-icon {
+    border: none;
+    color: $gray-light;
+    -webkit-box-shadow: none !important;
+    box-shadow: none !important;
+
+    &:focus,
+    &:not(.disabled):hover,
+    &:not([disabled]):hover {
+      background-color: transparent;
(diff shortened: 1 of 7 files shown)


amundsen_application/static/css/_buttons-default.scss
```diff
@@ -28,6 +28,37 @@
...
+    &:not([disabled]):hover {
+      background-color: $gray-lighter;
+      border-color: $gray-lighter;
+    }
+  }
+
+  &.btn-icon {
+    border: none;
+    color: $gray-light;
+    -webkit-box-shadow: none !important;
```
danwom: I don't think we need to specify -webkit for box-shadow. 

https://caniuse.com/#search=box-shadow

amundsen_application/static/js/components/OwnerEditor/index.tsx
```diff
@@ -145,7 +145,7 @@ class OwnerEditor extends React.Component<OwnerEditorProps, OwnerEditorState> {
             placeholder='Enter an email address'
             ref={ this.inputRef }
           />
-          <button className="btn btn-light add-btn" type="submit" aria-label="Add Item">
+          <button className="btn btn-light add-button" type="submit" aria-label="Add Item">
```
danwom: Shouldn't this also be a `btn-icon`?

amundsen_application/static/css/_buttons-default.scss
```diff
@@ -28,6 +28,37 @@
...
+
+    &:focus,
+    &:not(.disabled):hover,
+    &:not([disabled]):hover {
+      background-color: $gray-lighter;
+      border-color: $gray-lighter;
+    }
+  }
+
+  &.btn-icon {
```
danwom: I think the class name should indicate that this is a flat button with no border. We already have some buttons that have icons inside.

amundsen_application/static/js/components/OwnerEditor/index.tsx
danwom: Oh nvm this is inside the modal

amundsen_application/static/js/components/OwnerEditor/styles.scss
```diff
@@ -3,25 +3,10 @@
 @import 'variables';
 
 .owner-editor-component {
-  .btn.add-list-item {
-    color: $gray-light;
+  .btn.add-item-button {
     padding: 4px 1px;
```
danwom: I the padding and height should get baked into the `btn-icon` class, perhaps with a second modifier that indicates it comes with text.

amundsen_application/static/css/_buttons-default.scss
ttannis: `btn-flat-icon`?

amundsen_application/static/js/components/OwnerEditor/styles.scss
ttannis: So you're suggesting a new style, or modifying the existing one to apply the relevant styles to text if it exists?

ttannis:
@danwom 
1. Renamed new style to `btn-flat-icon`
2. Moved `padding` to `btn-flat-icon`, but kept the random `1px` left/right padding on the `add-item-button` because it just happens the line up a bit better because of how the icon is drawn.
3. Made the one-off changes to `AvatarLabel` for a 24px avatar and removing `width: 80%`

PR #43, opened 2019-03-06
Added support for UserListItems

danwom:
- Restyled ResourceListItem to support multiple columns
- Added timestamp for TableListItem
- Added UserListItem


danwom:
Tables
screen shot 2019-03-06 at 1 36 40 pm


Users
screen shot 2019-03-06 at 1 16 31 pm

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
```diff
@@ -3,10 +3,8 @@ import { Link } from 'react-router-dom';
 
 import { LoggingParams, TableResource} from '../types';
 
-import './styles.scss';
-
 interface TableListItemProps {
-  item: TableResource;
+  table: TableResource;
   logging: LoggingParams;
```
ttannis: Slipped my mind previously -- `logging` should be optional on all the list item components. We use these values for instrumentation when a user clicks on the item from the search results, but we won't be using them when someone clicks on the item from the profile page.

amundsen_application/static/js/components/common/ResourceListItem/UserListItem/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  render() {
+    const { user } = this.props;
+    return (
+      <li className="list-group-item">
+        <Link className="resource-list-item user-list-item" to={ this.getLink() }>
+          <Avatar name={ user.name } size={ 24 } round={ true } />
+          <div className="content">
+            <div className="col-xs-12 col-sm-6">
+              <div className="main-title">{ user.name }</div>
```
ttannis: We're missing support for the `Flag` component next to the name if the user is not active.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
```diff
@@ -15,38 +13,41 @@ class TableListItem extends React.Component<TableListItemProps, {}> {
...
+              <div className="main-title truncated">{ `${table.schema_name}.${table.name}`}</div>
+              <div className="description truncated">{ table.description }</div>
+            </div>
+            <div className="col-sm-3">
+              <div className="secondary-title">Frequent Users</div>
+              {/*TODO - Replace with a link to a real user*/}
+              <div className="description"><a href="#">Ash Ketchum</a></div>
+            </div>
+            <div className="col-sm-3">
+              <div className="secondary-title">Last Updated</div>
```
ttannis: This is supposed to be "Latest Data" according to the new design, which I hope was intentional and which I think is a better phrase. "Last Updated" might be misinterpreted to think this date was maybe the last time someone edited a.k.a. "updated" some metadata on that resource in Amundsen, when really this the date for which the table has the most recent data.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
```diff
@@ -15,38 +13,41 @@ class TableListItem extends React.Component<TableListItemProps, {}> {
...
+              <div className="description">
```
ttannis: I also think this column should be smaller and not given the same space as `Frequent Users`. In our designs it seems intentionally further pushed to the right hand side of the list item. 

I think this makes sense because the information in this column is always just going to a be a date that gets formatted in such a way that it's always 11-12 characters and therefore guaranteed to fit into some kind of minimum width. Meanwhile the `Frequent Users` section is an unknown list where we will want to render as many items as possible so it needs more real-estate than an almost constant length date string. Thoughts?

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
danwom: As I was revisiting the logging parameters, I realized there are deep changes that I would like to make. I think it makes sense to break this into a separate PR as not to pollute this one.

danwom: Of the logging parameters I think `source` should be required, but `index` should be optional. The SearchList has optional logging params, but I think they should be required. Etc.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
danwom: Gave the `Frequent Users` column 4 columns when possible, but down to 3 columns when the date needed more space.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
ttannis: I'm not sure I understand the thinking, can you elaborate? I *think* I understand the idea behind keeping `source` -- we could send something like `profile_page` as the `source` for other options and not send the `index` because it doesn't matter in that case. Is that where you're coming from?

amundsen_application/static/js/components/common/Flag/index.tsx
```diff
@@ -32,14 +32,14 @@ const Flag: React.SFC<FlagProps> = ({ caseType, text, labelStyle }) => {
   // TODO: After upgrading to Bootstrap 4, this component should leverage badges
   // https://getbootstrap.com/docs/4.1/components/badge/
   return (
-    <span className={`flag label ${labelStyle}`}>{convertText(text, caseType)}</span>
+    <span className={`flag label label-${labelStyle}`}>{convertText(text, caseType)}</span>
```
ttannis: nit: I think there's value in expecting someone to enter `label-success` instead of `success`, because then our supported `labelStyles` can just match the existing bootstrap styles that everyone is familiar with. 

What I'm getting at is that someone coming in to use this component will have to remember that our supported styles don't need the `label-` prefix, so I think it might be more convenient to just match what we're leveraging.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
ttannis: What content is hidden? I think we should reconsider the idea of hiding any information that Product has deemed useful for our users.

amundsen_application/static/js/components/common/Flag/index.tsx
danwom: I was thinking we should abstract away the concept of the `label` and `label-{default | alert| etc}` classes so users of the Flag component don't care what the CSS classes are used under the hood. Also consider this will get converted to `badge badge-alert` in the future.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
danwom: `Frequent Users` and `Latest Update` go away in XS screens. There is literally no space to put them in. 

screen shot 2019-03-07 at 3 23 18 pm

We might be able to get away with showing just 2 columns, but at the expense of hiding the table name, which is most important.
screen shot 2019-03-07 at 3 21 13 pm


That being said, I don't think most people browse Amundsen on mobile so this should be a < 1% issue.

danwom: We could also put them below, but that would make the search results really long. Probably better to hide them.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
danwom: Yes, every time you display a `ResoureListItem` it will have a `source` that should be logged. I suppose they will also have an `index` given that they are in a list. Whether or not recording that index is useful is another question of which I am doubtful.

If anything this would be an argument for making the logging params mandatory for all cases, rather than making them optional.

ttannis: ^ alright lets complete this discussion when that PR comes.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
ttannis: Is that first image what the current behavior is? If so then I think that's fine. I thought "hidden" meant that the information would just completely go away, but that current view still lets the user know there is information there and that they need to view it on a larger screen.

amundsen_application/static/js/components/common/Flag/index.tsx
ttannis: I see. So this way when we switch to badges we won't have to update the `labelStyle` value on each usage, makes sense.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
danwom: Resolved this conversation offline.

PR #44, opened 2019-03-06
Skeleton implementation of ProfilePage w/ mock data

ttannis:
This PR implements skeleton support for the profile page. 
image

**Summary of Changes**
_Python_
1. Added a `user` route to `api/metadata/v0`. This method assumes that there will be a request argument called `user_id`. It currently returns a test payload for both an active user (pass any `user_id`) and an inactive user (pass `user_id` = ‘alumni’)
2. Updated the `User` class in `models/user` to support the new parameters, and to fail User validation if we don’t have a `user_id`.

_Redux_
1. Added a `profilePageUser` to the user reducer, and added support for a `getUserById()` method. 
2. The logic to get the current user also calls `getUserById` and merges the results. This is because we need information like owned resources stored in the `currentUser` state.

_React_
1. Added the 4 new svg files for new icons. Our official design shows these as $gray-dark by default, and I assume $brand-color-4 on hover for consistency. Created a new class `.btn-flat-icon-dark` for this, but will run this by design to see if we can keep the $gray-light and $brand-color-4 on hover like we use elsewhere.
2. For our  “breadcrumb” concept, I’ve set up as a `<div>` that will take the `container` style, but have a shorter `margin-top`.
3. `NavBar` avatar links to the profile for the current user. 
4. Created a new `ProfilePage` component and linked it to the route "/user/:userId”. Note that this component was written to render properly if the user is missing any properties. This was done in the hopes that its makes it easier to implement the shimmer effect css as we move away from `LoadingSpinner`. The page should be able to render in some blank state without the data from the request.


amundsen_application/static/js/components/ProfilePage/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+        </div>
+      </div>
+    )
+  }
+
+  render() {
+    return (
+      <DocumentTitle title={ `${this.state.user.display_name} - Amundsen Profile` }>
+        <div>
+          <div className="container amundsen-crumb">
```
danwom: The breadcrumb should be it's own component, but that can be saved for a follow up.

amundsen_application/static/js/ducks/user/reducer.ts
```diff
@@ -2,27 +2,48 @@ import {
...
 } from './types';
 
-type UserReducerAction = GetCurrentUserRequest | GetCurrentUserResponse;
+type UserReducerAction =
+  GetCurrentUserRequest | GetCurrentUserResponse |
+  GetUserRequest | GetUserResponse ;
 
 export interface UserReducerState {
   currentUser: CurrentUser;
+  profilePageUser: User;
```
danwom: I think `CurrentUser` might be mistaken for the current user being rendered, which is why you might have made `profilePageUser` so explicit. What do you think about disambiguating the users with something like

`currentUser -> loggedInUser`
 `profilePageUser -> profileUser or userProfile`

This can also be a TODO

amundsen_application/api/metadata/v0.py
```diff
@@ -553,3 +553,34 @@ def _log_update_table_tags(*, table_key: str, method: str, tag: str) -> None:
...
+            'first_name': 'Firstname',
+            'last_name': 'Lastname',
+            'email': 'test@test.com',
+            'display_name': 'Firstname Lastname',
+            'profile_url': 'https://github.com/lyft/amundsenfrontendlibrary',
+            'user_id': user_id,
+            'github_name': 'lyft',
+            'is_active': True,
+            'manager_name': 'Roald Amundsen',
+            'role_name': 'Software Engineer',
```
danwom: Shouldn't this be `Explorer`? :)

amundsen_application/static/css/styles.scss
```diff
@@ -16,3 +16,8 @@ form {
   margin-bottom: 0;
 }
 
+.container {
+  &.amundsen-crumb {
```
danwom: This should go into a new component.

I think every page should have the standard top-margin (64px) whether or not a crumb is present. The crumb itself is exempt from the standard margin and can do so with absolute positioning or a negative margin-top (personal preference). There are a few options here.

amundsen_application/static/js/components/ProfilePage/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+    const { user } = nextProps;
+    return { user };
+  }
+
+  componentDidMount() {
+    this.props.getUserById(this.userId);
+  }
+
+  // TODO: consider moving logic for empty content into Tab component
+  createEmptyTabMessage = (message: string) => {
```
danwom: I considered whether the tab component should provide a default content container or if the consuming component would rather have full control on the display. Will need to think about this some more.

amundsen_application/static/css/_buttons-default.scss
```diff
@@ -60,6 +60,32 @@
     }
   }
 
+  // TODO: Used for profile page icon buttons, need to check if we want to unify
+  // Exact design uses dark text by default
+  // Assuming hover color should follow $brand-color-4 precedent
```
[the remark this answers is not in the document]
ttannis: Let me just change to `.btn-flat-icon` since that seems to be what we're both fine with, and then see if there's a strong opinion to change it back.

amundsen_application/static/js/components/ProfilePage/index.tsx
ttannis: No problem. I'll leave the `TODO` in here and we can resolve it in the follow up ticket when we have backend support.

amundsen_application/static/js/ducks/user/reducer.ts
ttannis: I like those names. Will go ahead and make the change.

amundsen_application/api/v0.py
```diff
@@ -11,10 +11,10 @@
 blueprint = Blueprint('api', __name__, url_prefix='/api')
 
 
-@blueprint.route('/current_user', methods=['GET'])
+@blueprint.route('/auth_user', methods=['GET'])
 def current_user() -> Response:
```
danwom: Might as well change this function name as well.

amundsen_application/static/js/components/ProfilePage/index.tsx
```diff
@@ -88,77 +88,64 @@ class ProfilePage extends React.Component<ProfilePageProps, ProfilePageState> {
...
       <DocumentTitle title={ `${this.state.user.display_name} - Amundsen Profile` }>
         <div>
-          <div className="container amundsen-crumb">
-            <Link to={`/`}>
-              <button className='btn btn-flat-icon'>
-                <img className='icon icon-left'/>
-                <span>Search Results</span>
-              </button>
-            </Link>
+          <Breadcrumb path='/' text='Search Results'/>
```
danwom: Is this `<Breadcrumb>` intentionally outside of the container element? The left position might behave strangely.

amundsen_application/static/js/components/common/Breadcrumb/styles.scss
```diff
@@ -0,0 +1,6 @@
+@import 'variables';
+
+.container.amundsen-breadcrumb {
+  margin-top: 13px;
+  margin-bottom: -32px;
+}
```
danwom: Would prefer to nest the breadcrumb inside the regular container with the rest of the content with:

```
.amundsen-breadcrumb {
  margin-top: -35px;
  margin-bottom: 16px;
}
```
Basically `margin top + margin bottom + height == 0` and the element will have no affect on the flow of elements below it.

danwom: I also don't mind taking this since you're working on other projects.

ttannis: I'll file a ticket for breadcrumbs and put it on the tabledetail page as well.

amundsen_application/api/metadata/v0.py
```diff
@@ -553,3 +553,34 @@ def _log_update_table_tags(*, table_key: str, method: str, tag: str) -> None:
         logging.exception(message)
         payload = jsonify({'msg': message})
         return make_response(payload, HTTPStatus.INTERNAL_SERVER_ERROR)
+
+
+# TODO: Implement real support
+@metadata_blueprint.route('/user', methods=['GET'])
+def get_user() -> Response:
+    try:
+        user_id = get_query_param(request.args, 'user_id')
```
feng-tao: what is user id? and what will the profile page url look like?

cc @ttannis @danwom

feng-tao: Currently all the user node is indexed with email address, it would be good if the FE send the email(convert the user id into email) before sending the request to metadata.

amundsen_application/api/metadata/v0.py
```diff
@@ -553,3 +553,34 @@ def _log_update_table_tags(*, table_key: str, method: str, tag: str) -> None:
...
+            'profile_url': 'https://github.com/lyft/amundsenfrontendlibrary',
+            'user_id': user_id,
+            'github_name': 'lyft',
+            'is_active': True,
+            'manager_name': 'Roald Amundsen',
+            'role_name': 'Software Engineer',
+            'slack_url': 'https://slack.com',
+            'team_name': 'Amundsen Team',
+        }
+        if user_id == 'alumni':
```
feng-tao: why user_id will be alumni?

amundsen_application/api/metadata/v0.py
ttannis: This is whatever value we will use to identify the user. It can be anything, even based on some existing parameter like email. The url will look like `amundsen.lyft.net/user/:user_id`

feng-tao: cool 👍

ttannis: My hope was that we would set an explicit id parameter on the User node even if internally we choose to set it to the email.

amundsen_application/api/metadata/v0.py
ttannis: This was just for testing so that I could type in `amundsen-url/user/alumni` in my dev environment to check the alumni version of the page.

PR #51, opened 2019-03-08
Update search to support multiple resource types

danwom:
- Added Tabs to search results
- Split `executeSearch` into two actions:
-- `searchAll` will search all resource types. The options allow you to search different page indexes for different resources. This is useful for loading up `&selectedTab=users&tabIndex=5`, since you don't necessarily want to fetch page 5 for all tabs.
-- `searchResource` will run a search on a single resource type. This is primarily used for search pagination.
- The URL should always reflect the current state of the search and can be refreshed or shared to maintain state.

### DIFF
--- a/amundsen_application/static/js/components/SearchPage/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/index.tsx
@@ -10,7 +10,8 @@ import { ResourceType, TableResource } from "../common/ResourceListItem/types";
 
 import {
   DashboardSearchResults,
-  ExecuteSearchRequest,
+  SearchAllRequest,
+  SearchResourceRequest,
   TableSearchResults,
   UserSearchResults
 } from "../../ducks/search/types";
@@ -31,7 +32,8 @@ export interface StateFromProps {
 }
 
 export interface DispatchFromProps {
-  executeSearch: (term: string, pageIndex: number) => ExecuteSearchRequest;
+  searchAll: (term: string, pageIndex: number) => SearchAllRequest;
+  searchResource: (resource: ResourceType, term: string, pageIndex: number) => SearchResourceRequest;
   getPopularTables: () => GetPopularTablesRequest;
 }
 
@@ -45,7 +47,8 @@ interface SearchPageState {
 
 class SearchPage extends React.Component<SearchPageProps, SearchPageState> {
   public static defaultProps: SearchPageProps = {
-    executeSearch: () => undefined,
+    searchAll: () => undefined,
+    searchResource: () => undefined,
     getPopularTables: () => undefined,
     searchTerm: '',
     popularTables: [],
@@ -81,7 +84,7 @@ class SearchPage extends React.Component<SearchPageProps, SearchPageState> {
     const pageIndex = params['pageIndex'];
     if (searchTerm && searchTerm.length > 0) {
       const index = pageIndex || '0';
-      this.props.executeSearch(searchTerm, index);
+      this.props.searchAll(searchTerm, index);
     }
   }
 
@@ -112,13 +115,16 @@ class SearchPage extends React.Component<SearchPageProps, SearchPageState> {
   // TODO - Update pagination event to search only the selected tab
   handlePageChange(pageNumber) {
     // subtract 1 : pagination component indexes from 1, while our api is 0-indexed
-    this.updateQueryString(this.props.searchTerm, pageNumber - 1);
+    // this.updateQueryString(this.props.searchTerm, pageNumber - 1);
+
+    this.props.searchResource(ResourceType.table, this.props.searchTerm, pageNumber - 1);
+
   }
 
   updateQueryString(searchTerm, pageIndex) {
     const pathName = `/search?searchTerm=${searchTerm}&pageIndex=${pageIndex}`;
     window.history.pushState({}, '', `${window.location.origin}${pathName}`);
-    this.props.executeSearch(searchTerm, pageIndex);
+    this.props.searchAll(searchTerm, pageIndex);
   }
 
   renderPopularTables = () => {
@@ -168,7 +174,7 @@ class SearchPage extends React.Component<SearchPageProps, SearchPageState> {
 
     return (
       <div className="col-xs-12">
-        <TabsComponent tabs={ tabConfig } defaultTab="tables" />
+        <TabsComponent tabs={ tabConfig } defaultTab="tables" onSelect={(tab)=> {console.log(`Tab: ${tab} selected`)}}/>
       </div>
     );
   };
--- a/amundsen_application/static/js/components/SearchPage/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/index.tsx
@@ -199,6 +199,8 @@ class SearchPage extends React.Component<SearchPageProps, SearchPageState> {
     const startIndex = (RESULTS_PER_PAGE * page_index) + 1;
     const endIndex = RESULTS_PER_PAGE * ( page_index + 1);
 
+
+    // TODO - Move error messages into Tab Component
     // Check no results
     if (total_results === 0 && searchTerm.length > 0) {
       return (
--- a/amundsen_application/static/js/components/SearchPage/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/index.tsx
@@ -9,14 +9,15 @@ import InfoButton from '../common/InfoButton';
 import { ResourceType, TableResource } from "../common/ResourceListItem/types";
 
 import {
-  ExecuteSearchRequest,
   DashboardSearchResults,
+  ExecuteSearchRequest,
   TableSearchResults,
   UserSearchResults
 } from "../../ducks/search/types";
 import { GetPopularTablesRequest } from '../../ducks/popularTables/types';
 // TODO: Use css-modules instead of 'import'
 import './styles.scss';
+import TabsComponent from "../common/Tabs";
 
 const RESULTS_PER_PAGE = 10;
 
@@ -39,6 +40,7 @@ type SearchPageProps = StateFromProps & DispatchFromProps;
 interface SearchPageState {
   pageIndex: number;
   searchTerm: string;
+  // selectedTab: string;
 }
 
 class SearchPage extends React.Component<SearchPageProps, SearchPageState> {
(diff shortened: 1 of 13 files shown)


danwom:
Given this is a comprehensive revamp of all of the search related FE code, I'd want to get feedback from @ttannis on that part before merging.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -59,67 +63,106 @@ class SearchPage extends React.Component<SearchPageProps, SearchPageState> {
...
     const params = qs.parse(window.location.search);
-    const searchTerm = params['searchTerm'];
-    const pageIndex = params['pageIndex'];
+    const { searchTerm, pageIndex, selectedTab} = params;
+
+    this.updateSelectedTab(selectedTab);
     if (searchTerm && searchTerm.length > 0) {
       const index = pageIndex || '0';
-      this.props.executeSearch(searchTerm, index);
+      this.props.searchAll(searchTerm, this.getSearchOptions(index));
```
ttannis: This method relies on the value of `this.state.selectedTab`, which is reset a couple lines up in `updateSelectedTab`. There is no guarantee that by the time we call `getSearchOptions` that the component will have completed whatever black-box logic is involved in `setState` -- it's possible for `this.state.selectedTab` to not yet have been updated to what you intend it to be.  I would suggest explicitly passing the  intended `selectedTab` into `getSearchOptions`.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -59,67 +63,106 @@ class SearchPage extends React.Component<SearchPageProps, SearchPageState> {
...
 
   componentDidMount() {
     this.props.getPopularTables();
 
     const params = qs.parse(window.location.search);
-    const searchTerm = params['searchTerm'];
-    const pageIndex = params['pageIndex'];
+    const { searchTerm, pageIndex, selectedTab} = params;
+
+    this.updateSelectedTab(selectedTab);
```
ttannis: If someone types in an invalid option for `selectedTab`, it looks like the search will still execute but the UI won't update the selected tab in `updateSelectedTab`. What if we set `selectedTab` to default to table when it is not valid.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -59,67 +63,106 @@ class SearchPage extends React.Component<SearchPageProps, SearchPageState> {
...
+      case ResourceType.user:
+        return this.props.users.page_index;
+      case ResourceType.dashboard:
+        return this.props.dashboards.page_index;
+    }
+    return 0;
+  };
+
+  onSearchBarSubmit = (searchTerm: string) => {
+    this.props.searchAll(searchTerm, this.getSearchOptions(0));
```
ttannis: nit: if pageIndex is `0` we could just return some default options, or perhaps the default search options should be baked into `this.props.searchAll()`. That way we don't have to run through the logic in `this.getSearchOptions` just to get the known output `{ dashboardIndex: 0, userIndex: 0, tableIndex: 0}`

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -167,63 +198,69 @@ class SearchPage extends React.Component<SearchPageProps, SearchPageState> {
...
+          onSelect={ this.onTabChange }
+        />
+      </div>
+    );
+  };
+
+
+  // TODO: Hard-coded text strings should be translatable/customizable
+  getTabContent = (results) => {
+    const startIndex = (RESULTS_PER_PAGE * results.page_index) + 1;
```
ttannis: Perhaps I've gotten lost in the diffs, but how does the UI currently handle a search that returns no results for the tab, or if the given index is out of bounds (possible with someone typing in/modifying a url to form their query)

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -59,67 +63,106 @@ class SearchPage extends React.Component<SearchPageProps, SearchPageState> {
...
+  onSearchBarSubmit = (searchTerm: string) => {
+    this.props.searchAll(searchTerm, this.getSearchOptions(0));
+    this.updatePageUrl(searchTerm, this.state.selectedTab,0);
+  };
+
+  onPaginationChange = (pageNumber) => {
+    // subtract 1 : pagination component indexes from 1, while our api is 0-indexed
+    const index = pageNumber - 1;
+
+    this.props.searchResource(this.state.selectedTab, this.props.searchTerm, index);
```
ttannis: Have we tested this? Will the page re-render on pagination change? If I understand correctly, the component will only re-render when it's state changes. Right now the only piece of state is `selectedTab`, which doesnt get updated in this flow.

amundsen_application/static/js/ducks/search/api/v0.ts
```diff
@@ -1,24 +1,36 @@
...
-function transformSearchResults(data: SearchResponse): SearchReducerState {
-  return {
-    searchTerm: data.search_term,
-    dashboards: data.dashboards,
-    tables: data.tables,
-    users: data.users,
-  };
+export function searchAll(action: SearchAllRequest) {
+  const { term, options } = action;
+  let baseUrl = '/api/search/v0';
```
ttannis: nit: can make `'/api/search/v0'` a constant. Not a big deal here, but useful if we had a lot of methods and want to update the base url, we'd just have to do it in one place.

amundsen_application/static/js/ducks/search/api/v0.ts
```diff
@@ -1,24 +1,36 @@
...
-    const data = error.response ? error.response.data : {};
-    return transformSearchResults(data);
-  });
+
+export function searchResource(action: SearchResourceRequest) {
+  const { term, pageIndex, resource } = action;
+  return axios.get(`/api/search/v0/${resource}?query=${term}&page_index=${pageIndex}`)
+    .then((response: AxiosResponse) => {
+      const { data } = response;
+      return {
```
ttannis: I think it makes sense here for this method to return just the resource and the results of that resource. see comment in the `reducer.ts`

amundsen_application/static/js/ducks/search/reducer.ts
```diff
@@ -44,10 +58,26 @@ const initialState: SearchReducerState = {
...
-    case ExecuteSearch.FAILURE:
+    // SearchAll will reset all resources with search results or the initial state
+    case SearchAll.SUCCESS:
+      return {
+        searchTerm: newState.searchTerm,
+        dashboards: newState.dashboards || initialState.dashboards,
+        users: newState.users || initialState.users,
+        tables: newState.tables || initialState.tables,
+      };
+    // SearchResource will set only a single resource and preserves search state for other resources
```
ttannis: Why not have the payload of `SearchResource.SUCCESS` reflect that by returning the resource name and results, and then just updating that one piece of state?

amundsen_application/static/js/components/SearchPage/index.tsx
danwom: I see what you mean. I'll create a separate 'validateTab' method and then use whatever value gets passed back from that as a valid tab.

amundsen_application/static/js/components/SearchPage/index.tsx
danwom: I made each field inside the `options` object optional, but I did not make the 'options' object itself optional. Will update.

amundsen_application/static/js/components/SearchPage/index.tsx
danwom: `searchResource` will update the `searchState` object in the appropriate tab. This will get passed back into the `SearchPage` as props.

Tested an verified.

amundsen_application/static/js/components/SearchPage/index.tsx
danwom: I'll need to modify `createErrorMessage` to accept a custom string for the tab type and call it in this function. I put a TODO for now, but it should be easy to fix.

amundsen_application/static/js/ducks/search/reducer.ts
danwom: I see. I was trying to convert `search_term` to `searchTerm` while passing the rest of the fields along. If I leave `search_term` as is and pass along the response object I can just spread the new object and overwrite just the things that are included.

amundsen_application/static/js/components/SearchPage/index.tsx
ttannis: Today I learned the component will re-render when the store updates its props...and that those props don't *need* to be on the component state. (though correct me if you've found anything that says otherwise). 

Some components I wrote will eventually need some cleanup if you want to be on the lookout for those -- we definitely have some components where I passed values to the component state which I now realize don't need to be on the component state because they are only updated from the store -- and not via any internal component interactions. 😓

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -225,11 +193,35 @@ class SearchPage extends React.Component<SearchPageProps, SearchPageState> {
...
-    const endIndex = RESULTS_PER_PAGE * ( results.page_index + 1);
-    let title =`${startIndex}-${Math.min(endIndex, results.total_results)} of ${results.total_results} results`;
+  getTabContent = (results, tabLabel) => {
+    const { searchTerm } = this.props;
+    const { page_index, total_results } = results;
+    const startIndex = (RESULTS_PER_PAGE * page_index) + 1;
+    const endIndex = RESULTS_PER_PAGE * ( page_index + 1);
+
+    // Check no results
+    if (total_results === 0 && searchTerm.length > 0) {
```
ttannis: nit: I'd be a fan of putting support into the Tab component itself for an `emptySectionMessage`. That way we can have consistent styling as the app scales and we potentially use this concept on other pages.

amundsen_application/static/js/containers/SearchPage/index.tsx
```diff
@@ -9,7 +9,7 @@ import SearchPage, { DispatchFromProps, StateFromProps } from '../../components/
 
 export const mapStateToProps = (state: GlobalState) => {
   return {
-    searchTerm: state.search.searchTerm,
+    searchTerm: state.search.search_term,
```
ttannis: fun fact: there are `camelCase` and `decamelize` libraries that we can consider using in the future to get rid of this discrepancy. More boiler plate but I think at least that logic can be quarantined to `api.ts` folders when we parse the response from python. I'll file a ticket so we don't lose the idea, and can revisit later when feature demands die down.

amundsen_application/static/js/components/SearchPage/index.tsx
danwom: Will leave a TODO and file a ticket.

PR #52, opened 2019-03-13
WIP Bookmark implementation

danwom:


amundsen_application/static/js/ducks/bookmarks/api/v0.ts
```diff
@@ -0,0 +1,39 @@
+import axios, { AxiosResponse, AxiosError } from 'axios';
+
+
+// Add or remove bookmarks for the logged in user
+export function addBookmark(action) {
```
ttannis: Would we not also pass in the user's id? Bookmarking involves two resources, not sure how the metadata service is supposed to know which user to add or remove the bookmark from without the user id.

amundsen_application/static/js/ducks/bookmarks/api/v0.ts
```diff
@@ -0,0 +1,39 @@
...
+  return axios.get(`/api/metadata/v0/bookmarks`)
+    .then((response: AxiosResponse) => {
+      return response.data;
+    })
+
+}
+
+
+// Get other user's bookmarks
+export function getBookmarksForUser(action) {
```
ttannis: Why do we need separate methods for the logged in user and another user? A user is a user so don't we only need a method that takes the user id and returns bookmarks?

amundsen_application/static/js/ducks/bookmarks/api/v0.ts
danwom: For actions that involve updating the logged-in user's data, I think it's safer to add the userId from the authentication layer. Otherwise a user could arbitrarily add or remove bookmarks for another user.

amundsen_application/static/js/ducks/bookmarks/api/v0.ts
danwom: Original thought process was in line with the comment above but I think we can fold `getBookmarks` and `getBookmarksForUser` together. We'd still want to keep separate reducer/saga actions since one will update the `myBookmarks` cache.

PR #53, opened 2019-03-15
changed 'last_updated' to 'last_updated_epoch'

danwom:
- Conditionally hide the timestamp when not present

### DIFF
--- a/amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
+++ b/amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
@@ -22,10 +22,9 @@ class TableListItem extends React.Component<TableListItemProps, {}> {
   render() {
     const { table } = this.props;
 
-    // TODO - default last_updated timestamp is just for testing
-    table.last_updated = table.last_updated || Date.now();
-    const dateTokens = new Date(table.last_updated).toDateString().split(' ');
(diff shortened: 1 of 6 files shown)


amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
```diff
@@ -22,10 +22,9 @@ class TableListItem extends React.Component<TableListItemProps, {}> {
   render() {
     const { table } = this.props;
 
-    // TODO - default last_updated timestamp is just for testing
-    table.last_updated = table.last_updated || Date.now();
-    const dateTokens = new Date(table.last_updated).toDateString().split(' ');
-    const lastUpdated = `${dateTokens[1]} ${dateTokens[2]}, ${dateTokens[3]}`;
+    const hasLastUpdated = table.last_updated_epoch !== 0 && table.last_updated_epoch !== null;
```
ttannis: nit: If we want to check for non-zero and non-null, `const hasLastUpdated = !!table.last_updated_epoch` would also work.

danwom: Once `popular_tables` gets updated with a timestamp it'll just be tables with a timestamp of 0 for non-partitioned tables. (or is this a bug? @jinhyukchang)

PR #56, opened 2019-03-19
Consolidate Containers

ttannis:
Our React components follow the container-component pattern, with containers connected to Redux for dispatching actions to the backend. We're making an organizational decision to not separate out containers into their own directory. 

**What this PR does:**
1. Moved the logic to connect a component to the redux store into the file where the logic for that component is written. Notice we are exporting the connected component as the default, but also export the wrapped component class, as well as the `mapStateToProps` and `mapDispatchToProps` methods. This is because the pattern we will implement for unit testing will be importing these specific chunks for unit testing -- not the entire connected component. 
2. Updated all imports  as necessary.
3. I moved the `Feedback` component out of `common`. It never truly belonged there. I also separated all of the different feedback forms into their own folders.
4. Moved the `OwnerEditor` component into `TableDetail`. 

**What this PR does not do**
1. Clean up our imports. Imports were only modified to match files that were moved around. We still suffer from not having our types defined in a consistent pattern.


amundsen_application/static/js/components/Feedback/FeedbackForm/BugReportFeedbackForm/index.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
```
danwom: I'm beginning to think there are places where we probably don't need a new folder for each subcomponent. What do you think about a structure like:

```
components/Feedback/AbstractFeedbackForm.tsx
components/Feedback/BugReportFeedbackForm.tsx
components/Feedback/FeatureRequestFeedbackForm.tsx
components/Feedback/RatingFeedbackForm.tsx
components/Feedback/FeatureRequestFeedbackForm.tsx
components/Feedback/Feedback.tsx
```

ttannis: We might not need it, but I'd rather be consistent. In addition each component is eventually going to have a `tests` folder nested inside of their directories.

PR #57, opened 2019-03-20
Design Overhaul pt2

danwom:
- Added sizes for standard .btn (32px) and .btn-lg (48px)
- Colors - rework of gray, text, and button colors
- Search Bar - Added placeholder color, focus highlighting, enlarged in wide screens
- Tags - TagInput and TagInfo updated according to mocks

### DIFF
--- a/amundsen_application/static/css/_variables-default.scss
+++ b/amundsen_application/static/css/_variables-default.scss
@@ -3,8 +3,14 @@
  These must be defined here:
 
   $brand-color-1, $brand-color-2, $brand-color-3, $brand-color-4, $brand-color-5
-  $font-family-sans-serif-bold,
-  $font-weight-sans-serif-bold
+
+  $font-family-sans-serif-bold, $font-weight-sans-serif-bold
+
+  $text-dark, $text-medium, $text-light
+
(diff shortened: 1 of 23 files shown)


amundsen_application/static/css/_variables-default.scss
```diff
@@ -26,7 +26,13 @@ $gray-lightest: lighten($gray-base, 95%) !default; // #f2f2f2
 
 // Scaffolding
 $body-bg:    #fff !default;
-$text-color: $gray-dark !default;
+// TODO - replace $text-color with the new $text-light, medium, dark
+
+$text-dark: $gray-dark !default;
+$text-medium: $gray !default;
+$text-light: $gray-lighter !default;
+$text-color: $text-dark !default;
```
danwom: Instead of picking one of the 7 gray colors for our text fields, we should explicitly define our text colors. I like having `$text-dark`, `$text-medium`, and `$text-light`.

`$text-color` is used only by Bootstrap to define the `color` on the `body`, but we could also replace `$text-dark` and use `$text-color`, `$text-medium`, and `$text-light` instead.

danwom:
We can always make additional changes to this in future commits. I think this is a good starting point for improving the look and feel.

PR #59, opened 2019-03-22
Added action logging instrumentation

danwom:
- Added a generic event logger that can be called by the FE
- Added event logging for nav links, tags, table details links, stats, and scrolling
- Added a generic ScrollTracker


amundsen_application/api/log/v0.py
```diff
@@ -0,0 +1,47 @@
...
+    :param value: (Optional)
+    :param source: (Optional)
+    :return:
+    """
+    @action_logging
+    def _log_generic_action(*, command: str, key: str, value: str, source: str) -> None:
+        pass  # pragma: no cover
+
+    try:
+        command = get_query_param(request.args, 'command', '"command" is a required parameter.')
```
jinhyukchang: Is there a case that we don't need key, value, and source?
If we can make it all required, it would be great to have conventional REST API resource URL `/api/log/v0/command/{command}/source/{source}/...`

danwom: I'm still trying to figure out the best way to map out the different values that we need to record. There will definitely be cases that each parameter is not required.

For example we want to log the action `"Click on Announcements"`. 

```
{
    command: 'click',
    key: ??,
    value: 'Announcements',
    source: 'navbar',
}
```
Or

```
{
    command: 'navigation',
    label: 'Announcements',
    value: '/announcements', ??
    source: 'click',
}
```
There are are probably 10s of permutations that make sense. Will follow up with PM to figure out the right fields.

amundsen_application/static/js/ducks/utilMethods.ts
```diff
@@ -25,3 +26,14 @@ export function filterFromObj(initialObj: object, rejectedKeys: string[]): objec
...
+
+export function log(command: string, target_id: string, target_type: string, label?: string, location?: string, index?: number) {
+  const params = { command, target_id, target_type, label, location, index };
+  ['target_type', 'label', 'location', 'index'].forEach((propName) => {
+    if (params[propName] == null || params[propName] === '') {
+      delete params[propName];
+    }
+  });
+  logAction(params);
+}
```
danwom: This originally started with just 4 parameters, but now it has 6. I'm on the fence on whether to just accept a "params" object and validate fields on that. The downside is that everywhere this is used the caller will have to build an object like

```
{ 
  command: "click", 
  target_id: "nav::browse", 
  target_type: "link",
  label: "Browse",
  location: "navbar" 
}
```

danwom:
@jinhyukchang @feng-tao @ttannis This is still a work in progress but I would like some general feedback before I implement the rest of the events.

The general idea is to expose a generic logging API that accepts a set of named parameters that will be flexible enough to track any events we implement in the future. So far we have:
```
Required
  command - legacy name of event
  target_id - unique name for this element the action occurred on
Optional
  target_type - type of element that the action occurred on
  label - display text 
  location - What page or where on the page
  index - Index in list, if applicable 
```

amundsen_application/static/js/ducks/utilMethods.ts
ttannis: I think that going with an object that has a defined interface will be the cleanest way to go. The parameters create a lot of room for error because we will have to remember to call this method with parameters in the right order.

amundsen_application/static/js/components/Tags/TagInfo/index.tsx
```diff
@@ -1,12 +1,15 @@
...
 import { Tag } from '../types';
+import { log } from "../../../ducks/utilMethods";
+
 
 import './styles.scss';
 
 interface TagInfoProps {
   data: Tag;
   compact?: boolean;
+  location: string;
```
ttannis: I think we should look into isolation what is needed for logging away from the properties and behavior of the component as much as possible. Besides perhaps requiring an `id` and hooking up the component to call the `log` method when interacted with, can we try to isolate everything else into some kind of logging utils?

For example is any of this information available from the `event` or the `target`? Or perhaps what if we grabbed `window.location` in the log method?

danwom: This currently isn't `window.location`, but something like "navbar",  "frequent users", or "column details". The page URL might also be a useful thing to log. Maybe we can add something in the generic handler to automatically add the URL.

ttannis: Even so -- beyond giving it the component the ability to call the log method I'm still wondering how we can avoid introducing new properties for things that are not inherent to the component's functional purpose.

Another idea...what if our ids were generated to have all of the extra information that don't relate the the component's functional purpose. Perhaps we could have utility methods to generate ids of a specific pattern, and parse out the information we need in the `log` method.

danwom: I think the analytics events should be highly structured and predictable as someone or some system downstream will ultimately need to query these events. I agree that certain events will need some unique fields that don't really fit into any of the predefined fields, but I don't think the ID is the right place to put this data.

We could try something like the [fieldsObject] in the Google Analytics API listed below.

Screen Shot 2019-04-09 at 11 01 07 AM


Personally I am in favor of killing our custom logger in favor of pushing all events into GA, but that is out of scope for now. As an alternative, we send events to both our custom logger as well as GA. I'd only need to modify the `logAction` function.

ttannis: Regarding predictability, the idea is that the information sent in the actual logging event would still be structured. `id`s would be generated according to a predefined structure. Information would also be parsed out of that `id` via some utility method, and the payload sent with the analytics event would still be the same -- the difference is the component itself would contain no actual logic for the logging, besides calling the log methods. For example why pass a string like "navbar" to the component if we can include it in the `id` and parse it out when we are creating the payload to send with the analytics event.

This is just one idea -- what I'm raising here is the concept of keeping our components clean and decoupled from what we need for logging, and brainstorming how to go about that. I'm wondering how we can avoid adding new properties that are unrelated to the component's purpose just so that we can pass it to the logger, because that approach will not scale well. I believe it could be possible to grab all of the information we need from passing just the DOM element and dealing with it accordingly outside of the component. 

It would be great if all the component had to do was leverage utilities like `logClick(element: HTMLElement)` and `logScroll(window:  Window?)`, and isolated logic that lives outside of the component would handle the rest.

danwom: You're suggesting we take something like this
```
<a href={url}
   target='_blank'
   onClick={() => logAction({
     command: 'click',
     target_id: 'frequent-users',
     target_type: 'link',
     label: fullName,
     location: 'table-details',
   })}
>
```
and convert it to 

```
<a href={url}
   id=`frequent-users:link:${fullName}:table-details`
   target='_blank'
   onClick={logClick(this)}
>

logClick = (target: HTMLElement) => {
  let parts = target.id.split(':')
  logAction({
    command: "click",
    target_id: parts[0],
    target_type: parts[1],
    label: parts[2],
    location: parts[3],
  });
}
```

While this looks cleaner, it isn't not much different from defining a function like `logAction(action, target_id, target_type, label, location)`, with the added requirement that you need to memorize the ID structure. Another approach that many analytics libraries use (and the one I'm most familiar with) is to annotate the element with some analytics attributes.


```
<a
   href="#"
   data-metrics-event-name="commentPageClick"
   data-metrics-comment-id={commentId}
   data-metrics-page-num={currentPage + 1}>
   Next
</a>
```
from: https://github.com/nfl/react-metrics

```
<a href="file.pdf"
   analytics-on="click"
   analytics-event="Download"
   analytics-properties="{ category: 'Content Actions' }">Download</a>
```
from: https://github.com/angulartics/angulartics

danwom:
The more I look into implementing these events, the more I'm convinced we need to migrate everything over to Google Analytics.

amundsen_application/static/js/components/Tags/TagInfo/index.tsx
ttannis: > While this looks cleaner, it isn't not much different from defining a function like `logAction(action, target_id, target_type, label, location)`, with the added requirement that you need to memorize the ID structure. Another approach that many analytics libraries use (and the one I'm most familiar with) is to annotate the element with some analytics attributes.

What I’m suggesting is more like:
```
<a href={url}
   id=generateIdWithParametersThatCantBeInferred()
   target='_blank'
   onClick={logClick(this)}
>

// lives in another file
logClick = (target: HTMLElement) => {
  logAction({
    command: "click" OR we can get this from event.type if passing the event and not just the target,
    target_id: either parsed from or the same as target.id,
    target_type: we can get this from the target.type,
    label: unless I misunderstand what this value is, we can also get this from somewhere in the target?
    location: while not the exact url, we can still parse ‘table_detail` from the url which if not on the target should be on the event,
  });
}
```
Generating and parsing the `id` would be handled by methods that take a payload of a specific interface. Developers would be guided by the TypeScript interface to know what to pass -- no memorization, only the need to know the interface just like everything else. (but this isn't the main point)

What I hope is bit clearer from above is that I think it may be possible to get many of those values by looking into the breadth of data that is automatically provided to us via the browser event APIs. If we don't go that exact route -- though it would be nice to have cleaner single line log methods instead --  the main concept I hope we can figure out is how to avoid adding one-off properties like `location`, or any other log properties in the future, to the component properties so that we can keep our components clean and only concerned with their function.

amundsen_application/api/log/v0.py
```diff
@@ -0,0 +1,60 @@
...
+            value=args.get('value', None)
+        )
+        message = 'success'
+        return make_response(jsonify({'msg': message}), HTTPStatus.OK)
+
+    except Exception as e:
+        message = 'Encountered exception: ' + str(e)
+        logging.exception(message)
+        payload = jsonify({'msg': message})
+        return make_response(payload, HTTPStatus.INTERNAL_SERVER_ERROR)
```
jinhyukchang: Is this error visible to user?

danwom: Right now an error will quietly fail on the FE. Failed analytics requests shouldn't disrupt user interactions. Of course these failures will still be logged and trigger escalation if necessary.

jinhyukchang: 👍

danwom:
@ttannis should be ready for review.

amundsen_application/static/js/components/common/ScrollTracker/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  componentDidMount() {
+    window.addEventListener("scroll", this.onScroll);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener("scroll", this.onScroll)
+  }
+
+  // TODO - Debounce this function
```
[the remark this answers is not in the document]
danwom: How do you feel about including lodash into Amundsen? Would be a great utility add.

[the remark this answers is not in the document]
ttannis: Up for doing that in a separate PR. Main concern is handling that before we deploy this code so we don't start firing large amounts of scroll logging events in production.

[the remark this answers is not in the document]
danwom: I tested the scroll events locally and there shouldn't be any performance concerns at this point. I also clean up the events so they don't stack on top of each other as you continue to use the site.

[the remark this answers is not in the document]
danwom: Alternatively we could just use 'debounce' which was taken from underscore's codebase: https://www.npmjs.com/package/debounce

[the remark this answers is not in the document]
ttannis: I'm not concerned about page performance so much thinking about whether or not we risk bloating our production logs. Out of curiosity what's the general magnitude of events getting fired when exploring a page for let's say a minute?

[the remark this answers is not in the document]
danwom: If the user scrolls all the way to the bottom of the table details page, a total of 4 analytics events will be fired (25, 50, 75, and 100). These events do not get sent multiple times because of the `shift()` on line 44.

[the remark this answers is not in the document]
ttannis: I see. This becomes more of a nit then, doesn't look like we'll blow up our production logs and we can get around to the debouncing when there's spare time.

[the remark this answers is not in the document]
danwom: Added throttling.

danwom:
I've added a default 'location' parameter that takes `window.location.pathname`

amundsen_application/static/js/components/common/ScrollTracker/index.tsx
```diff
@@ -0,0 +1,61 @@
+import * as React from 'react';
+import { throttle } from 'throttle-debounce';
+import { logAction } from "../../../ducks/utilMethods";
```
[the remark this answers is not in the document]
danwom: Thought I caught all of these.

danwom:
The requirements/specs for this task were very vague so I ended up going back and forth on how this worked and what data it collected. I probably should have spent more time fleshing out the implementation details before building it.

PR #60, opened 2019-03-22
Pull changes from #56 into feature/people

ttannis:


ttannis:
@danwom For whatever reason it looks like this considered our last 3 changes. Had to fix some stuff up after the merge but the changes look as expected to me. 

There are 2 `.scss` files with changes where I removed duplicate settings, please confirm if those look good.

amundsen_application/static/js/components/SearchPage/SearchBar/styles.scss
```diff
@@ -11,7 +11,6 @@
       top: 28px;
       left: 24px;
       top: 18px;
```
danwom: Duplicate `top` properties from merge. Should be `28px`.

amundsen_application/static/css/_buttons-default.scss
```diff
@@ -73,10 +73,6 @@
     box-shadow: 0 1px 3px 0 rgba(17, 17, 31, 0.12);
   }
```
danwom: You can actually delete both `.btn-shadow` classes. I added them in because they were in one of Weibke's mocks, and then removed them since they weren't part of the standard buttons that Matt produced.

amundsen_application/static/js/components/Feedback/FeedbackForm/RatingFeedbackForm/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+
+import AbstractFeedbackForm, { DispatchFromProps, StateFromProps } from '../../FeedbackForm';
+
+import { GlobalState } from "../../../../ducks/rootReducer";
+import { submitFeedback, resetFeedback } from '../../../../ducks/feedback/reducer';
+
+export class RatingFeedbackForm extends AbstractFeedbackForm {
+  constructor(props) {
```
verdan: Do we need to call the constructor method of each subclass of `AbstractFeedbackForm` ? If not, we can remove them also, as they are not serving any specific purpose here.

amundsen_application/static/js/components/Feedback/FeedbackForm/RatingFeedbackForm/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+    super(props)
+  }
+
+  renderCustom() {
+    const ratings = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+    const radioButtonSet = ratings.map(rating => (
```
verdan: this functionality could be placed in a separate function like `renderRadioButtonSet`, and be called directly `{ this.renderRadioButtonSet() }`. Would be really handy to write the unit tests.

ttannis: It looks like `renderCustom()` was already split out for this purpose.

amundsen_application/static/js/components/Feedback/FeedbackForm/RatingFeedbackForm/index.tsx
ttannis: 👍 Feel free to remove in a PR.

amundsen_application/static/js/components/Feedback/FeedbackForm/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  protected constructor(props) {
+    super(props);
+
+    this.state = {
+      sendState: this.props.sendState
+    };
+  }
+
+  static getDerivedStateFromProps(nextProps, prevState) {
```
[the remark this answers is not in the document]
ttannis: You're correct. This pattern exists because of my initial misunderstanding of getDerivedStateFromProps and we will have one PR in the future to clean up all the spots where we have this redundancy.

amundsen_application/static/js/components/Feedback/FeedbackForm/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+    const { sendState } = nextProps;
+    return { sendState };
+  }
+
+  submitForm = (event) => {
+    event.preventDefault();
+    const form = document.getElementById(AbstractFeedbackForm.FORM_ID) as HTMLFormElement;
```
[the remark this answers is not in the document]
ttannis: I think you're correct. This PR is a reorganizational one and as a result won't increase the scope to include code cleanup, but please feel free to submit a PR for this.

PR #63, opened 2019-04-02
Update readme

feng-tao:
Add publications


feng-tao:
cc @jinhyukchang @ttannis

PR #81, opened 2019-04-15
Refactor call pattern for switching between clients

danwom:
Created a utility function to abstract away the logic that switches between backend clients and simple API endpoints.

### DIFF
--- a/amundsen_application/api/utils/request_utils.py
+++ b/amundsen_application/api/utils/request_utils.py
@@ -1,6 +1,6 @@
 from typing import Dict
 import requests
-from flask import Response
+
 
 def get_query_param(args: Dict, param: str, error_msg: str = None) -> str:
     value = args.get(param)
@@ -10,16 +10,17 @@ def get_query_param(args: Dict, param: str, error_msg: str = None) -> str:
     return value
 
 
-def make_request_wrapper(method: str, url: str, client: any, headers: any, timeout: int) -> Response:
+# TODO: Define an interface for envoy_client
+def make_request_wrapper(method: str, url: str, client, headers, timeout: int):  # type: ignore
     if client is not None:
-        if method == 'GET':
+        if method == 'DELETE':
+            return client.delete(url, headers=headers, raw_response=True)
+        elif method == 'GET':
             return client.get(url, headers=headers, raw_response=True)
+        elif method == 'POST':
+            return client.post(url, headers=headers, raw_response=True)
         elif method == 'PUT':
             return client.put(url, headers=headers, raw_response=True)
-        elif method == 'DELETE':
-            return client.delete(url, headers=headers, raw_response=True)
-        elif method == 'UPDATE':
-            return client.update(url, headers=headers, raw_response=True)
         else:
             raise Exception('Unsupported method')
     else:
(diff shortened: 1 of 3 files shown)


amundsen_application/api/utils/request_utils.py
```diff
@@ -1,9 +1,36 @@
...
 
 def get_query_param(args: Dict, param: str, error_msg: str = None) -> str:
     value = args.get(param)
     if value is None:
         msg = 'A {0} parameter must be provided'.format(param) if error_msg is not None else error_msg
         raise Exception(msg)
     return value
+
+
+def make_request_wrapper(method: str, url: str, client: any, headers: any, timeout: int) -> Response:
```
danwom: Got any suggestions for a different name?

ttannis:
> _No description provided._

Descriptions will be good in the future, especially if we expect them from others.

ttannis:
@jinhyukchang When we initially wrote some of this code the idea of creating an abstraction for this logic was your proposal if I remember correctly. We didn't discuss it too deeply at the time, was this what you had in mind?

danwom:
@jinhyukchang Mind taking a look?

amundsen_application/api/utils/request_utils.py
ttannis: Maybe just `request_wrapper` if we're going for nouns...maybe `make_request` or `get_response` if we're going for a verb? No strong attachments there though.

amundsen_application/api/utils/request_utils.py
```diff
@@ -7,3 +8,30 @@ def get_query_param(args: Dict, param: str, error_msg: str = None) -> str:
         msg = 'A {0} parameter must be provided'.format(param) if error_msg is not None else error_msg
         raise Exception(msg)
     return value
+
+
+# TODO: Define an interface for envoy_client
+def make_request_wrapper(method: str, url: str, client, headers, timeout: int):  # type: ignore
```
jinhyukchang: Could you add docstring mentioning that it's expecting envoy client and headers?
Also, how about `timeout` to `timeout_sec` (assuming it's second)?

amundsen_application/api/utils/request_utils.py
```diff
@@ -7,3 +8,30 @@ def get_query_param(args: Dict, param: str, error_msg: str = None) -> str:
...
+        if method == 'DELETE':
+            return client.delete(url, headers=headers, raw_response=True)
+        elif method == 'GET':
+            return client.get(url, headers=headers, raw_response=True)
+        elif method == 'POST':
+            return client.post(url, headers=headers, raw_response=True)
+        elif method == 'PUT':
+            return client.put(url, headers=headers, raw_response=True)
+        else:
+            raise Exception('Unsupported method')
```
jinhyukchang: How about including method that has passed?

amundsen_application/api/utils/request_utils.py
```diff
@@ -7,3 +8,30 @@ def get_query_param(args: Dict, param: str, error_msg: str = None) -> str:
...
+            if method == 'DELETE':
+                return s.delete(url, timeout=timeout)
+            elif method == 'GET':
+                return s.get(url, timeout=timeout)
+            elif method == 'POST':
+                return s.post(url, timeout=timeout)
+            elif method == 'PUT':
+                return s.put(url, timeout=timeout)
+            else:
+                raise Exception('Unsupported method')
```
jinhyukchang: Same here. How about including method that has passed?

Issue #79, opened 2019-04-15
Update setup script to add requirements


verdan:
In order to make amundsen work using pip, we need to add the requirements under `install_requires` in setup.py. 

P.S: We don't have that much requirements, so we can avoid dividing the requirements in separate files i.e., test, dev etc, and do something like this: (but yeah that would add extra testing packages for end users)

```
requirements_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'requirements3.txt')
with open(requirements_path) as requirements_file:
    requirements = requirements_file.readlines()

setup(
    ...
    install_requires=requirements,
    ...
)
```

verdan:
cc: @feng-tao

feng-tao:
https://github.com/lyft/amundsenfrontendlibrary/pull/82

feng-tao:
pr is merged. close this issue.

Issue #6, opened 2019-02-10
Install instructions do not end up in working environment 


bolkedebruin:
Hi Guys! Awesome that it is open sourced now! I was just trying things out and didn't get too far yet.

1. The linked metadata service is missing (you probably know this)

I also followed the instructions to a minimal view but got this

```
(venv) Bolkes-MBP:amundsenfrontendlibrary bolke$ python3 amundsen_application/wsgi.py
Registering post_exec action_log entry_point: logging_action_log = amundsen_application.log.action_log_callback:logging_action_log
2019-02-10T21:03:53+0100.686 [INFO] __init__.create_app:43 (54639:MainThread) - Created app with config name amundsen_application.config.LocalConfig
 * Serving Flask app "amundsen_application" (lazy loading)
 * Environment: production
   WARNING: Do not use the development server in a production environment.
   Use a production WSGI server instead.
 * Debug mode: off
2019-02-10T21:03:53+0100.748 [INFO] _internal._log:88 (54639:MainThread) -  * Running on http://0.0.0.0:5000/ (Press CTRL+C to quit)
2019-02-10T21:04:08+0100.256 [ERROR] app.log_exception:1761 (54639:Thread-1) - Exception on / [GET]
Traceback (most recent call last):
  File "/Users/bolke/Documents/dev/amundsenfrontendlibrary/venv/lib/python3.7/site-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/bolke/Documents/dev/amundsenfrontendlibrary/venv/lib/python3.7/site-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/bolke/Documents/dev/amundsenfrontendlibrary/venv/lib/python3.7/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/bolke/Documents/dev/amundsenfrontendlibrary/venv/lib/python3.7/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/Users/bolke/Documents/dev/amundsenfrontendlibrary/venv/lib/python3.7/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/bolke/Documents/dev/amundsenfrontendlibrary/venv/lib/python3.7/site-packages/flask/app.py", line 1799, in dispatch_request
[...]

PR #82, opened 2019-04-16
Update setup.py to include requirment3 file

feng-tao:
Put all the dependency from requirement3 as install requirement. This is from issue https://github.com/lyft/amundsenfrontendlibrary/issues/79

Also bump FE pypi version.

### DIFF
--- a/requirements3.txt
+++ b/requirements3.txt
@@ -46,7 +46,7 @@ flask==1.0.2
 # The wheel project provides a bdist_wheel command for setuptools >= 0.8.0
 # License: MIT
 # Upstream url: https://github.com/pypa/wheel
-wheel==0.31.1
+wheel==0.33.1
 
 # HTTP for Humans
 # License: Apache
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ def build_js() -> None:
 
     try:
         subprocess.check_call(['npm install --only=prod'], cwd=PACKAGE_DIR, shell=True)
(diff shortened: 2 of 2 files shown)


feng-tao:
cc @verdan @ttannis @danwom @jinhyukchang

verdan:
@feng-tao as pointed out here: https://github.com/lyft/amundsenfrontendlibrary/issues/6#issuecomment-483230811 I think it would be better if we can also add the `npm run build` command since we are updating the version. At the moment installing amundsen via pip does not generate the `dist/` and hence no templates are available.

feng-tao:
@verdan good point, updated the pr :)

feng-tao:
thanks @verdan @jinhyukchang for the review.

verdan:
@feng-tao @jinhyukchang I tried to use the new version i.e., `1.0.1` but the dist directory is still not there. I'm trying to use amundsen using `pip` and without the `dist` directory I'm unable to load the templates. Can you please check as what's wrong?

P.S: I also tried to make a package using `python setup.py sdist` and can confirm that the script is now generating the `dist` directory.

feng-tao:
@verdan , I create a pr for this (https://github.com/lyft/amundsenfrontendlibrary/pull/85) also point the issue I found. Please take a look when you have a chance.

ttannis:
@verdan looks like there's some misunderstanding about how the javascript assets are built, which stems from the fact that we left some outdated code in this file. To decouple this from Can you file an issue with the problems you're having or what you're trying to solve, and we can iron things out from there.

PR #90, opened 2019-04-18
Integrate metadata user API

danwom:
- Update User interfaces with new fields
- Minor adjustments to search and profile page layouts (narrower on md and lg screens)


danwom:
*built on top of https://github.com/lyft/amundsenfrontendlibrary/pull/81

danwom:
@ttannis mind taking a look?

feng-tao:
will take a look

amundsen_application/api/v0.py
```diff
@@ -4,18 +4,30 @@
...
 
 @blueprint.route('/auth_user', methods=['GET'])
 def current_user() -> Response:
-    if (app.config['AUTH_USER_METHOD']):
-        user = app.config['AUTH_USER_METHOD'](app)
+    if app.config['AUTH_USER_METHOD']:
+        user_info = app.config['AUTH_USER_METHOD'](app)
     else:
-        user = load_user({'user_id': 'undefined', 'display_name': '*'})
+        user_info = {'email': 'test@test.com', 'user_id': 'test'}
```
feng-tao: should we just throw exception if the user is not authenticated?

danwom: That's a good point.

Maybe this was just something that was there since the beginning and never got cleaned up. I wonder if there are any instances where we want to test something without going through auth.

danwom:
@feng-tao currently there are no links to view a user's page. The only way to get to this route is to manually type in `/user/user_id` into the browser. The user will see whatever data comes back from metadata service, but I'm not concerned about this use case.

amundsen_application/api/v0.py
ttannis: The default open source version of Amundsen doesn't use auth, and there's a change other's wont have any auth configured in their development environments. 

Throwing an exception is fine, but if we remove this line we'd need to make sure the pieces of the application that leverage this information will work as expected if `email` and `user_id` are missing -- I found it easier to add in the required `User` fields here at the start to avoid chasing down potential bugs, knowing that in reality people would configure auth for a real use case.

amundsen_application/api/v0.py
```diff
@@ -4,18 +4,30 @@
...
 
 blueprint = Blueprint('api', __name__, url_prefix='/api')
 
 
 @blueprint.route('/auth_user', methods=['GET'])
 def current_user() -> Response:
-    if (app.config['AUTH_USER_METHOD']):
-        user = app.config['AUTH_USER_METHOD'](app)
+    if app.config['AUTH_USER_METHOD']:
+        user_info = app.config['AUTH_USER_METHOD'](app)
```
ttannis: I'm concerned about changing the return type of a config method for the sake of a single feature, unless the feature can't be implemented at all without the change. It's not conducive to maintaining backwards compatibility, but I recognize that we can get away with it in this case due to the nature of how this method is currently used.

I am also concerned about the change in our internal system that this change would imposes on an unrelated feature. I believe we should discourage the introduction of changes that affect unrelated features, especially since this method could have been written without needing to impose that change.

amundsen_application/static/js/components/NavBar/tests/index.spec.tsx
```diff
@@ -40,8 +40,19 @@ describe('NavBar', () => {
     beforeEach(() => {
         props = {
           loggedInUser:  {
+            email: 'test@test.com',
```
ttannis: nit: You can consider putting this in `fixtures/globalState` so we can ruse it if we need `loggedInUser` test data anywhere else in the future.

amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
```diff
@@ -89,7 +89,7 @@ class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
   render() {
     const subTextClass = `subtext ${this.state.optionalSubTextClass}`;
     return (
-      <div id="search-bar" className="col-xs-12 col-md-offset-1 col-md-10">
```
ttannis: Curious what the result of removing these styles were, is the layout still consistent?

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/tests/index.spec.tsx
```diff
@@ -49,22 +49,23 @@ describe('TableListItem', () => {
...
-          subject.instance().getDateLabel = jest.fn(() => 'Mar 28, 2019')
+          subject.instance().getDateLabel = jest.fn(() => 'Mar 28, 2019');
           props.table.last_updated_epoch = 1553829681;
           subject.setProps(props);
           expect(subject.find('.content').children().at(1).children().at(1).text()).toEqual('Mar 28, 2019');
         });
     });
 
+    /*
+    Note: Jest will convert date to UTC, expect to see different strings for an epoch value in the tests vs UI
```
ttannis: As discussed offline, I'll look into this. Feel free to comment out any tests involving date/time that block development.

amundsen_application/api/v0.py
danwom: Changing this back to return a User object.

amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
danwom: These classes were moved to the parent element in Search Page line 254 so the layout remains the same.

amundsen_application/api/metadata/v0.py
```diff
@@ -516,7 +518,27 @@ def _log_update_table_tags(*, table_key: str, method: str, tag: str) -> None:
...
+
+    if status_code == HTTPStatus.OK:
+        message = 'Success'
+    else:
+        message = 'Encountered error: failed to fetch user with user_id: {0}'.format(user_id)
+        logging.error(message)
+
+    payload = {
+        'msg': message,
+        'user': dump_user(load_user(response.json())),
```
jinhyukchang: What would happen here if request failed?

amundsen_application/api/v0.py
```diff
@@ -4,18 +4,30 @@
 from flask import current_app as app
 from flask.blueprints import Blueprint
 
+from amundsen_application.api.metadata.v0 import USER_ENDPOINT
+from amundsen_application.api.utils.request_utils import request_wrapper
 from amundsen_application.models.user import load_user
 
+REQUEST_SESSION_TIMEOUT = 10
```
jinhyukchang: How about adding unit here? `REQUEST_SESSION_TIMEOUT_SEC`

Also, 10 sec on this call seems very high considering p95 latency is under 300ms. Are we using 10 sec for most of metadata service call?

amundsen_application/api/v0.py
```diff
@@ -4,18 +4,30 @@
...
+    url = '{0}{1}/{2}'.format(app.config['METADATASERVICE_BASE'], USER_ENDPOINT, user_info['user_id'])
+
+    response = request_wrapper(method='GET',
+                               url=url,
+                               client=app.config['METADATASERVICE_REQUEST_CLIENT'],
+                               headers=app.config['METADATASERVICE_REQUEST_HEADERS'],
+                               timeout_sec=REQUEST_SESSION_TIMEOUT)
 
-    return user.to_json()
+    return load_user(response.json()).to_json()
```
jinhyukchang: Could you check status code?

amundsen_application/models/user.py
```diff
@@ -13,76 +13,68 @@
...
 
 
 class UserSchema(Schema):
-    first_name = fields.Str(allow_none=True)
-    last_name = fields.Str(allow_none=True)
-    email = fields.Str(allow_none=True)
-    display_name = fields.Str(required=True)
-    profile_url = fields.Str(allow_none=True)
-
+    email = fields.Str()
```
jinhyukchang: Why removing `allow_none=True`?

amundsen_application/api/metadata/v0.py
danwom: Forgot to wrap this in try-catch.

amundsen_application/api/v0.py
danwom: I can reduce this to something a bit more reasonable like 3s, but I think the 10s value originates from data preview which can take a long time. It makes sense to configure these separately.

amundsen_application/api/v0.py
danwom: Will wrap in try-catch

amundsen_application/models/user.py
danwom: Removing `allow_none=True` didn't seem to change anything. I'm actually not sure what these `fields.Str()` do or if we even need them. It seems like the `UserSchema` class is just used to preprocess/validate a `Dict` before passing it to the `User` class.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -173,7 +171,7 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
     ];
 
     return (
-      <div className="col-xs-12 col-md-offset-1 col-md-10">
+      <div>
```
danwom: @ttannis originally removed this div but the test for `SearchPage › renderSearchResults › renders TabsComponent with correct props` failed. It looks like there is a difference between `content.find(TabsComponenet).props()` and `content.props()` where content is `<TabsComponent>`

amundsen_application/api/v0.py
jinhyukchang: Yeah, preview is an outlier and I believe we configured it as 2 minutes.

amundsen_application/static/js/components/SearchPage/index.tsx
ttannis: This is expected behavior according to Enzyme: https://github.com/airbnb/enzyme/issues/7

If I'm understanding that issue correctly, we might have to use `.props` to access the props directly instead of the method `.props()` which will return whatever information is associated with the root node of Enzyme's representation of this component?

PR #91, opened 2019-04-19
Add Elasticsearch service

jhettler:
Add Elasticsearch to docker-compose. Users are able to get demo with search working.

PR #92, opened 2019-04-22
Update neo4j docker entry to include the network info

feng-tao:
As pointed out by @amweiss, the network info is removed from neo4j container in https://github.com/lyft/amundsenfrontendlibrary/pull/91. We should add it back to make the neo4j container accessible by other container(metadata).


feng-tao:
PTAL @amweiss, @jhettler , @jinhyukchang, @ttannis, @danwom

PR #98, opened 2019-04-24
Implement Custom Typography

danwom:
Revamp of typography across the entire app
- Removed dependency on bootstrap typography
- Defined new variables for header and body font family/weight
- Defined a set of predefined classes such as `title-1`, `title-2`, `body-1`, etc.
- Replaced ad hoc font styles with predefined classes whenever possible.


danwom:
For reference these are the type faces from our design team. 

OSS type

amundsen_application/static/css/_buttons-default.scss
```diff
@@ -3,8 +3,8 @@
 
   &.btn-primary,
   &.btn-default {
-    font-family: $font-family-sans-serif-bold;
-    font-weight: $font-weight-sans-serif-bold;
+    font-family: $font-family-header;
+    font-weight: $font-weight-header-regular;
```
danwom: `$font-family-header` and `$font-weight-header-regular` like this means I've just done a global find and replace. My intent is to replace as *most* of these with one of the predefined typeography classes.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
```diff
@@ -35,20 +35,20 @@ class TableListItem extends React.Component<TableListItemProps, {}> {
...
                 {/*<label> </label>*/}
               {/*</div>*/}
             {/*</div>*/}
             {
               hasLastUpdated &&
               <div className="hidden-xs col-sm-3 col-md-2">
-                <div id="secondary-title" className="secondary-title">Latest Data</div>
-                <div id="secondary-description" className="description truncated">
+                <div id="secondary-title" className="subtitle-3">Latest Data</div>
+                <div id="secondary-description" className="body-3 truncated">
```
danwom: This is what the pattern looks like in practice. I've deleted the `main-title`, `description`, and `secondary-title` classes.

danwom:
@ttannis do you mind taking a look at the pattern before I apply it to the rest of the codebase? 

FYI: @verdan

amundsen_application/static/css/_buttons-default.scss
ttannis: Not sure what you mean, is this change intended or not? Can you clarify a bit more?

amundsen_application/static/css/_fonts-default.scss
```diff
@@ -1 +1,36 @@
-// TODO - Import Open Sans or other default font
+@import 'variables';
```
ttannis: nit: indentation in this file is off.

amundsen_application/static/css/_typography-custom.scss
```diff
@@ -0,0 +1 @@
+// This file is intentionally left blank and should be overwritten by in the build process.
```
ttannis: nit: I'm in favor of a more explicit explanation like `...should be used to add new typography or overwrite defaults from _typography-default.scss`. For example if someone wants to fork our repo and override files that way, I find instruction of having some "build process" confusing.

amundsen_application/static/css/_typography-default.scss
```diff
@@ -0,0 +1,92 @@
+@import 'variables';
+
+// Headers use Roboto
```
ttannis: Curious, why the mixing of fonts for headers vs the rest of the content?

amundsen_application/static/js/components/common/ResourceListItem/styles.scss
```diff
@@ -20,32 +20,5 @@
...
-    }
-
-    .secondary-title {
-      color: $text-medium;
-      font-family: $font-family-sans-serif-bold;
-      font-weight: $font-weight-sans-serif-bold;
-    }
-
-    .description {
-      color: $text-medium;
```
ttannis: Looks like there was some styling lost here that I don't see covered elsewhere.

amundsen_application/static/css/_buttons-default.scss
danwom: We don't really have a single "bold font" and a single "regular font" anymore. We have "Roboto" for headers and "Open Sans" for everything else. Both of those fonts have variants at different weights (400, 500, 600, 700) that use different files ("-Regular", "-Semibold", "-Bold").

It didn't make sense to use the typography system builtin with Bootstrap, so I did a global find and replace for `$font-family-sans-serif-bold -> $font-family-header` and `$font-weight-sans-serif-bold -> $font-weight-header-regular` just so it would compile. I plan to revisit these rules and make adjustments accordingly.

amundsen_application/static/css/_typography-default.scss
danwom: This was a design decision to use Roboto for headers. *It is what it is.*

The bigger question is when to use, say, an `h3` vs using a `title-1`

amundsen_application/static/js/components/common/ResourceListItem/styles.scss
danwom: Changed 'body-3' to `body-3-secondary' with the medium font color.

bolkedebruin:
I strongly suggest including *removing* the glyphicons* fonts, as they are under a restrictive License. See: https://www.glyphicons.com/license/

The same issue exists in Superset btw and was caught at Apache.

danwom:
@bolkedebruin I believe Glyphicons were originally added as they shipped with Bootstrap 3 by default. We're only using them in a couple places so we can easily remove.

bolkedebruin:
Gotcha. Glyphicons is vague on this, as bootstrap’s license is Apache 2. But it’s unclear how much due diligence they did and if they were able to relicense.

amundsen_application/static/js/components/TableDetail/DataPreviewButton/styles.scss
```diff
@@ -7,8 +7,9 @@
     min-height: 90%;
 
     .modal-title {
-      font-family: $font-family-sans-serif-bold;
-      font-weight: $font-weight-sans-serif-bold;
+      font-size: 20px;
+      font-family: $font-family-header;
+      font-weight: $font-weight-header-bold;
```
danwom: Seems like the `<Modal.Title>` component automatically renders as an `h4` so I am manually defining what would otherwise be an `h3`.

danwom:
Ready for review

PR #99, opened 2019-04-24
Increase Application Test Coverage

ttannis:
This PR adds unit test coverage for the following components: `AnnouncementPage`,`BrowsePage`, `Footer`, `NavBar`, `NotFoundPage` and `ProfilePage`.  

`Footer` will require more effort to cover in the future, I ran into problems mocking the `moment-timezone` library.

image

### DIFF
--- a/amundsen_application/static/js/components/AnnouncementPage/index.tsx
+++ b/amundsen_application/static/js/components/AnnouncementPage/index.tsx
@@ -26,7 +26,7 @@ export interface DispatchFromProps {
   announcementsGet: () => AnnouncementsGetRequest;
 }
 
-type AnnouncementPageProps = StateFromProps & DispatchFromProps;
+export type AnnouncementPageProps = StateFromProps & DispatchFromProps;
 
 export class AnnouncementPage extends React.Component<AnnouncementPageProps, AnnouncementPageState> {
   constructor(props) {
@@ -37,11 +37,6 @@ export class AnnouncementPage extends React.Component<AnnouncementPageProps, Ann
     };
   }
 
-  static getDerivedStateFromProps(nextProps, prevState) {
-    const { posts } = nextProps;
-    return { posts };
-  }
-
   componentDidMount() {
     this.props.announcementsGet();
   }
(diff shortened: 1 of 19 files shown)


feng-tao:
We want to integrate with test coverage module for FE repo.

amundsen_application/static/js/components/AnnouncementPage/index.tsx
```diff
@@ -50,32 +45,34 @@ export class AnnouncementPage extends React.Component<AnnouncementPageProps, Ann
     return (
       <div key={`post:${postIndex}`} className='post-container'>
         <div className='post-header'>
-          <div className='post-title'>{post.title}</div>
-          <div className='post-date'>{post.date}</div>
+          <div id='post-title' className='post-title'>{post.title}</div>
+          <div id='post-date' className='post-date'>{post.date}</div>
         </div>
-        <div className='post-content'>
+        <div id='post-content' className='post-content'>
```
danwom: For repeated elements we should try to use unique IDs. Maybe `id={'post-content-${postIndex}'}`

ttannis:
> We may want to integrate with test coverage module for FE repo.

@feng-tao can you clarify what you mean by that?

amundsen_application/static/js/components/Footer/tests/index.spec.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react';
+import moment from 'moment-timezone';
+import { mount } from 'enzyme';
+
+import { Footer, FooterProps, mapDispatchToProps, mapStateToProps } from '../';
+
+import globalState from 'fixtures/globalState';
+
+/* TODO: Issues with mocking moment-timezone, all tests are failing
```
danwom: I believe `Jest` will default to using UTC while the tests cases assumes that `moment.local` will use "PT".

You could add a timezone to `config.ts` and then use it for both the component and the tests. As a side effect, users will see a single timezone for a given deployment, but I think that is fine to say data was indexed relative to server time.

feng-tao:
It would be good to integrate with https://github.com/marketplace/codecov so that every pr would show whether it increases or decreases coverage.

ttannis:
> It would be good to integrate with https://github.com/marketplace/codecov so that every pr would show whether it increases or decreases coverage.

Awesome! I'll take a look and see if I can add that soon.

amundsen_application/static/js/components/Footer/tests/index.spec.tsx
ttannis: The problem I had was beyond mocking the timezone itself, I was ideally trying to mock the object and methods we use because we don't actually need to execute moment's code. As discussed offline I'll revisit this as a separate task given that it won't affect our code coverage goal until we attempt to hit perhaps 90%+

amundsen_application/static/js/components/AnnouncementPage/index.tsx
ttannis: Also going to move away from finding elements by id, it will eventually cause headaches for sure and is not a best practice.  

I'm still playing around with what's best, we can always drill down normally like `subject.find('div').find('div').find(...)`, but that can also get overwhelming. My current thinking is that this is also why some folks break their components down into so many sub-components to make testing easier. 

I've been trying to do this without refactoring components, but starting to feel worried about introducing any bad test practices. I think I will attempt to tackle `SearchPage` and it's nested components using as many best practices as I'm aware of and see how we feel about that.

PR #106, opened 2019-04-26
Added a 'is_view' and 'Table View' indicator on the table details page

danwom:


danwom:
CC: @youngyjd

youngyjd:
💯 awesome!

amundsen_application/static/js/components/common/Flag/styles.scss
```diff
@@ -1,6 +1,7 @@
 @import 'variables';
 
 .flag {
-  margin-right: 4px;
-  margin-left: 4px;
+  display: inline-block;
```
ttannis: I wonder if these changes affect how the `Flag` was intended to be laid out on the `ProfilePage`. Can we either:
1. Verify the use of `Flag` on `ProfilePage` still renders according to the design with these styles OR
2. Override these styles for `Flag` on `TableDetail` page only, and then revisit #1 later to consolidate styles if we can.

danwom: The profile page currently overrides these styles and still renders correctly.

PR #111, opened 2019-04-29
Update README.md

feng-tao:
update the stale slack link

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Please note that the mock images only served as demonstration purpose.
 ## Get Involved in the Community
 
 Want help or want to help? 
-Join our [slack channel](https://bit.ly/2FVq37z) and our [mailing list](https://groups.google.com/forum/#!forum/amundsen-dev)
+Join our [slack channel](https://github.com/lyft/amundsenfrontendlibrary#amundsen) and our [mailing list](https://groups.google.com/forum/#!forum/amundsen-dev)
 
 ## Getting started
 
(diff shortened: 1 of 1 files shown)


feng-tao:
cc @danwom @jinhyukchang @ttannis

jornh:
LGTM - one thought though: If the Slack token has plans to go stale more often it would be more DRY (don't-repeat-yourself) to just link from that section to the badges section at https://github.com/lyft/amundsenfrontendlibrary#amundsen .. maybe?

feng-tao:
Good idea @jornh :) PR updated :)

README.md
```diff
@@ -44,7 +44,7 @@ Please note that the mock images only served as demonstration purpose.
 ## Get Involved in the Community
 
 Want help or want to help? 
-Join our [slack channel](https://bit.ly/2tFsN1X) and our [mailing list](https://groups.google.com/forum/#!forum/amundsen-dev)
+Join our [slack channel](https://github.com/lyft/amundsenfrontendlibrary#amundsen) and our [mailing list](https://groups.google.com/forum/#!forum/amundsen-dev)
```
ttannis: I think it's misleading that a link titled `slack channel` links back to our github page. I support the url only being in one spot though, so I think here we should give clear instruction like:
`Use the button in our [header](https://github.com/lyft/amundsenfrontendlibrary#amundsen) to join our slack channel...`

feng-tao: updated

PR #112, opened 2019-04-29
Fix navbar and footer state bugs

danwom:
The navbar and footer components were incorrectly referencing `state` for values that only existed in `props`.

### DIFF
--- a/amundsen_application/static/js/components/Footer/index.tsx
+++ b/amundsen_application/static/js/components/Footer/index.tsx
@@ -19,17 +19,9 @@ interface DispatchFromProps {
 
 export type FooterProps = StateFromProps & DispatchFromProps;
 
-interface FooterState {
-  lastIndexed: number;
-}
-
-export class Footer extends React.Component<FooterProps, FooterState> {
+export class Footer extends React.Component<FooterProps> {
   constructor(props) {
     super(props);
-
-    this.state = {
-      lastIndexed: this.props.lastIndexed,
-    };
   }
 
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


amundsen_application/static/js/components/Footer/index.tsx
```diff
@@ -19,17 +19,9 @@ interface DispatchFromProps {
 
 export type FooterProps = StateFromProps & DispatchFromProps;
```
danwom: Does the interface `StateFromProps` still make sense here? I tried renaming to just `props` but that didn't seem right.

ttannis: I took these terms from some other docs/tutorials and I've seen many other projects use these terms. Not sure if it's a "standard" but common enough that I'd want to keep it unless we invest dedicated thought into it. 

I think the intent was to use the name to make a connection with `mapStateToProps` and `mapDispatchToProps`, and so these interfaces got named `StateFromProps` and `DispatchFromProps`....but that begs the question why not call them `StateToProps` and `DispatchToProps` instead of using "from". 

I'm onboard if anyone does think of a better naming standard though, but a mass name-change PR is my preference as opposed to doing it right now.

amundsen_application/static/js/components/NavBar/index.tsx
```diff
@@ -32,10 +32,6 @@ interface NavBarState {
 export class NavBar extends React.Component<NavBarProps, NavBarState> {
```
ttannis: We need to remove the state.

PR #118, opened 2019-04-30
Unit test coverage for SearchPage & nested components

ttannis:
This PR further increases code coverage. It intentionally introduces some new patterns (previous tests will be updated eventually) and has a few relevant updates to configs and docs.

**Summary of Changes**
1. Increase coverages thresholds. 
2. Add unit tests for `SearchPage`, `SearchBar`, and `SearchList`.
image
3. Updated `globalState` test fixture with more dummy data for tests.
4. Updated testing section of the developer guide, and added a script to run with no coverage when developing. This is helpful for being able to run a subset of tests during development.
5. Created a doc with some recommended practices. Hopefully it evolves as others contribute and we gain more insight into the different libraries we use: https://github.com/lyft/amundsenfrontendlibrary/blob/0859cdd9811ef4f57a693dfc74a8b5a84e70923f/docs/recommended_practices.md


ttannis:
> The `constants` pattern might enable us to implement internationalization in the future.

Yes, exactly! We need to keep isolating all hardcoded strings and provide the capability to override them with a mapping. That will get us most of the way there and we'll ultimately need a tiny bit of help from design because internationalized designs need to assume text will be anywhere between between really short or really long. 😺

PR #119, opened 2019-05-01
Adds the PR template for Amundsen frontend

verdan:
Mostly copied from Apache Airflow, and using the same in metadata repository.

Fixes: https://github.com/lyft/amundsenmetadatalibrary/issues/16

cc: @danwom @feng-tao


feng-tao:
@verdan thanks. Could you add one for databuilder repo as well?

ttannis:
@verdan thanks for adding this! I'll have some updates to this later to specify what will be expected of changes to the application source (e.g. instruction about docstrings only applies to python code, and there's also a few other commands to add for testing)

PR #135, opened 2019-05-05
Docs: Add logo config example

jornh:
Verbatim copy of a Slack answer given by @danwom (better give credit where it’s due 🙂)

Having a few specific examples to go from in config docs always give you a good start

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -13,3 +13,11 @@ Certain features of the React application import variables from an [AppConfig](h
 ### Custom Fonts & Styles
 Fonts and css variables can be customized by modifying [fonts-custom.scss](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/css/_fonts-custom.scss) and
 [variables-custom.scss](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/css/_variables-custom.scss).
+
+### Examples
+
+#### 1) To add a custom logo you need to do the following:
+
+1. Add your logo to the folder in `amundsen_application/static/images/`
+2. Add the config for `logoPath`. For example Lyft uses the value  `"/static/images/lyft-logo.svg"`
+3. Rebuild/redeploy. The `npm run build` step will rebuild scripts file to reference the new logo image, and the `python3 setup.py install` step will updated deployed files and make the new image available on the server.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -10,14 +10,12 @@ For more information on how the configuration is being loaded and used, please r
 ### Application Config
 Certain features of the React application import variables from an [AppConfig](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/config/config.ts#L5) object. The configuration can be customized by modifying [config-custom.ts](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/config/config-custom.ts).
 
-### Custom Fonts & Styles
-Fonts and css variables can be customized by modifying [fonts-custom.scss](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/css/_fonts-custom.scss) and
-[variables-custom.scss](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/css/_variables-custom.scss).
-
-### Examples
-
-#### 1) To add a custom logo you need to do the following:
+#### Example: Add a custom logo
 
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -13,3 +13,11 @@ Certain features of the React application import variables from an [AppConfig](h
 ### Custom Fonts & Styles
 Fonts and css variables can be customized by modifying [fonts-custom.scss](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/css/_fonts-custom.scss) and
 [variables-custom.scss](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/css/_variables-custom.scss).
+
+### Examples
```
ttannis: Given that these example are related to the Application Config, how about nesting these examples under the `Application Config` section above?

docs/configuration.md
```diff
@@ -13,3 +13,11 @@ Certain features of the React application import variables from an [AppConfig](h
...
+
+#### 1) To add a custom logo you need to do the following:
```
ttannis: nit: `Add a custom logo`. Succinct section titles can make doc easier to scan for folks looking to find a single specific piece of info.

docs/configuration.md
```diff
@@ -13,3 +13,11 @@ Certain features of the React application import variables from an [AppConfig](h
...
+
+1. Add your logo to the folder in `amundsen_application/static/images/`
+2. Add the config for `logoPath`. For example Lyft uses the value  `"/static/images/lyft-logo.svg"`
```
ttannis: A more explicit direction about which file to edit  would also be helpful. Perhaps:
"Set the the `logoPath` key on the `AppConfigCustom` object in [custom-default.ts](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/config/config-custom.ts). For example Lyft uses the value  `/static/images/lyft-logo.svg`"

docs/configuration.md
jornh: OK, my original intention was to have the React example(s?) [where I now think one example may actually be sufficient] in one section covering all of React, as the tricky part really is the rebuilding steps which is common across AppConfig and Font/Styling.

Anyway I nested it, and collapsed titles. I think that can also work!

docs/configuration.md
jornh: Agree, will fix

docs/configuration.md
jornh: Well it worked for the first test user who received the directions 🙂  - also being more explicit about location adds duplication/redundancy between code and doc. 

I think having the link to the repo location - now right above - is a good balance.

jornh:
Ready for review again!

docs/configuration.md
```diff
@@ -10,6 +10,12 @@ For more information on how the configuration is being loaded and used, please r
 ### Application Config
 Certain features of the React application import variables from an [AppConfig](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/config/config.ts#L5) object. The configuration can be customized by modifying [config-custom.ts](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/config/config-custom.ts).
 
+#### Example: Add a custom logo
+
+1. Add your logo to the folder in `amundsen_application/static/images/`
+2. Set the the `logoPath` key on the `AppConfigCustom` object in [config-custom.ts](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/config/config-custom.ts). For example Lyft uses the value `"/static/images/lyft-logo.svg"`
+3. Rebuild/redeploy. The `npm run build` step will rebuild scripts file to reference the new logo image, and the `python3 setup.py install` step will updated deployed files and make the new image available on the server.
```
danwom: I know I originally wrote this in slack, but I wonder if splitting step 3 into two steps (rebuild and deploy) makes more sense. I think it works either way, so I'll leave it to you.

ttannis: @jornh to your note `...the tricky part really is the rebuilding steps which is common across AppConfig and Font/Styling.` Feel free to also take Step 3 from this example and add it either at the beginning or end of the whole "React Application" section, since as you've pointed out it's a common common step and we won't want to write this exact same step each time for every example we add. Fine to merge as is though until we add more examples. Thanks!

jornh: Well, you contributed first. I just carried it a few steps 😉 

I was thinking the same with a bit more splitting. Anyway, important part is content is there... we can keep bikeshedding forever.

 Please merge at will!

jornh: Merge away!

PR #144, opened 2019-05-09
Update User Model

ttannis:
### Summary of Changes
Based on our current production ready features, all User objects must have a `display_name`. We also have a future feature that will require `user_id`. This PR modifies some logic in the User model to ensure that those features work:
1. `display_name` and `user_id` are not allowed to be `None` on the User object.
2. `email` is required because it is the fallback for the two values that are not allowed to be `None`.
3. Added a helper method to check for `None` and empty string equivalents.
4. Updated the ValidationError messages to provide more helpful information about what values our implementation expects.

This PR will the current case where the NavBar Avatar is not rendering correctly because the `display_name` was allowed to be an empty string.

### Tests

No tests were added because the main flow of the User model is already 83% covered with existing tests.

### Documentation

None. 

### CheckList

### DIFF
--- a/amundsen_application/models/user.py
+++ b/amundsen_application/models/user.py
@@ -1,4 +1,4 @@
-from typing import Dict
+from typing import Dict, Optional
 
 from marshmallow import Schema, fields, pre_load, post_load, validates_schema, ValidationError
 
@@ -50,8 +50,8 @@ class User:
 
 
 class UserSchema(Schema):
-    display_name = fields.Str(allow_none=True)
-    email = fields.Str(allow_none=True)
+    display_name = fields.Str(allow_none=False)
+    email = fields.Str(required=True)
     employee_type = fields.Str(allow_none=True)
     first_name = fields.Str(allow_none=True)
     full_name = fields.Str(allow_none=True)
@@ -63,19 +63,19 @@ class UserSchema(Schema):
     role_name = fields.Str(allow_none=True)
     slack_id = fields.Str(allow_none=True)
     team_name = fields.Str(allow_none=True)
-    user_id = fields.Str(required=True)
+    user_id = fields.Str(allow_none=False)
 
     @pre_load
     def preprocess_data(self, data: Dict) -> Dict:
-        if not data.get('user_id', None):
-            data['user_id'] = data.get('email', None)
+        if _str_no_value(data.get('user_id')):
+            data['user_id'] = data.get('email')
 
-        if not data.get('profile_url', None):
+        if _str_no_value(data.get('profile_url')):
             data['profile_url'] = ''
             if app.config['GET_PROFILE_URL']:
                 data['profile_url'] = app.config['GET_PROFILE_URL'](data['user_id'])
 
-        if not data.get('display_name', None):
+        if _str_no_value(data.get('display_name')):
             data['display_name'] = data.get('full_name', data.get('email'))
 
         return data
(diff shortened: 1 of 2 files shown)


amundsen_application/models/user.py
```diff
@@ -86,11 +86,20 @@ def make_user(self, data: Dict) -> User:
...
+        if _str_no_value(data.get('display_name')):
+            raise ValidationError('"display_name", "full_name", or "email" must be provided')
 
-        if not data.get('user_id', None):
-            raise ValidationError('"user_id" must be provided')
+        if _str_no_value(data.get('user_id')):
+            raise ValidationError('"user_id" or "email" must be provided')
+
+
+def _str_no_value(s: Optional[str]) -> bool:
```
feng-tao: not sure if you need a function for this. Would `not s` serve the same purpose?

amundsen_application/models/user.py
```diff
@@ -86,11 +86,20 @@ def make_user(self, data: Dict) -> User:
 
     @validates_schema
     def validate_user(self, data: Dict) -> None:
-        if not data.get('display_name', None):
-            raise ValidationError('"display_name" must be provided')
+        if _str_no_value(data.get('display_name')):
```
feng-tao: if not data.get('display_name') ?

amundsen_application/models/user.py
feng-tao: I see, so there is a case like s = ' '?

ttannis: Yes, we were getting and empty string for the `display_name`, and the previous code made it such that it accepted the empty string instead of trying to assign `full_name` or `email` as a fallback.

verdan:
nit: The code is fine I guess, however I am not a big fan of having method names starting with underscore outside a class for no reason. If you think this could be used anywhere else, why not put that in a utility or helper class and expose it as a static method. Or maybe extend Schema with your own custom validator methods if you think you'll need some.

ttannis:
> nit: The code is fine I guess, however I am not a big fan of having method names starting with underscore outside a class for no reason. If you think this could be used anywhere else, why not put that in a utility or helper class and expose it as a static method. Or maybe extend Schema with your own custom validator methods if you think you'll need some.

Fair. I considered making a utility, but opted to only go that route when we have another use case. I can add this as private method on the class.

amundsen_application/models/user.py
danwom: I was playing with the python interpreter and `not ''` returns `True`. I'm curious why the if block on line 78 does not run.
```if not data.get('display_name', None):```

Comment (feng-tao) [reply amundsen_application/models/user.py]: it should run, my understanding is that the display_name could be empty string with a space? In that case it will be true when evaluating.

Comment (danwom) [reply amundsen_application/models/user.py]: I don't recall ever seeing any responses like `" "` returned from metadata service. The fields are mostly `""` when empty. Not sure if this is the actual root cause of the bug.

File: amundsen_application/models/user.py [hunk 3]
```diff
@@ -63,19 +63,19 @@ class UserSchema(Schema):
...
 
-        if not data.get('profile_url', None):
+        if _str_no_value(data.get('profile_url')):
             data['profile_url'] = ''
             if app.config['GET_PROFILE_URL']:
                 data['profile_url'] = app.config['GET_PROFILE_URL'](data['user_id'])
 
-        if not data.get('display_name', None):
+        if _str_no_value(data.get('display_name')):
             data['display_name'] = data.get('full_name', data.get('email'))
```
danwom: `data.get('full_name', data.get('email'))` is the root cause of the bug, when `full_name` is an empty string, it assigns that value to `data['display_name']`. Perhaps something like:

```
if not data.get('display_name'):
    if data.get('full_name'):
        data['display_name'] = data.get('full_name')
    else:
        data['display_name'] = data.get('email')
```
_disclaimer: this code could probably be cleaned up_

amundsen_application/models/user.py
ttannis: The `"  "` is not the root cause of the bug, but accounting for it is just an extra precaution. The only contract between frontend and metadata is that this value is going to be a string.

ttannis:
> do you want to add a small unit test to cover the case here?

Added several tests to explicitly cover all of the main logic in the user model.

PR #156, opened 2019-05-13
Add Bookmarks Feature

danwom:
### Summary of Changes

- Added the Bookmarks feature where users can save or unsave table resources.
  - BookmarkIcon component renders a filled or unfilled star next to table resources. Clicking on the icon will save or unsave the bookmark. Icons can be rendered in regular or large versions.
  - "My Bookmarks" component is added to the homepage. This displays all of the bookmarked table resources in a paginated format.
- Created a Preloader component that fetches data needed for app-wide functionality.
  - Bookmark and User data is fetched in Preloader
  - Removed user data fetching in NavBar

### Tests
Added new tests for
- Preloader, BookmarkIcon, BookmarkList, Metadata API
Modified existing tests for
- Table Details, TableListItem,

### Documentation

### CheckList


danwom:
Ready for review @ttannis @feng-tao @jinhyukchang 

Sorry for the large PR. Initially I thought this would be a small, simple change and then it kept growing...

CC: @verdan @jornh

danwom:
Homepage
Screen Shot 2019-05-22 at 3 07 17 PM


Table Details
Screen Shot 2019-05-22 at 3 07 49 PM

ttannis:
> Ready for review @ttannis @feng-tao @jinhyukchang
> 
> Sorry for the large PR. Initially I thought this would be a small, simple change and then it kept growing...
> 
> CC: @verdan @jornh

No problem. Should be able to review by end of day tomorrow.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -151,6 +152,7 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
     };
     return (
         <div className="search-list-container">
+          <BookmarkList />
```
ttannis: If I had to choose, this could go after popular tables. On a smaller screen I don't even see any hint that Popular Tables still exists if I have 5+ bookmarks...at least the other way around "My Bookmarks" title at least fits into view. If Product is on board, after this PR how do you feel about taking some time to think about a good "HomePage" layout?

amundsen_application/static/js/ducks/bookmark/sagas.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    // TODO - Consider adding the newly bookmarked resource directly to local store. This would save a round trip.
+    response = yield call(getBookmarks);
+    yield put({ type: AddBookmark.SUCCESS, payload: response.bookmarks });
+  } catch(e) {
+    yield put({ type: AddBookmark.FAILURE, payload: response });
+   }
+}
+export function* addBookmarkWatcher(): SagaIterator {
+  yield takeEvery(AddBookmark.ACTION , addBookmarkWorker)
```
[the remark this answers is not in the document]
ttannis: On second thought with what I'm describing we would actually want `takeLeading`, not `takeLastest`.

amundsen_application/static/js/components/SearchPage/index.tsx
ttannis: Resolving as we spoke offline about plans for a "HomePage", layout will be decided as part of that PR.

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -457,11 +457,11 @@ describe('SearchPage', () => {
...
-      expect(content.children().at(0).find('label').text()).toEqual(POPULAR_TABLES_LABEL);
+      expect(content.children().at(1).find('label').text()).toEqual(POPULAR_TABLES_LABEL);
     });
 
     it('renders InfoButton with correct props', () => {
-      expect(content.children().at(0).find(InfoButton).props()).toMatchObject({
+      expect(content.children().at(1).find(InfoButton).props()).toMatchObject({
         infoText: POPULAR_TABLES_INFO_TEXT,
       });
     });
```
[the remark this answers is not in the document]
ttannis: Resolving because we spoke offline about plans to implement a "HomePage" reasonably soon.

amundsen_application/api/utils/data_utils.py
```diff
@@ -0,0 +1,29 @@
+from typing import Dict
+
+from amundsen_application.api.utils.request_utils import get_query_param
+
+
+def get_table_key(args: Dict) -> str:
```
feng-tao: docstring. Since we put the function in a util file, others should be able to leverage it.

amundsen_application/api/utils/data_utils.py
```diff
@@ -0,0 +1,29 @@
...
+    db = get_query_param(args, 'db')
+    cluster = get_query_param(args, 'cluster')
+    schema = get_query_param(args, 'schema')
+    table = get_query_param(args, 'table')
+    table_key = '{db}://{cluster}.{schema}/{table}'.format(**locals())
+    return table_key
+
+
+def marshall_table_partial(table: Dict) -> Dict:
```
feng-tao: docstring.

amundsen_application/api/utils/data_utils.py
```diff
@@ -0,0 +1,29 @@
...
+    table_name = table.get('table_name')
```
feng-tao: table_name = table.get('table_name', ''). The default value is None . Same for all the other string fields.

amundsen_application/api/utils/data_utils.py
```diff
@@ -0,0 +1,29 @@
...
+    db = table.get('database')
+    return {
+        'cluster': cluster,
+        'database': db,
+        'description': table.get('table_description'),
+        'key': '{0}://{1}.{2}/{3}'.format(db, cluster, schema_name, table_name),
+        'name': table_name,
+        'schema_name': schema_name,
+        'type': 'table',
+        'last_updated_epoch': table.get('last_updated_epoch', None),
```
feng-tao: could you add a note/ todo about combining this two different terms(search/metadata) into a single term?

amundsen_application/api/metadata/v0.py
```diff
@@ -179,6 +156,7 @@ def _map_user_object_to_schema(u: Dict) -> Dict:
         ]
 
         results = {key: response.json().get(key, None) for key in params}
+        results["key"] = table_key
```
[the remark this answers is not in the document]
feng-tao: +1 on single quote

amundsen_application/api/metadata/v0.py
```diff
@@ -494,3 +472,68 @@ def _log_get_user(*, user_id: str) -> None:
         logging.exception(message)
         payload = jsonify({'msg': message})
         return make_response(payload, HTTPStatus.INTERNAL_SERVER_ERROR)
+
+
+@metadata_blueprint.route('/user/bookmark', methods=['GET'])
+def get_bookmark() -> Response:
```
feng-tao: docstring on what this endpoint is about.

amundsen_application/api/metadata/v0.py
```diff
@@ -494,3 +472,68 @@ def _log_get_user(*, user_id: str) -> None:
...
+    try:
+        user_id = request.args.get('user_id')
+        if user_id is None:
+            if app.config['AUTH_USER_METHOD']:
```
feng-tao: does this part of auth work with flaskoidc? cc @verdan

amundsen_application/api/metadata/v0.py
```diff
@@ -494,3 +472,68 @@ def _log_get_user(*, user_id: str) -> None:
...
+                user_id = app.config['AUTH_USER_METHOD'](app).user_id
+            else:
+                raise Exception('AUTH_USER_METHOD is not configured')
+
+        url = '{0}{1}/{2}/follow/'.format(app.config['METADATASERVICE_BASE'], USER_ENDPOINT, user_id)
```
feng-tao: we only use bookmark endpoint . But we could put a todo/note on putting the share endpoint into a constant once we introduce own/read endpoint in FE.

amundsen_application/api/metadata/v0.py
```diff
@@ -494,3 +472,68 @@ def _log_get_user(*, user_id: str) -> None:
...
+        status_code = response.status_code
+
+        payload = {
+            'msg': 'success',
+            'response': response.json()
+        }
+
+        return make_response(jsonify(payload), status_code)
+    except Exception as e:
+        message = 'Encountered exception: ' + str(e)
```
feng-tao: could we just combine the two lines into single line? We normally just put the payload directly into the same line instead of creating an additional variable for that.

amundsen_application/api/utils/data_utils.py
```diff
@@ -0,0 +1,29 @@
+from typing import Dict
```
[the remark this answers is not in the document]
feng-tao: will the functions within the file be shared by search/metadata together later?

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -8,6 +8,7 @@ import Pagination from 'react-js-pagination';
 import SearchBar from './SearchBar';
 import SearchList from './SearchList';
 
+import BookmarkList from 'components/common/Bookmark/BookmarkList'
```
feng-tao: small nit: should we enforce alphabetical order?

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -151,6 +152,7 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
     };
     return (
         <div className="search-list-container">
```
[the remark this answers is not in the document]
danwom: Filed some tickets for this.

amundsen_application/static/js/components/TableDetail/types.ts
```diff
@@ -76,6 +76,7 @@ export interface TableMetadata {
   database: string;
   is_editable: boolean;
   is_view: boolean;
+  key: string;
```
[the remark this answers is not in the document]
danwom: Filed a ticket for this.

amundsen_application/api/metadata/v0.py
danwom: Since the URL pattern requires the `user_id` to go in the middle of the string `[metadataservice]/user/[user_id]/follow`, it makes it difficult use a single constant. 

We could do something like
```
'{0}{1}/{2}/{3}'.format(app.config['METADATASERVICE_BASE'], USER_ENDPOINT, user_id, FOLLOW_ENDPOINT)
```
which seems somewhat complicated. Maybe a utility function could generate this in a better way?

This might be breaking the restful pattern but something like the following might be easier.
```
user/follow/[user_id]
user/own/[user_id]
user/read/[user_id]
```

amundsen_application/api/utils/data_utils.py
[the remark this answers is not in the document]
danwom: Currently this is only used in metadata, with search to follow shortly. 
Ideally, I think the data passed from metadata and search services should share a consistent data format and FE shouldn't have to worry about generating or renaming fields such as `key`. Perhaps these data formats belong in Amundsen Common.

verdan:
> great job! If user search items, will user's bookmark still be hidden or shown?

same question. Is there a way user can access bookmarks from navigation?

amundsen_application/api/metadata/v0.py
verdan: This should be fine. I haven't configured the auth_user_method yet, but I am pretty sure that I would be able to get this info from from flask's app context. thanks for confirming 👍

amundsen_application/api/metadata/v0.py
verdan: not sure why the following is complicated
```
'{0}{1}/{2}/{3}'.format(app.config['METADATASERVICE_BASE'], USER_ENDPOINT, user_id, FOLLOW_ENDPOINT)
```
but I agree with @danwom we should NOT go against the restful pattern.

amundsen_application/api/metadata/v0.py
```diff
@@ -494,3 +472,68 @@ def _log_get_user(*, user_id: str) -> None:
...
+        user_id = request.args.get('user_id')
+        if user_id is None:
+            if app.config['AUTH_USER_METHOD']:
+                user_id = app.config['AUTH_USER_METHOD'](app).user_id
+            else:
+                raise Exception('AUTH_USER_METHOD is not configured')
+
+        url = '{0}{1}/{2}/follow/'.format(app.config['METADATASERVICE_BASE'], USER_ENDPOINT, user_id)
+
+        response = request_metadata(url=url, method=request.method)
```
verdan: If the response is not a success message, this would not go in the Exception block, and will raise exception when try to proceed normal try block. 
I think you should call `raise_for_status()` method of response, to properly raise the exception. 

Also, the whole try except block is repetitive, how about make this a decorator?

danwom:
> > great job! If user search items, will user's bookmark still be hidden or shown?
> 
> same question. Is there a way user can access bookmarks from navigation?

@feng-tao @verdan currently the `Bookmark List` will hide/show the same as the `Popular Tables`. @ryanlieu is working on some changes to easily get back to the default "home" view.

amundsen_application/static/js/ducks/bookmark/sagas.ts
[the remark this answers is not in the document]
danwom: I'm not sure we should use `takeLeading` or `takeLatest` here. I can see the use case for clicking on the same bookmark, but if a user adds two different bookmarks this could lead to some weird edge cases where one is added and the other is stuck.

We don't have a good pattern established for this yet, but I think we could look into making better use of `isLoading` to disable buttons while an existing request is in flight. Maybe this could be a follow up task?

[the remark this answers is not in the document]
ttannis: Ah that's completely right. There is an approach for sharing actions between components involving appending unique prefixes or suffixes to the action type, in which case we could then use `takeLeading`, but that is its own endeavor. I don't think the edge case I raised is worth sinking time into that right now given our other work. I support resolving this comment and making use of `isLoading` later, or forgoing it and putting the effort into the Redux layer when we have the time -- your call!

amundsen_application/config.py
```diff
@@ -50,7 +51,11 @@ class LocalConfig(Config):
...
     REQUEST_HEADERS_METHOD = None
 
-    AUTH_USER_METHOD = None
+    AUTH_USER_METHOD = None  # type: Optional[function]
     GET_PROFILE_URL = None
 
     MAIL_CLIENT = None
+
+
+class TestConfig(LocalConfig):
```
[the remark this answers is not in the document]
danwom: This works as expected. I forgot what it was, but there was a different issue (also resolved) that looked like `TestConfig` was failing.

amundsen_application/static/js/components/SearchPage/index.tsx
danwom: Imports are loosely sorted within each group of imports. We have 3rd party imports on lines 1-6, relative imports within the same folder on lines 8-9, `/components` on lines 11-14, and `/ducks` on lines 16-27. Constants on line 32 could be grouped with other relative imports.

We don't have an explicit pattern, but we could certainly look into establishing a stronger one. For example, should we sort by import path or named imported? Should imports be grouped or alphabetize altogether? Where do styles belong? I'd be happy to revisit in a follow up task.

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -70,6 +71,7 @@ export class TableDetail extends React.Component<TableDetailProps & RouteCompone
       database: '',
       is_editable: false,
       is_view: false,
+      key: '',
```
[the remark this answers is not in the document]
danwom: I'll keep this in mind in the future. Perhaps we should file a task to cleanup this pattern?

amundsen_application/api/metadata/v0.py
danwom: Filed: https://github.com/lyft/amundsenfrontendlibrary/issues/177

Updating this pattern is probably out of scope, but that's a good idea. I'm not exactly certain we're doing the best with exception handling here.

danwom:
@ttannis @feng-tao Ready for re-review.

feng-tao:
The python part LGTM.

amundsen_application/static/js/components/TableDetail/index.tsx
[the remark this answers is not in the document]
ttannis: Mentioning this nit mostly for others and for new components. I've been doing it/will continue to do it as part of unit tests, but this component has remained untested until we understand the specs for the upcoming redesign.

Issue #69, opened 2019-04-08
Document how to integrate amundsenfrontend with Apache superset


feng-tao:
At Lyft, we integrate Amundsen with Apache superset for table data preview as well as sql exploration. We should document how we do the integration (e.g which endpoint we call superset to get the preview data etc)

feng-tao:
cc @danwom @ttannis

feng-tao:
FYI @verdan

verdan:
@feng-tao it would be nice to have this doc available anytime soon :)

feng-tao:
@verdan , I only knew we used https://github.com/apache/incubator-superset/blob/959c35d506759bdbea48db55ef61677f4c1ebd7d/superset/views/core.py#L2589 to get the data preview from superset sqllab. But I will defer it to @ttannis  as she and Max(@mistercrunch) did the setup :( I never setup superset locally before.

feng-tao:
And for the context, we have an internal version of superset client in Amundsen which uses internal ACL header to query superset as we need to pass user session info to superset to Authz and see if the user is allowed to see the actual data.

I think we could modify the client to make it open source able. The difficult part is that both of us haven't setup the master branch of superset locally before. So it will be a bit hard to test it out our client whether it works against the OSS version.

feng-tao:
cc @youngyjd as he did setup superset locally before.

ttannis:
Addressed with https://github.com/lyft/amundsenfrontendlibrary/pull/159

PR #159, opened 2019-05-15
Add `base_superset_client` & documentation

ttannis:
### Summary of Changes
This PR addresses https://github.com/lyft/amundsenfrontendlibrary/issues/69:
1. Created a `base_superset_preview_client` as a starting point for using Superset as the provider for Amundsen's data preview feature.
2. Implemented an `example_superset_preview_client`. Note that this is only an example for getting data from a default table in an instance of Superset, running on localhost with security/auth features disabled. It will not work out-of-the-box in production without more custom logic.
3. Updated the parameters for the data preview feature to include the `database`. The request to Superset's `sql_json` endpoint requires a `database_id`, so we pass the database name for the client to incorporate logic for the potential use case of querying different databases. 

### Tests

1. Added `tests/unit/base/test_superset_preview_client.py` to fully cover the testable logic in that file.
2. Unrelated cleanup: Modified some assertions in `test/unit/api/search/test_v0.py` to use a config value for comparison instead of the hardcoded equivalent.

### Documentation

1. Added doc to explain how to approach the implementation of `base_superset_preview_client`: https://github.com/lyft/amundsenfrontendlibrary/blob/fec747b9aa73cddd79ed533dc1cfd3932d5ef21f/docs/examples/superset_preview_client.md

### CheckList

### DIFF
--- a/amundsen_application/base/examples/example_superset_preview_client.py
+++ b/amundsen_application/base/examples/example_superset_preview_client.py
@@ -7,6 +7,7 @@ from typing import Any, Dict  # noqa: F401
 
 from amundsen_application.base.base_superset_preview_client import BaseSupersetPreviewClient
 
+# 'main' is an existing default Superset database which serves for demo purposes
 DEFAULT_DATABASE_MAP = {
     'main': 1,
 }
@@ -39,8 +40,12 @@ class SupersetPreviewClient(BaseSupersetPreviewClient):
 
             # Generate the sql query for the desired data preview content
             try:
+                # 'main' is an existing default Superset schema which serves for demo purposes
                 schema = 'main'  # OR params.get('schema') in a real use case
+
+                # 'ab_role' is an existing default Superset table which serves for demo purposes
                 table_name = 'ab_role'  # OR params.get('tableName') in a real use case
+
                 request_data['sql'] = 'SELECT * FROM {schema}.{table} LIMIT 50'.format(schema=schema, table=table_name)
             except Exception as e:
                 logging.error('Encountered error generating request sql: ' + str(e))
--- a/amundsen_application/base/examples/example_superset_preview_client.py
+++ b/amundsen_application/base/examples/example_superset_preview_client.py
@@ -1,0 +1,51 @@
+import logging
+import requests
+import uuid
+
+from requests import Response
+from typing import Any, Dict  # noqa: F401
+
+from amundsen_application.base.base_superset_preview_client import BaseSupersetPreviewClient
+
+DEFAULT_DATABASE_MAP = {
+    'main': 1,
+}
+DEFAULT_URL = 'http://localhost:8088/superset/sql_json/'
+
+
+class SupersetPreviewClient(BaseSupersetPreviewClient):
+    def __init__(self,
+                 *,
+                 database_map: Dict[str, int] = DEFAULT_DATABASE_MAP,
+                 url: str = DEFAULT_URL) -> None:
+        self.database_map = database_map
+        self.headers = {}
+        self.url = url
+
+    def post_to_sql_json(self, *, params: Dict, headers: Dict) -> Response:
+        """
+        Returns the post response from Superset's `sql_json` endpoint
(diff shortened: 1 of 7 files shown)


feng-tao:
cc @verdan @bolkedebruin @jornh

ttannis:
cc @verdan 

Here's what we have in the works so far.

amundsen_application/base/examples/example_superset_preview_client.py
```diff
@@ -0,0 +1,51 @@
+import logging
+import requests
+import uuid
+
+from requests import Response
+from typing import Any, Dict  # noqa: F401
+
+from amundsen_application.base.base_superset_preview_client import BaseSupersetPreviewClient
+
+DEFAULT_DATABASE_MAP = {
```
feng-tao: The repo has a directory for example/docker/neo4j/conf(maybe we should move it to metadata), but it would be good to put this file under the same example dir.

amundsen_application/base/examples/example_superset_preview_client.py
```diff
@@ -0,0 +1,51 @@
...
+            request_data['client_id'] = uuid.uuid4()
+
+            # Superset's sql_json endpoint requires the id of the database that it will execute the query on
+            database_name = 'main'  # OR params.get('database') in a real use case
+            request_data['database_id'] = self.database_map.get(database_name, '')
+
+            # Generate the sql query for the desired data preview content
+            try:
+                schema = 'main'  # OR params.get('schema') in a real use case
+                table_name = 'ab_role'  # OR params.get('tableName') in a real use case
```
feng-tao: should we mention that `ab_role` is an existing superset table which serves for demo purpose?

amundsen_application/base/examples/example_superset_preview_client.py
```diff
@@ -0,0 +1,51 @@
...
+                request_data['sql'] = 'SELECT * FROM {schema}.{table} LIMIT 50'.format(schema=schema, table=table_name)
+            except Exception as e:
+                logging.error('Encountered error generating request sql: ' + str(e))
```
feng-tao: not sure why we will encounter exception here? And this exception doesn't seem to be necessary given you will catch any exception in outer block?

docs/examples/superset_preview_client.md
```diff
@@ -0,0 +1,57 @@
+# Overview
+
+Amundsen's data preview feature requires that developers create a custom implementation of `base_preview_client` for requesting that data. This feature assists with data discovery by providing the end user the option to view a sample of the actual resource data so that they can verify whether or not they want to transition into exploring that data, or continue their search.
+
+[Superset](https://github.com/apache/incubator-superset) is an open-source business intelligence tool that can be used for data exploration. Amundsen's data preview feature was created with Superset in mind, and it is what we leverage internally at Lyft to support the feature. This document provides some insight into how to configure Amundsen's frontend application to leverage Superset for data previews.
```
feng-tao: Apache Superset?

amundsen_application/base/examples/example_superset_preview_client.py
ttannis: The example logic here is very simple and we wouldn't encounter an error on these 3 lines specifically, but I set this up as if engineers would copy-paste and then fill in their own, more complicated logic.

amundsen_application/base/examples/example_superset_preview_client.py
ttannis: Yeah, I agree we should call this our directly in the file in case people don't read the doc.

amundsen_application/base/examples/example_superset_preview_client.py
ttannis: My understanding was that we would move this file to the parent repository? I do like this file here though, as it makes sense to me to have examples for our abstract base clients in the base directory.

feng-tao: sg

PR #167, opened 2019-05-21
Reverse frequent user order positioning

ryanlieu:
### Summary of Changes

Change z index of frequent user avatars so the most frequent user is at the top

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/css/_avatars.scss
+++ b/amundsen_application/static/css/_avatars.scss
@@ -1,0 +1,7 @@
+.sb-avatar > div {
+  border: 1px solid white;
+}
+
+.sb-avatar > img {
+  margin: 0px;
+}
--- a/amundsen_application/static/js/components/TableDetail/index.tsx
+++ b/amundsen_application/static/js/components/TableDetail/index.tsx
@@ -124,12 +124,11 @@ export class TableDetail extends React.Component<TableDetailProps & RouteCompone
     })
   };
 
-  getAvatarForUser(fullName, profileUrl) {
+  getAvatarForUser(fullName, profileUrl, zIndex) {
     const popoverHoverFocus = (
       <Popover id="popover-trigger-hover-focus">
         {fullName}
       </Popover>);
-
     if (profileUrl.length !== 0) {
       return (
         <OverlayTrigger key={fullName} trigger={['hover', 'focus']} placement="top" overlay={popoverHoverFocus}>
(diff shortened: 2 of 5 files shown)


amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -146,7 +145,7 @@ export class TableDetail extends React.Component<TableDetailProps & RouteCompone
       <OverlayTrigger key={fullName} trigger={['hover', 'focus']} placement="top" overlay={popoverHoverFocus}>
         <div style={{display: 'inline-block', marginLeft: '-5px',
           backgroundColor: 'white', borderRadius: '90%'}}>
-          <Avatar name={fullName} size={25} round={true} style={{border: '1px solid white'}}/>
+          <Avatar name={fullName} size={25} round={true} style={{zIndex, position: 'relative'}}/>
```
danwom: The Avatars here are inheriting the `border: 1px solid white` rule as defined in `NavBar/styles.scss`. I don't think styles like that should leak out into other places that were not originally intended. If the intent is that all Avatars should have the border, then it should be defined at the appropriate level. 

As I mentioned earlier, the NavBar stylesheet could use some cleanup.

ryanlieu: So I got rid of the inline styling here because it looked like it was unnecessary (.sb-avatar already had a border attribute that was being doubled up by the inline style). With that inline style the border also looked asymmetrical like in [here](https://jira.lyft.net/secure/attachment/224446/image-2019-04-04-16-15-41-093.png)

danwom: Yes, I completely understand that. I'm just saying `NavBar/styles.scss` could use some TLC.

ryanlieu: ended up moving the avatar specific global styling into a new file, should hopefully make it a little more clear where global avatar specific styling lives rather than inside of components

amundsen_application/static/css/_avatars.scss
```diff
@@ -0,0 +1,7 @@
+.sb-avatar > div {
+  border: 1px solid white;
+}
+
+.sb-avatar > img {
```
ryanlieu: just took this from the AvatarLabel component, might want to scope this styling down to just avatars inside of component-labels since this might lead to margin issues if we want sb-avatars with images in other elements

jornh:
Overall a great idea wrt. UI experience! ❤️

PR #171, opened 2019-05-22
Fixed state to match URL changes

ryanlieu:
### Summary of Changes

Made changes to the search page so that changes to the URL would be reflected in changes to the state of the search page. Two things that were fixed in this process: the back and forward buttons on the search page as well as the home button.

### Tests

Modified SearchPage test so that SearchPage wrapper expects RouteComponentProps from react router.

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/SearchPage/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/index.tsx
@@ -4,6 +4,7 @@ import { bindActionCreators } from 'redux';
 import * as DocumentTitle from 'react-document-title';
 import * as qs from 'simple-query-string';
 import Pagination from 'react-js-pagination';
+import { RouteComponentProps } from 'react-router';
 
 import SearchBar from './SearchBar';
 import SearchList from './SearchList';
@@ -48,7 +49,7 @@ export interface StateFromProps {
   searchTerm: string;
   popularTables: TableResource[];
   tables: TableSearchResults;
-  dashboards: DashboardSearchResults
+  dashboards: DashboardSearchResults;
   users: UserSearchResults;
 }
 
@@ -58,7 +59,7 @@ export interface DispatchFromProps {
   getPopularTables: () => GetPopularTablesRequest;
 }
 
-export type SearchPageProps = StateFromProps & DispatchFromProps;
+export type SearchPageProps = StateFromProps & DispatchFromProps & RouteComponentProps<any>;
 
 interface SearchPageState {
   selectedTab: ResourceType;
@@ -77,17 +78,25 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
 
   componentDidMount() {
     this.props.getPopularTables();
-
-    const params = qs.parse(window.location.search);
+    const params = qs.parse(this.props.location.search);
     const { searchTerm, pageIndex, selectedTab } = params;
-
-    const currentTab = this.getSelectedTabByResourceType(selectedTab);
+    const { term, index, currentTab } = this.getSanitizedUrlParams(searchTerm, pageIndex, selectedTab);
     this.setState({ selectedTab: currentTab });
-    if (searchTerm && searchTerm.length > 0) {
-      const index = pageIndex || 0;
-      this.props.searchAll(searchTerm, this.createSearchOptions(index, currentTab));
-      // Update the page URL with validated parameters.
-      this.updatePageUrl(searchTerm, currentTab, index);
+    if (term !== "") {
+      this.props.searchAll(term, this.createSearchOptions(index, currentTab));
+      if (currentTab !== selectedTab || pageIndex !== index) {
+        this.updatePageUrl(term, currentTab, index);
+      }
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    if (this.props.location.search !== prevProps.location.search) {
+      const params = qs.parse(this.props.location.search);
+      const { searchTerm, pageIndex, selectedTab } = params;
+      const { term, index, currentTab } = this.getSanitizedUrlParams(searchTerm, pageIndex, selectedTab);
+      this.setState({ selectedTab: currentTab });
+      this.props.searchAll(term, this.createSearchOptions(index, currentTab));
     }
   }
 
@@ -110,6 +119,13 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
     };
   };
 
+  getSanitizedUrlParams = (searchTerm: string, pageIndex: number, selectedTab: ResourceType) => {
+    const currentTab = this.getSelectedTabByResourceType(selectedTab);
+    const index = pageIndex || 0;
+    const term = searchTerm ? searchTerm : "";
+    return {term, index, currentTab};
+  };
+
   getPageIndexByResourceType = (tab: ResourceType): number => {
     switch(tab) {
       case ResourceType.table:
@@ -123,7 +139,6 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
   };
 
   onSearchBarSubmit = (searchTerm: string): void => {
-    this.props.searchAll(searchTerm);
     this.updatePageUrl(searchTerm, this.state.selectedTab,0);
   };
 
@@ -141,7 +156,7 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
 
   updatePageUrl = (searchTerm: string, tab: ResourceType, pageIndex: number): void => {
     const pathName = `/search?searchTerm=${searchTerm}&selectedTab=${tab}&pageIndex=${pageIndex}`;
-    window.history.pushState({}, '', `${window.location.origin}${pathName}`);
+    this.props.history.push(pathName);
   };
 
   renderPopularTables = () => {
--- a/amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
+++ b/amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
@@ -32,7 +32,6 @@ import globalState from 'fixtures/globalState';
 
 describe('SearchPage', () => {
   const setStateSpy = jest.spyOn(SearchPage.prototype, 'setState');
-
   const setup = (propOverrides?: Partial<SearchPageProps>) => {
     const props: SearchPageProps = {
       searchTerm: globalState.search.search_term,
@@ -43,7 +42,28 @@ describe('SearchPage', () => {
       searchAll: jest.fn(),
       searchResource: jest.fn(),
       getPopularTables: jest.fn(),
-      ...propOverrides
+      history: {
+        length: 2,
+        action: "POP",
+        location: jest.fn() as any,
+        push: jest.fn(),
+        replace: jest.fn(),
+        go: jest.fn(),
+        goBack: jest.fn(),
+        goForward: jest.fn(),
+        block: jest.fn(),
+        createHref: jest.fn(),
+        listen: jest.fn(),
+      },
+      location: {
+        search: '/search?searchTerm=testName&selectedTab=table&pageIndex=1', 
+        pathname: 'mockstr',
+        state: jest.fn(),
+        hash: 'mockstr',
+      },
+      match: jest.fn() as any,
+      staticContext: jest.fn() as any,
+      ...propOverrides,
     };
     const wrapper = shallow<SearchPage>(<SearchPage {...props} />)
     return { props, wrapper };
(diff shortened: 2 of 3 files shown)


amundsen_application/static/js/ducks/search/types.ts
```diff
@@ -42,6 +42,7 @@ export interface SearchAllRequest {
   options: SearchAllOptions;
   term: string;
   type: SearchAll.ACTION;
+  payload?: SearchReducerState;
```
danwom: I don't think the `SearchAllRequest` ever has a `payload`. I'm guessing you added this to suppress an error on line 61 of `search/reducer.ts`. This is caused because not every action in the reducer function is guaranteed to have a `payload`.

There are some alternatives. You could:

1. Move `const newState = action.payload` from `reducer.ts` line 61 into the sections on line 70, 76. Or you can just reference `action.payload` directly.

2. Require `SearchAllRequest` to have a payload and move `term` and `type` into payload. This goes against the established pattern we have where none of the Requests have a payload, but I wouldn't be opposed to following this pattern for all redux actions. @ttannis would also need to get onboard here. There are a few patterns that I have stronger opinions about now that I am more familiar with redux/saga.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -91,6 +92,18 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
     }
   }
 
+  componentDidUpdate(prevProps) {
+    if (this.props.location !== prevProps.location) {
```
ttannis: It looks like we have some shared logic here between `componentDidMount` and `componentDidUpdate`. Can we consolidate that into a separate method? (will require updating tests)

amundsen_application/static/js/ducks/search/types.ts
ttannis: Spoke offline. Fine with this change and I'll pick up the task to clean up what we're doing here alongside unit test additions.

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -43,7 +43,11 @@ describe('SearchPage', () => {
       searchAll: jest.fn(),
       searchResource: jest.fn(),
       getPopularTables: jest.fn(),
-      ...propOverrides
+      history: mock,
```
ttannis: It's a better practice to declare these each individually as `jest.fn()` instead of reusing the same object. The reason is that in the event that we need to mock a specific functionality of any of these objects, we'd end up applying that change to all of these parameters instead of just the one it was intended for.

amundsen_application/static/js/components/SearchPage/index.tsx
ttannis: One more comment, was this `if` statement intended to prevent this logic from being called twice on the initial render? (On initial render `componentDidMount` will get called, which makes changes to the component so `componetDidUpdate` will also fire). If that is the case, `location` is an object and `if (this.props.location !== prevProps.location)` could always evaluate to true. 

Can you confirm if this is the case? if so, we could do `if (this.props.location.search !== prevProps.location.search)` to compare the actual search strings

amundsen_application/static/js/ducks/search/reducer.ts
```diff
@@ -58,19 +58,26 @@ const initialState: SearchReducerState = {
 };
 
 export default function reducer(state: SearchReducerState = initialState, action: SearchReducerAction): SearchReducerState {
-  const newState = action.payload;
   switch (action.type) {
+    // Updates search term to reflect action
```
ttannis: I notice a side effect of this is that since the search team gets updated when the request is made, the component will re-render because components connected to Redux will re-render when their properties are updated. As a result the component will temporarily render in a state where it has as search term, but no search results. This leads to the user seeing a brief flash of: "Your search -- <searchTerm> -- did not match any tables result".

There are many ways we can address this, from rethinking our logic to introducing a loading state on the `SearchPage` in the same way that it exists on the `TableDetailPage`. I'm in favor of introducing the loading state in order to make these pages consistent, which we can do as a follow up PR.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -91,6 +92,18 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
     }
```
ttannis: nit: This isn't part of your PR, but on another look I don't think we need to call `this.updatePageUrl()` in this method unless `selectedTab != currentTab` or `pageIndex != index`. Otherwise we're just updating the url with the exact parameters we just got from it. 

No need to make this change in this PR, but I wanted to call it out and we can clean that up on another PR.

amundsen_application/static/js/ducks/search/reducer.ts
ryanlieu: Ah, yeah was going to mention this, I think a loading state would make the most amount of sense here. Should I just file a follow up task?

ttannis: Already done, you should see them assigned to you in our internal queue.

ryanlieu: :+1:

amundsen_application/static/js/components/SearchPage/index.tsx
ryanlieu: fixed this updatePageUrl call to only fix the url when selectedTab or pageIndex were out of sync with the state of the page

amundsen_application/static/js/components/SearchPage/index.tsx
ryanlieu: Right, yeah the object comparison here isn't great - comparing search directly makes more sense

ryanlieu:
Addressed comments, updated and added tests for componentDidUpdate and getSanitizedURLParams

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -110,6 +117,14 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
     };
   };
 
+  getSanitizedURLParams = (searchTerm: string, pageIndex: number, selectedTab: ResourceType) => {
+    const currentTab = this.getSelectedTabByResourceType(selectedTab);
+    this.setState({ selectedTab: currentTab });
```
danwom: (nit: Url, not URL) 

I'd prefer `setState` not to be called in the `getSanitizedUrlParams` function. The 'getter' function shouldn't be updating the state.

```
const { sanitizedTerm, index, sanitizedTab } = this.getSanitizedURLParams(searchTerm, pageIndex, selectedTab);
this.setState({ selectedTab: sanitizedTab });
this.props.searchAll(sanitizedTerm, ...    )

```

ryanlieu:
Moved setstate out of getSanitizedUrlParams and changed function name to getSanitizedUrlParams

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -74,11 +94,11 @@ describe('SearchPage', () => {
       props = setupResult.props;
       wrapper = setupResult.wrapper;
 
-      mockResourceType = ResourceType.user;
-      getSelectedTabByResourceTypeSpy = jest.spyOn(wrapper.instance(), 'getSelectedTabByResourceType').mockImplementation(() => {
-        return mockResourceType;
+      mockSatinitizedURLParams = { 'term': 'testName', ' index': 1, 'currentTab': 'table' };
```
danwom: nit: this can also be renamed to `mockSatinitizedUrlParams`

ryanlieu: oops gotcha, also Satinized is definitely not the right word hahaha

ryanlieu:
Fixed typos

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -155,14 +210,84 @@ describe('SearchPage', () => {
     });
 
     afterAll(() => {
-      getSelectedTabByResourceTypeSpy.mockRestore();
       createSearchOptionsSpy.mockRestore();
     });
   });
 
+  describe('componentDidUpdate', () => {
+    let props;
```
ttannis: To be a true unit test, we could make these improvements
1. Mock `getSanitizedUrlParams`
2. Mock `createSearchOptions`
3. Test the parameters that `searchAll` was called with.

This is a good practice even because even though the tests pass, this unit test has a dependency on the logic of the `getSanitizedUrlParams` and  `createSearchOptions`. When it comes to mocking, we mock/spyOn the methods we want to verify (like `setState` and `searchAll`), as well as the methods we want to treat as a black box (like `getSanitizedUrlParams` and  `createSearchOptions`).

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -155,14 +210,84 @@ describe('SearchPage', () => {
...
+      wrapper = setupResult.wrapper;
+      const mockPrevProps = {
+        location: {
+          search: '/search?searchTerm=previous2&selectedTab=table&pageIndex=0', 
+          pathname: 'mockstr',
+          state: jest.fn(),
+          hash: 'mockstr',
+        }
+      };
+      searchAllSpy = jest.spyOn(props, 'searchAll');
```
ttannis: Because `setStateSpy` is reused between tests, we should call `setStateSpy.mockClear();` before these blocks of test run. 

There actually multiple options here:
1. Create a new `setStateSpy` for every test
2. Call `mockClear()` before every test block that uses `setStateSpy`
3. Call `mockClear()` after every test block that uses `setStateSpy`
4. Some people do both 2 & 3 to be safe.

At the moment I have personally opted for Option 2 to create some consistency, but we need to make sure we do at least one of those options to prevent bugs in our tests.

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -155,14 +210,84 @@ describe('SearchPage', () => {
...
+      const mockResourceType = ResourceType.table;
+      getSelectedTabByResourceTypeSpy = jest.spyOn(wrapper.instance(), 'getSelectedTabByResourceType').mockImplementation(() => {
+        return mockResourceType;
+      });
+      
+      wrapper.instance().getSanitizedUrlParams('current', 0, ResourceType.table)
+    });
+
+    it('calls getSelectedTabByResourceType with correct value', () => {
+      expect(getSelectedTabByResourceTypeSpy).toHaveBeenCalledWith('table');
```
ttannis: Although this works, consider that the piece of logic we're testing here at this line https://github.com/lyft/amundsenfrontendlibrary/blob/b7486b019d303fa8e3d8ef7cdcc0618b036b314b/amundsen_application/static/js/components/SearchPage/index.tsx#L123 can be summarized as "getSelectedTabByResource should be called with the 3rd parameter passed into getSantizedUrlParams". This current check doesn't truly test that previous statement, because it's leveraging our own knowledge of how the component works:
1. We know that `ResourceType.table` evaluates to the string `'table'`. We should write the tests assuming we don't know that.
2. We aren't actually testing that the value is passed appropriately, because `ResourceType.table` and the hardcoded string `'table'` are two different objects. 

What we could do instead to make this test more robust is to create an actual variable for this value and test that it is passed accordingly:
```
testTab = ResourceType.table;
wrapper.instance().getSanitizedUrlParams('current', 0, testTab);
...
expect(getSelectedTabByResourceTypeSpy).toHaveBeenCalledWith(testTab);
```

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -155,14 +210,84 @@ describe('SearchPage', () => {
...
+    });
+
+    it('output of getSanitizedUrlParams is expected', () => {
```
ttannis: This test verifies the ideal case of `getSanitizedUrlParams`, which is that the searchTerm and index exist. To make the test of this method more robust, we should also set up the cases that would test the following https://github.com/lyft/amundsenfrontendlibrary/blob/b7486b019d303fa8e3d8ef7cdcc0618b036b314b/amundsen_application/static/js/components/SearchPage/index.tsx#L124-L125

which would involve adding a test like:
```
const expected = {'term': '', 'index': 0, 'currentTab': mockResourceType};
expect(wrapper.instance().getSanitizedUrlParams(undefined, undefined, ResourceType.table)).toEqual(expected);
```
nit: note the use of expecting `mockResourceType` simply because that is what the the mocked `getSelectedTabByResourceType` will return.

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -74,11 +94,11 @@ describe('SearchPage', () => {
       props = setupResult.props;
```
ttannis: We should be able to remove the line above:
`window.history.pushState({}, '', '/search?searchTerm=testName&selectedTab=table&pageIndex=1');`

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -93,40 +113,66 @@ describe('SearchPage', () => {
...
     });
 
-    it('calls getSelectedTabByResourceType with correct value', () => {
-      expect(getSelectedTabByResourceTypeSpy).toHaveBeenCalledWith('table');
-    });
-
-    it('calls setState with result of getSelectedTabByResourceType', () => {
-      expect(setStateSpy).toHaveBeenCalledWith({ selectedTab:  mockResourceType });
+    it('calls setState', () => {
+      expect(setStateSpy).toHaveBeenCalledWith({ selectedTab: ResourceType.table });
```
ttannis: checking `expect(setStateSpy).toHaveBeenCalledWith({ selectedTab: mockSanitizedUrlParams.currentTab });` more accurately tests the logic of this method.

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -93,40 +113,66 @@ describe('SearchPage', () => {
...
+            pathname: 'mockstr',
+            state: jest.fn(),
+            hash: 'mockstr',
+          }
+        });
+        updatePageUrlSpy = jest.spyOn(wrapper.instance(), 'updatePageUrl');
+        wrapper.instance().componentDidMount();
+      });
       it('calls searchAll', () => {
         expect(searchAllSpy).toHaveBeenCalledWith('testName', mockSearchOptions);
```
ttannis: checking `expect(searchAllSpy).toHaveBeenCalledWith(mockSanitizedUrlParams.term, mockSearchOptions);` will not more accurately test the logic.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -110,6 +119,13 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
     };
   };
 
+  getSanitizedUrlParams = (searchTerm: string, pageIndex: number, selectedTab: ResourceType) => {
+    const currentTab = this.getSelectedTabByResourceType(selectedTab);
```
ttannis: nit: It would appear this method could also have isolated the following shared logic as well
```
const params = qs.parse(this.props.location.search);
const { searchTerm, pageIndex, selectedTab } = params;
```

PR #173, opened 2019-05-24
Update quick start and docker-compose

amweiss:
### Summary of Changes

Update quick start to just use docker-compose and not docker-machine and fix setting the default password in neo4j so it can be run easier.

### Tests

No tests committed, but I ran through the setup on a machine with nothing to start and it worked out of the box.

### Documentation

Updated the installation.md file.

### CheckList
- [X] PR title addresses the issue accurately and concisely. Example: "Updates the version of Flask to v1.0.2"
- [X] PR includes a summary of changes, including screenshots of any UI changes. 
- [X] PR adds unit tests, updates existing unit tests, __OR__ documents why no test additions or modifications are needed.
- [X] In case of new functionality, my PR adds documentation that describes how to use it.
- [X] PR passes all tests documented in the [developer guide](https://github.com/lyft/amundsenfrontendlibrary/blob/master/docs/developer_guide.md#testing)
- [X] I have squashed multiple commits if they address the same issue. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)"

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -1,6 +1,7 @@
 # Installation
 
 ## Install standalone application directly from the source
+
 The following instructions are for setting up a standalone version of the Amundsen application. This approach is ideal for local development.
 ```bash
 # Clone repo
@@ -32,11 +33,11 @@ $ python3 amundsen_application/wsgi.py
 ```
 
 ## Bootstrap a default version of Amundsen using Docker
+
 The following instructions are for setting up a version of Amundsen using Docker. At the moment, we only support a bootstrap for connecting the Amundsen application to an example metadata service.
 
 1. Install `docker` and  `docker-compose`.
-2. Setup your local environment.
-   * Clone [this repo](https://github.com/lyft/amundsenfrontendlibrary) or download the [docker-amundsen.yml](https://github.com/lyft/amundsenfrontendlibrary/blob/master/docker-amundsen.yml) file directly.
+2. Clone [this repo](https://github.com/lyft/amundsenfrontendlibrary) or download the [docker-amundsen.yml](https://github.com/lyft/amundsenfrontendlibrary/blob/master/docker-amundsen.yml) file directly.
 3. Enter the directory where the `docker-amundsen.yml` file is and then:
     ```bash
     $ docker-compose -f docker-amundsen.yml up -d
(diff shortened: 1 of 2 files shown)


ttannis:
@amweiss what's the reasoning behind no longer using docker-machine + virtualbox and virtualenv?

amweiss:
> @amweiss what's the reasoning behind no longer using docker-machine + virtualbox and virtualenv?

It was extra work that isn't needed for a basic runtime to work. docker-compose is sufficient to get everything running by itself. If people want to create a more isolated environment via docker-machine, they can, but it's not as "quick" for normal usage.

docs/installation.md
```diff
@@ -34,42 +34,28 @@ $ python3 amundsen_application/wsgi.py
...
+5. You can verify dummy data has been ingested by viewing in Neo4j by visiting [`http://localhost:7474/browser/`](http://localhost:7474/browser/) and run `MATCH (n:Table) RETURN n LIMIT 25` in the query box. You should see two tables
+   1. `hive.test_schema.test_table1`
+   2. `dynamo.test_schema.test_table2`
+6. View UI at [`http://localhost:5000`](http://localhost:5000) and try to search `test`, it should return some result.
+7. From where the `docker-amundsen.yml` file is, run the following command when done:
+    ```bash
+    $ docker-compose -f docker-amundsen.yml down
     ```
-9. Verify dummy data has been ingested by viewing in Neo4j by visiting `http://YOUR-DOCKER-HOST-IP:7474/browser/` and run `MATCH (n:Table) RETURN n LIMIT 25` in the query box. You should see two tables -- `hive.test_schema.test_table1` and `dynamo.test_schema.test_table2`.
-10. View UI at `http://YOUR-DOCKER-HOST-IP:5000/table_detail/gold/hive/test_schema/test_table1` or `/table_detail/gold/dynamo/test_schema/test_table2`
```
ttannis: Verifying the UI at these urls has value because it checks that the  metadata service is running as expected. To summarize -- the old Step 9 verified Neo4j had the dummy data, the old Step 10 verified that the metadata service is correctly returning that data, and the old Step 11 verified that the search service can correctly search for that data.

I would opt for still keeping the step to view these urls to verify the metadata service, however we can just simply modify your new Step 6 to read:

> 6. View UI at [`http://localhost:5000`](http://localhost:5000) and try to search `test`, it should return some results. Click on one of the results, it should navigate to a page that displays the metadata for that resource.

amweiss: Step 5 still has them do that

ttannis: Step 5 asks them to verify that the data exists in Neo4 at `http://localhost:7474/browser/`. We would still want folks to verify that the application UI itself correctly displays that metadata from Neo4j at `http://localhost:5000/table_detail/...`

amweiss: I'm actually more tempted to remove step 5 altogether as well since it's not really part of a quick start. Those steps seem more appropriate for debugging.

ttannis:
> > @amweiss what's the reasoning behind no longer using docker-machine + virtualbox and virtualenv?
> 
> It was extra work that isn't needed for a basic runtime to work. docker-compose is sufficient to get everything running by itself. If people want to create a more isolated environment via docker-machine, they can, but it's not as "quick" for normal usage.

My understanding is that one of the reasons docker-machine is useful, if for those running on older operating systems. Given that we have not documented Amundsen as only able to run on certain more modern operating systems, instead of completely deleting these steps we could separate it as on optional set of steps for those who need it. 

cc @feng-tao Any concerns about having our quickstart no longer using docker-machine?

feng-tao:
@ttannis I think we should remove the docker-machine dependency if the docker-compose could work out of the box. I believe this is the case per the latest slack discussion.

docs/installation.md
ttannis: We intended for this documentation to include the steps to set up the tool and verify that it works. I would consider the last 3 steps of the old doc to be verification rather than debugging. 

If separating the setup from the verification is desirable then we can split these steps into two separate sections as opposed to deleting the instructions that help verify things work as expected. However I do prefer to keep them in the same flow, so that our installation instructions follow the pattern: setup --> verify it works --> teardown.

amweiss:
@ttannis and @feng-tao I moved the verification to a section at the bottom of the page but included both the detail UI page as well as the neo4j UI check.

feng-tao:
thanks @amweiss :)

PR #182, opened 2019-05-29
Added loading spinner for search requests

ryanlieu:
### Summary of Changes

Title, we needed some form of loading state so that delays between search requests wouldn't look awkward. There's a small hack here where we need to handle GetPopularTables actions from the Search reducer but this'll be addressed in a follow up with the decoupling of popular tables and search in the new home page

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/SearchPage/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/index.tsx
@@ -97,8 +97,11 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
       const params = qs.parse(this.props.location.search);
       const { searchTerm, pageIndex, selectedTab } = params;
       const { term, index, currentTab } = this.getSanitizedUrlParams(searchTerm, pageIndex, selectedTab);
+      const prevTerm = qs.parse(prevProps.location.search).searchTerm;
       this.setState({ selectedTab: currentTab });
-      this.props.searchAll(term, this.createSearchOptions(index, currentTab));
+      if (term !== prevTerm) {
+        this.props.searchAll(term, this.createSearchOptions(index, currentTab));
+      }
     }
   }
 
@@ -252,7 +255,7 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
       );
   };
 
-  renderTabs = () => {
+  renderContent = () => {
     if (this.props.isLoading) {
       return (<LoadingSpinner/>);
     }
@@ -269,7 +272,7 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
         <div className="row">
           <div className="col-xs-12 col-md-offset-1 col-md-10">
             <SearchBar handleValueSubmit={ this.onSearchBarSubmit } searchTerm={ searchTerm }/>
-            { this.renderTabs() }
+            { this.renderContent() }
           </div>
         </div>
       </div>
--- a/amundsen_application/static/js/components/SearchPage/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/index.tsx
@@ -97,8 +97,8 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
       const params = qs.parse(this.props.location.search);
       const { searchTerm, pageIndex, selectedTab } = params;
       const { term, index, currentTab } = this.getSanitizedUrlParams(searchTerm, pageIndex, selectedTab);
-      const prevTerm = qs.parse(prevProps.location.search).searchTerm;
       this.setState({ selectedTab: currentTab });
+      const prevTerm = qs.parse(prevProps.location.search).searchTerm;
       if (term !== prevTerm) {
         this.props.searchAll(term, this.createSearchOptions(index, currentTab));
       }
--- a/amundsen_application/static/js/components/SearchPage/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/index.tsx
@@ -98,7 +98,7 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
       const { searchTerm, pageIndex, selectedTab } = params;
       const { term, index, currentTab } = this.getSanitizedUrlParams(searchTerm, pageIndex, selectedTab);
       this.setState({ selectedTab: currentTab });
-      const prevTerm = qs.parse(prevProps.location.search).searchTerm;
+      const prevTerm = prevProps.searchTerm;
       if (term !== prevTerm) {
         this.props.searchAll(term, this.createSearchOptions(index, currentTab));
       }
--- a/amundsen_application/static/js/components/SearchPage/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/index.tsx
@@ -8,6 +8,7 @@ import { RouteComponentProps } from 'react-router';
 
 import SearchBar from './SearchBar';
 import SearchList from './SearchList';
+import LoadingSpinner from 'components/common/LoadingSpinner';
 
 import InfoButton from 'components/common/InfoButton';
 import { ResourceType, TableResource } from 'components/common/ResourceListItem/types';
@@ -47,6 +48,7 @@ import {
 
 export interface StateFromProps {
   searchTerm: string;
+  isLoading: boolean;
   popularTables: TableResource[];
   tables: TableSearchResults;
   dashboards: DashboardSearchResults;
@@ -250,6 +252,16 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
       );
   };
 
+  renderTabs = () => {
+    if (this.props.isLoading) {
+      return (<LoadingSpinner/>);
+    }
+    if (this.props.searchTerm.length > 0) {
+      return this.renderSearchResults();
+    }
+    return this.renderPopularTables();
+  };
+
   render() {
     const { searchTerm } = this.props;
     const innerContent = (
@@ -257,8 +269,7 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
         <div className="row">
           <div className="col-xs-12 col-md-offset-1 col-md-10">
             <SearchBar handleValueSubmit={ this.onSearchBarSubmit } searchTerm={ searchTerm }/>
-            { searchTerm.length > 0 && this.renderSearchResults() }
-            { searchTerm.length === 0 && this.renderPopularTables()  }
+            { this.renderTabs() }
           </div>
         </div>
       </div>
@@ -277,6 +288,7 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
 export const mapStateToProps = (state: GlobalState) => {
   return {
     searchTerm: state.search.search_term,
+    isLoading: state.search.isLoading,
     popularTables: state.popularTables,
     tables: state.search.tables,
     users: state.search.users,
--- a/amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
+++ b/amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
@@ -35,6 +35,7 @@ describe('SearchPage', () => {
   const setup = (propOverrides?: Partial<SearchPageProps>) => {
     const props: SearchPageProps = {
       searchTerm: globalState.search.search_term,
+      isLoading: false,
       popularTables: globalState.popularTables,
       dashboards: globalState.search.dashboards,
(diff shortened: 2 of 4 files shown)


ttannis:
👀

amundsen_application/static/js/ducks/search/reducer.ts
```diff
@@ -64,25 +72,39 @@ export default function reducer(state: SearchReducerState = initialState, action
       return {
         ...state,
         search_term: action.term,
+        isLoading: true,
```
danwom: Looks like only the `SearchAll` action will trigger the loading spinner. Do you also want to show the spinner for a `SearchResource` action as well? Generally this will happen when a user clicks on a paginated result. It may or may not be a good user experience.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -250,15 +252,24 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
       );
   };
 
+  renderTabs = () => {
```
danwom: I think `renderContent` or `renderSearchResults` would be better names, since not every section here has tabs. This will also be where the default homepage content will be rendered as well.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -8,6 +8,7 @@ import { RouteComponentProps } from 'react-router';
 
 import SearchBar from './SearchBar';
 import SearchList from './SearchList';
+import LoadingSpinner from 'components/common/LoadingSpinner';
```
danwom: This might be out of scope, but the LoadingSpinner has a fixed height and width that may not look great in all places. Maybe look into how to make it smaller/bigger depending on the context.

amundsen_application/static/js/components/SearchPage/index.tsx
ttannis: +1 for `renderContent`

amundsen_application/static/js/ducks/search/reducer.ts
```diff
@@ -64,25 +72,39 @@ export default function reducer(state: SearchReducerState = initialState, action
...
+      };      
+    case GetPopularTables.SUCCESS:
+    case GetPopularTables.FAILURE:
+      return {
+        ...state,
+        isLoading: false,
+      };
     default:
-      return state;
+      return {
```
ttannis: Why the change?

amundsen_application/static/js/ducks/search/reducer.ts
```diff
@@ -40,6 +47,7 @@ export function searchResource(resource: ResourceType, term: string, pageIndex:
 
 const initialState: SearchReducerState = {
   search_term: '',
+  isLoading: true,
```
ttannis: Can we default to `false`, or do you currently end up seeing the no results message followed by the spinner followed by the actual search results when it completes? 

Would that also remove the need to have the following code?:
```
 case GetPopularTables.SUCCESS:
 case GetPopularTables.FAILURE:
      return {
        ...state,
        isLoading: false,
      };
```

danwom:
I just realized this about your previous PR but when a user uses the pagination controls, `onPaginationChange` fetches the new page data with `this.props.searchResource`, and then updates the page URL with `this.updatePageUrl`. The URL update then triggers `componentDidUpdate` which then calls `this.props.searchAll`. So this actually makes two requests for the same data.


This isn't a big deal now, but this might break some behavior integrations with multiple search resource types. (Someone recently implemented Dashboards)

amundsen_application/static/js/fixtures/globalState.ts
```diff
@@ -41,6 +41,7 @@ const globalState: GlobalState = {
   ],
   search: {
     search_term: 'testName',
+    isLoading: true,
```
ttannis: Change to false if you end up making the default value `false`.

amundsen_application/static/js/ducks/search/reducer.ts
ttannis: Similar comment here, I see you reset the loading state when `SearchResource` completes/fails but didn't set the loading state to true when `SearchResource` is called. I'd be interested to checkout what this looks and feels like when we use the LoadingSpinner for both cases.

amundsen_application/static/js/components/SearchPage/index.tsx
ryanlieu: I think this is probably out of scope for this PR, will file a follow up task

amundsen_application/static/js/components/SearchPage/index.tsx
ryanlieu: :+1:

amundsen_application/static/js/ducks/search/reducer.ts
ryanlieu: oops, leftover code from some testing 😅

ryanlieu:
@danwom I'm going to lumping in a fix to that multiple request bug to this MR as it was causing the spinner to show up even on pagination (which wasn't intended).

amundsen_application/static/js/ducks/search/reducer.ts
ryanlieu: just tried it out, that works and is way cleaner!

amundsen_application/static/js/ducks/search/reducer.ts
ryanlieu: updating so searchresource action also sets isLoading to true

amundsen_application/static/js/ducks/search/reducer.ts
```diff
@@ -61,27 +63,34 @@ export default function reducer(state: SearchReducerState = initialState, action
   switch (action.type) {
     // Updates search term to reflect action
     case SearchAll.ACTION:
+    case SearchResource.ACTION:
```
ttannis: Because searching a resource doesn't change the search_term, there's no need to update it here. We can keep out logic cleaner and more intentional by instead having:
```
case SearchResource.ACTION:
    return {
        ...state,
        isLoading: true,
     }
```

ryanlieu:
👀 @ttannis @danwom this should be gtg with tests

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -273,13 +270,47 @@ describe('SearchPage', () => {
       wrapper.instance().componentDidUpdate(mockPrevProps);
     });
 
-    it('calls setState', () => {
-      expect(setStateSpy).toHaveBeenCalledWith({ selectedTab: ResourceType.table });
-    });
+    describe('called with a new search term', () => {
+      it('calls setState', () => {
```
ttannis: We call `setState` regardless of what happens with the search term so it's misleading to bundle that into this case. These two checks can be separate tests and don't have to be bundled into the same block:
```
it('calls setState', () => {
    expect(setStateSpy).toHaveBeenCalledWith({ selectedTab: ResourceType.table });
});

it('calls searchAll if called with a new search term', () => {
   expect(searchAllSpy).toHaveBeenCalledWith(mockSanitizedUrlParams.term, mockSearchOptions);
});
```
 
However one thing we can keep is the fact that you added some more explicit titles for the tests, to help clarify the conditional logic that we're testing.

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -273,13 +270,47 @@ describe('SearchPage', () => {
...
+      });
 
-    it('calls searchAll', () => {
-      expect(searchAllSpy).toHaveBeenCalledWith(mockSanitizedUrlParams.term, mockSearchOptions);
+      it('calls searchAll', () => {
+        expect(searchAllSpy).toHaveBeenCalledWith(mockSanitizedUrlParams.term, mockSearchOptions);
+      });
     });
+
+    describe('called with a new page but with the same search term', () => {
```
ttannis: We can condense this by leveraging the setup that was used initially for these tests. This is because this test case doesn't necessarily need a new setup, we only have to call `wrapper.instance().componentDidUpdate()` with some new `mockPrevProps` and verify what happens with the `searchAllSpy`.
```
it('does not call searchAll if called with the same search term', () => {
    /* assuming we use the same setup from the previous tests and don't have to write all the boiler plate again */
    searchAllSpy.mockClear(); // because another test will have called it.
     const mockPrevProps = {
          location: {
            search: '/search?searchTerm=current&current=table&pageIndex=1', 
            pathname: 'mockstr',
            state: jest.fn(),
            hash: 'mockstr',
          }
     };
    wrapper.instance().componentDidUpdate(mockPrevProps);
    expect(searchAllSpy).not.toHaveBeenCalled();
});
```

The above should work, but let me know if you run into any issues.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -96,7 +98,10 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
       const { searchTerm, pageIndex, selectedTab } = params;
       const { term, index, currentTab } = this.getSanitizedUrlParams(searchTerm, pageIndex, selectedTab);
       this.setState({ selectedTab: currentTab });
-      this.props.searchAll(term, this.createSearchOptions(index, currentTab));
+      const prevTerm = qs.parse(prevProps.location.search).searchTerm;
```
danwom: Rather than re-parsing the previous props location, you can pull in the prev state and use that value. Interface for this function is:

`componentDidUpdate(prevProps, prevState, snapshot)`

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -606,6 +637,50 @@ describe('SearchPage', () => {
     });
   });
 
+  describe('renderContent', () => {
+    let content;
```
ttannis: I see you're not actually using the `content` variable in your tests so we can remove those.

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -606,6 +637,50 @@ describe('SearchPage', () => {
...
+    describe('popular tables', () => {
+      beforeAll(() => {
+        const setupResult = setup({ searchTerm: '' });
+        props = setupResult.props;
+        wrapper = setupResult.wrapper;
+        content = shallow(wrapper.instance().renderContent());
+      });
+      
+      it('renders popular tables', () => {
+        expect(wrapper.instance().renderContent()).toEqual(wrapper.instance().renderPopularTables());
```
ttannis: nit: Another approach would be to spy on `renderPopularTables` and `renderSearchResults` and check that they were called. Imagine if we had thousands of tests with several hundred of them doing similar checks -- spying on methods whenever we can is expected be faster than if those hundreds of tests were rendering content and checking the output each time. Fine to merge as is though if you like.

amundsen_application/static/js/components/SearchPage/index.tsx
ryanlieu: oh right, I could just set it prevProps.searchTerm right?

danwom: There's an important distinction between `state` and `props`. With the way we're using `Redux/Saga` the 'container' pulls values from the `globalState` and sends it as `props` on the `component`. 

Sometimes the component can use the props as-is, but other times it needs to consume props to update its internal `state`.

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -606,6 +637,50 @@ describe('SearchPage', () => {
     });
   });
 
+  describe('renderContent', () => {
+    let content;
+    let props;
+    let wrapper;
+    describe('popular tables', () => {
+      beforeAll(() => {
```
ttannis: nit: Since each we only have one case for each test, we don't necessarily need a new `describe` block and a `beforeAll`, we could also have done:

```
it('renders popular tables if searchTerm is empty', () => {
    const setupResult = setup({ searchTerm: '' });
    props = setupResult.props;
    expect(wrapper.instance().renderContent()).toEqual(wrapper.instance().renderPopularTables());
});
```
Note the addition of clarifying the conditional logic we're testing in the description as well.

ryanlieu:
👀 @ttannis @danwom addressed comments, should be good to go.

PR #185, opened 2019-05-31
Unit tests + Code cleanup for ducks directory (redux state management + API calls)

ttannis:
### Summary of Changes

This PR addresses specific pieces tech debt in our code, described below. Changes have been applied to `ducks/allTags`, `ducks/announcements`, and `ducks/bookmarks` for initial review. Upcoming PRs will complete this work by doing the following:
1. Apply the code cleanup across the rest of the `ducks` folders
2. Best effort code coverage across the rest of the `ducks` folders

##### Types
1. The interfaces and enums we define throughout the application current fall into three main categories -- component interfaces, ducks interfaces, and "common" interfaces that tend to be the type definitions for objects that are handled/processed by the Redux layer and displayed/used by React components. e.g.: `Tag`. For "common" interfaces, this PR defines a separate `interfaces` directory. This change is meant to help decrease our current problem of "spaghetti imports" as the "common" interfaces have not had a clear place to reside.

##### Axios API Calls
1. Any types for the `response.data` object on AxiosResponses have been moved from the reducer to the api file, where it is actually used.
2. Catching errors from the axios api calls have been removed. The reason is that the sagas that call these methods are set up to catch these errors to fire `ActionType.FAILURE` actions, so these errors need to either propagate to the saga OR get re-thrown after being caught. We prefer the former unless there is a use case where catching and re-throwing the error is necessary.

##### Reducers & Actions & Sagas
1. Types for each reducers actions have been removed, e.g. `AllTagsReducerAction`. The reason for this is that it was incorrect for us to assign such a specific type to this action, because actions of all types are technically passed through each reducer. If we wanted to assign a type to the `action` parameter of reducers it would have to be a generic type that can applied to all actions. If we want to ensure type safety with our actions, there is a library https://github.com/piotrwitek/typesafe-actions that we can consider. 
2. I've opted to change the `ActionType.ACTION` enum to `ActionType.REQUEST` to be more intuitive. (i.e. `REQUEST`, `SUCCESS`, `FAILURE` feels like a better set of names for this trio of options)
3. If a `payload` is involved in a request or response, for consistency this payload will always be an object with key-value pairs.

##### Unit Tests
1. Added unit tests for actions, reducers, and sagas. A new library -- `redux-saga-test-plan` -- was introduced to reduce the complexity of testing saga iterators.
2. Added unit tests for the custom logic involved in our `axios` api calls.

### Tests
Unit tests were added to cover `ducks/allTags`, `ducks/announcements`, and `ducks/bookmarks`. The unit test thresholds were also updated.

### CheckList
- [x] [n/a] In case of new functionality, my PR adds documentation that describes how to use it.


ttannis:
This PR is currently a work in progress. Changes have only only applied to one set of files to allow for some discussion before completing the task.

amundsen_application/static/js/components/BrowsePage/index.tsx
```diff
@@ -8,12 +8,14 @@ import './styles.scss';
...
 import LoadingSpinner from 'components/common/LoadingSpinner';
 import TagInfo from 'components/Tags/TagInfo';
-import { Tag } from 'components/Tags/types';
+import { TagsInterface } from 'interfaces/Tags';
 
 import { GlobalState } from 'ducks/rootReducer';
 import { getAllTags } from 'ducks/allTags/reducer';
 import { GetAllTagsRequest } from 'ducks/allTags/types';
 
+type Tag = TagsInterface.Tag;
```
danwom: I like splitting out the common types in to the interfaces folder, but I have some reservations about the namespaces. It seems like we have to import the namespace, then redefine each type for local usage.

It seems like the main benefit from this would be to avoid naming collisions, but I'm not sure how frequent those will occur, especially given that we should be importing from the common interfaces folder.

amundsen_application/static/js/ducks/allTags/api/v0.ts
```diff
@@ -1,17 +1,15 @@
-import axios, { AxiosResponse, AxiosError } from 'axios';
-
-import { AllTagsResponse } from '../types';
+import axios, { AxiosResponse } from 'axios';
 
 import { sortTagsAlphabetical } from 'ducks/utilMethods';
+import { TagsInterface } from 'interfaces/Tags';
+
+export type AllTagsResponseAPI = {
```
danwom: I'm all in favor for keeping the API response types inside the API. I wonder if there's a way to enforce the import rules-- so none of the API types are imported except for the unit tests.

If we are using namespaces, we could create a namepsace just for testing and export that. Not sure about syntax but something like


```
type AllTagsResponseAPI = { ... } // private

export namespace TagsTestInterface {
  export AllTagsResponseAPI;
}
```
In this way the AllTagsResponseAPI can be imported in Tests, and anywhere else that tries to import this will have some janky looking code that is obviously very `import TagsTestInterface.AllTagsResponseAPI`.

amundsen_application/static/js/components/BrowsePage/index.tsx
ttannis: I've moved away from namespaces entirely because even though folks still to use them, looking towards the future ES6 modules are preferred over namespaces. Folks can still use namespaces within modules if they want to quickly deal with name conflicts, but I think it will be best to not structure such a large change around a feature that sounds like the industry is moving away from.

amundsen_application/static/js/ducks/allTags/api/v0.ts
ttannis: Open to any further investigation surrounding how we can clean up which interfaces are available globally. Yet this PR will not contain any of that work since it's a separate endeavor.

ttannis:
Ready for review. @danwom as discussed we'll use this PR to make sure we're all set on the current code cleanup endeavors + tests.

amundsen_application/static/js/ducks/allTags/types.ts
```diff
@@ -1,22 +1,16 @@
...
-  msg: string;
-  tags: Tag[];
-}
-
-/* getAllTags */
 export enum GetAllTags {
-  ACTION = 'amundsen/allTags/GET_ALL_TAGS',
-  SUCCESS = 'amundsen/allTags/GET_ALL_TAGS_SUCCESS',
-  FAILURE = 'amundsen/allTags/GET_ALL_TAGS_FAILURE',
+  REQUEST = 'amundsen/allTags/GET_REQUEST',
```
danwom: Are we changing `ACTION` -> `REQUEST` across the board or just for HTTP requests?

amundsen_application/static/js/ducks/bookmark/reducer.ts
```diff
@@ -1,81 +1,71 @@
...
   myBookmarksIsLoaded: false,
   bookmarksForUser: [],
 };
 
- export default function reducer(state: BookmarkReducerState = initialState, action: BookmarkReducerAction): BookmarkReducerState {
+ export default function reducer(state: BookmarkReducerState = initialState, action): BookmarkReducerState {
   switch(action.type) {
     case RemoveBookmark.SUCCESS:
-      const { resourceKey } = action.payload;
+      const { resourceKey } = (<RemoveBookmarkResponse>action).payload;
```
danwom: nit: There's probably a way to let TS infer `action` type without casting. Not sure what's causing this to fail here.

amundsen_application/static/js/ducks/bookmark/tests/index.spec.ts
```diff
@@ -0,0 +1,308 @@
+import { expectSaga, testSaga } from 'redux-saga-test-plan';
+import * as matchers from 'redux-saga-test-plan/matchers';
+import { throwError } from 'redux-saga-test-plan/providers';
+
+import { Bookmark, ResourceType } from 'interfaces';
+
+import { addBookmark as addBkmrk, getBookmarks as getBkmrks, removeBookmark as removeBkmrk } from '../api/v0';
```
danwom: Seems like we have some naming collisions here. Why not use `addBookmarkAPI`?

amundsen_application/static/js/ducks/allTags/types.ts
ttannis: I was going to do it across the board. Any concerns?

amundsen_application/static/js/ducks/bookmark/reducer.ts
ttannis: The type checker doesn't fail without this specification. Rather, our code isn't type safe without this specification. Without it I could write something like `const { resourceKey } = action.payload.invalidAttribute` and the type checker would not catch this mistake.

I believe there are other approaches, I just haven't figured out the best way to leverage the TypeScripts ability to infer types when using `switch/case` or if the support is fully there yet. Another hack could be to use `if` statement here, but I opted for the casting. When I'm done with this work, I can open issue for a better/cleaner approach to type safe actions.

amundsen_application/static/js/ducks/bookmark/tests/index.spec.ts
ttannis: No solid reason, I just realized re-naming too many things would further obscure the actual content/intent of the PR so I opted for using an alias in the one file where there was a naming collision.

amundsen_application/static/js/ducks/allTags/types.ts
danwom: No concerns here.

ttannis:
@danwom let me know if you have any followups to the comments I just posted before I move on to the rest of the work. Thanks!

PR #190, opened 2019-06-04
Home page component

ryanlieu:
### Summary of Changes

Slightly larger PR that encapsulates the changes needed to create a separate home page component. Work is roughly split into three parts:

- Making SearchBar a "smart" (container) component vs. a "dumb" (presentational) component so that it can be re-used on multiple pages/locations (for example in case we want to put the search bar into the nav bar) 
- Putting SearchBar, Bookmarks, and PopularTables components into a new Homepage component that can be reached by navigating to /
- Fixing the breadcrumb component on TableDetails page to intelligently route back to search results (if table was accessed through search) or home (if table was accessed by directly loading the page or navigating from bookmarks/popular tables). 

### Tests

Will update tests after implementation gets looked over.

### Documentation

### CheckList


ryanlieu:
@ttannis @danwom this PR is good to review for implementation. Will write/update tests in parallel

amundsen_application/static/js/components/HomePage/index.tsx
```diff
@@ -0,0 +1,67 @@
+import * as React from 'react';
```
ttannis: nit: follow a loose import hierarchy of 3rd party imports (e.g. react/redux) --> absolute imports (e.g. components/ducks) --> relative imports (e.g. ../someFile)

amundsen_application/static/js/components/HomePage/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export type HomePageProps = DispatchFromProps & RouteComponentProps<any>;
+
+export class HomePage extends React.Component<HomePageProps> {
+  constructor(props) {
+    super(props);
+  }
+
+  createSearchOptions = (pageIndex: number, selectedTab: ResourceType) => {
```
ttannis: Why is this method here?

amundsen_application/static/js/components/HomePage/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+      userIndex: (selectedTab === ResourceType.user) ? pageIndex : 0,
+      tableIndex: (selectedTab === ResourceType.table) ? pageIndex : 0,
+    };
+  };
+  
+  componentDidMount() {
+    this.props.searchReset();
+  }
+
+  onSearchBarSubmit = (searchTerm: string): void => {
```
ttannis: Why is this method here?

amundsen_application/static/js/components/HomePage/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  };
+
+  render() {
+    return (
+      <>
+        <div className="container home-page">
+          <div className="row">
+            <div className="col-xs-12 col-md-offset-1 col-md-10">
+              <div className="search-list-container">
+                <SearchBar />
```
ttannis: `SearchBar` should be outside of the `.search-list-container` element. `BookMarkList` & `PopularTables` should also be in their own `<div>` as well with whatever classname we want to apply to create the appropriate spacing in between them as requested by Design. See next comment in `styles.css`.

amundsen_application/static/js/components/HomePage/styles.scss
```diff
@@ -0,0 +1,13 @@
+@import 'variables';
+
+.home-page {
+  .search-list-container {
```
ttannis: This style should have it's own name. In addition we have to increase the spacing between each of the lists on this page as requested by Design.

amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
```diff
@@ -11,11 +11,16 @@ import {
   SYNTAX_ERROR_PREFIX,
   SYNTAX_ERROR_SPACING_SUFFIX,
 } from './constants';
+import { RouteComponentProps, withRouter } from 'react-router';
```
ttannis: nit: Same comment about import order.

amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
```diff
@@ -11,11 +11,16 @@ import {
...
+import { GlobalState } from 'ducks/rootReducer';
+import { connect } from 'react-redux';
+
+export interface StateFromProps {
+  searchTerm: string;
+}
 
 export interface SearchBarProps {
```
ttannis: You can now call this `OwnProps` and declare:
`export type SearchBarProps = StateFromProps & OwnProps & RouteComponentProps<any>;`
and then be able to keep:
 `export class SearchBar extends React.Component<SearchBarProps, SearchBarState>`

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -258,7 +243,7 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
       <div className="container search-page">
         <div className="row">
           <div className="col-xs-12 col-md-offset-1 col-md-10">
-            <SearchBar handleValueSubmit={ this.onSearchBarSubmit } searchTerm={ searchTerm }/>
+            <SearchBar />
```
ttannis: We can also remove `onSearchBarSubmit` from this file as well as it's test.

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -126,6 +127,16 @@ export class TableDetail extends React.Component<TableDetailProps & RouteCompone
     })
   };
 
+  getBreadcrumb() {
```
ttannis: nit: `renderBreadcrumb` would be a more appropriate name.

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -330,7 +341,7 @@ export class TableDetail extends React.Component<TableDetailProps & RouteCompone
     } else {
         innerContent = (
           <div className="container table-detail">
-            <Breadcrumb path='/' text='Search Results'/>
+            { this.getBreadcrumb() }
```
ttannis: I'm fine with this approach for now, if we merge it in as is we should file a ticket to shift to the the alternative of putting this logic in `Breadcrumb` since we'll have to use it on the `ProfilePage` as well. cc @danwom since you're working on getting the `ProfilePage` ready.

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -376,6 +387,7 @@ export const mapStateToProps = (state: GlobalState) => {
     isLoading: state.tableMetadata.isLoading,
     statusCode: state.tableMetadata.statusCode,
     tableData: state.tableMetadata.tableData,
+    searchTerm: state.search.search_term,
```
ttannis: nit: maintain alphabetical order for object keys.

amundsen_application/static/js/index.tsx
```diff
@@ -17,7 +17,8 @@ import NavBar from './components/NavBar';
 import NotFoundPage from './components/NotFoundPage';
 import Preloader from "components/common/Preloader";
 import ProfilePage from './components/ProfilePage';
-import SearchPage from './components/SearchPage';
+import SearchPage from './components/SearchPage'; 
+import HomePage from './components/HomePage'
```
ttannis: nit: place `HomePage` higher up in alphabetical order, and looks like I missed this file when we switched to absolute imports. We can remove the relative imports in this file and switch to:
`import Foo from 'components/Foo'` and `import Thing from 'ducks/pathToThing'`

ryanlieu:
@ttannis this PR should be good to look over again, I've addressed your comments in the subsequent commits. 

One quick thing I was wondering about was the hard coding of the resource type in some of the searches. With the changes to search bar and breadcrumb, redirects are hard coded to selectedTab=table as there is no global redux store for selectedTab and page. This is fine for now as there's no other tabs we're currently displaying. I think it'd make sense to file a follow up task to add those two search variables to the global store a la searchTerm, but lmk if I should address that in this PR.

amundsen_application/static/js/components/HomePage/styles.scss
```diff
@@ -0,0 +1,13 @@
+@import 'variables';
+
+.home-page {
```
ttannis: Would be curious to see a screenshot of what this looks like -- is the first `SearchBar` pushed further down on this page than it is on the `SearchPage`? Is the `BookmarkList` further away fro the `SearchBar` on this page as well?

ttannis:
> @ttannis this PR should be good to look over again, I've addressed your comments in the subsequent commits.
> 
> One quick thing I was wondering about was the hard coding of the resource type in some of the searches. With the changes to search bar and breadcrumb, redirects are hard coded to selectedTab=table as there is no global redux store for selectedTab and page. This is fine for now as there's no other tabs we're currently displaying. I think it'd make sense to file a follow up task to add those two search variables to the global store a la searchTerm, but lmk if I should address that in this PR.

That was an intentional decision for the default selected tab to be table, and the default pageIndex to be 0.

amundsen_application/static/js/components/HomePage/styles.scss
ryanlieu: oh whoops shouldn't be applying the home-element-container class to SearchBar, will attach a screenshot of the fixed version

ryanlieu: Screen Shot 2019-06-11 at 2 21 45 PM

amundsen_application/static/js/components/common/Breadcrumb/index.tsx
```diff
@@ -1,14 +1,23 @@
...
 }
 
-const Breadcrumb: React.SFC<BreadcrumbProps> = ({ path, text }) => {
+export type BreadcrumbProps = StateFromProps;
+
+const Breadcrumb: React.SFC<BreadcrumbProps> = (props) => {
+  let path = '/';
+  let text = 'Home';
+  if (props.searchTerm) {
+    path = `/search?searchTerm=${props.searchTerm}&selectedTab=table&pageIndex=0`
```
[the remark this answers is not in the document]
ryanlieu: able to fix this with a one-liner, will include in this PR. Just doing a check on componentDidMount to see if term is the same as the searchTerm in redux. If it is we can just not call searchAll as the results for that searchTerm are stored in redux already.

ryanlieu:
@ttannis should be good to go for a final implementation pass, working on unit tests atm

amundsen_application/static/js/components/HomePage/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+    super(props);
+  }
+  
+  componentDidMount() {
+    this.props.searchReset();
+  }
+
+  render() {
+    return (
+      <>
```
danwom: The wrapper `<>` element is useful when you have multiple "root" elements on a JSX node. For example
```
// Not allowed
(
  <h1> header </h1>
  <body> body</body>
)

// Allowed
(
  <>
    <h1> header </h1>
    <body> body</body>
  </>
)


```

amundsen_application/static/js/components/common/Breadcrumb/index.tsx
```diff
@@ -1,14 +1,28 @@
...
+  searchTerm: string;
+}
+
+export type BreadcrumbProps = OwnProps & StateFromProps;
+
+export const Breadcrumb: React.SFC<BreadcrumbProps> = (props) => {
+  let path = props.path;
+  let text = props.text;
+  if (props.searchTerm) {
+    path = `/search?searchTerm=${props.searchTerm}&selectedTab=table&pageIndex=0`
```
danwom: Also worth considering: You could do `path = 'search?searchTerm=${props.searchTerm}'` and leave `selectedTab` and `pageIndex` as empty. Then the `SearchPage` could use a default value for these.

amundsen_application/static/js/components/HomePage/index.tsx
ttannis: i.e. you don't need the `<>` here

amundsen_application/static/js/components/HomePage/tests/index.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+import { shallow } from 'enzyme';
+
+import { mapDispatchToProps, HomePage, HomePageProps } from '../';
+
+import SearchBar from 'components/SearchPage/SearchBar';
+import BookmarkList from 'components/common/Bookmark/BookmarkList';
+import PopularTables from 'components/common/PopularTables';
+
+describe('SearchBar', () => {
```
ttannis: change title to 'HomePage'

amundsen_application/static/js/components/HomePage/tests/index.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const setup = (propOverrides?: Partial<HomePageProps>) => {
+    const props: HomePageProps = {
+      searchReset: jest.fn(),
+      history: {
```
ttannis: worthy mention: I'll see if i have time to look into how to better write tests that leverage the router. This works but I recognize it's pretty inconvenient.

amundsen_application/static/js/components/common/Breadcrumb/index.tsx
ryanlieu: the only issue with this is that the back button then requires two clicks to go back as the selectedTab and pageIndex are injected into the url. Will address this in a follow up PR.

amundsen_application/static/js/components/HomePage/tests/index.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+  let wrapper;
+  beforeAll(() => {
+    const setupResult = setup();
+    props = setupResult.props;
+    wrapper = setupResult.wrapper;
+  });
+
+  describe('render', () => {
+    it('contains Searchbar, BookmarkList, and PopularTables', () => {
+      wrapper.instance().render();
```
ttannis: Doesn't this pass without needing to call `render()`?

amundsen_application/static/js/components/ProfilePage/index.tsx
```diff
@@ -86,7 +86,7 @@ export class ProfilePage extends React.Component<ProfilePageProps> {
         <div className="container profile-page">
           <div className="row">
             <div className="col-xs-12 col-md-offset-1 col-md-10">
-              <Breadcrumb path='/' text='Search Results'/>
+              <Breadcrumb path="/" text="Home" />
```
ttannis: This one was fine to leave as `<Breadcrumb />`, but since we haven't exposed this in search yet it makes sense to leave it as directing to `Home`. Can you add a comment about removing the hardcoded path to `Home` when we are ready to expose this in production?

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -680,10 +645,7 @@ describe('SearchPage', () => {
 
     it('renders SearchBar with correct props', () => {
       const { props, wrapper } = setup();
-      expect(wrapper.find(SearchBar).props()).toMatchObject({
-        handleValueSubmit: wrapper.instance().onSearchBarSubmit,
-        searchTerm: props.searchTerm,
-      });
+      expect(wrapper.find(SearchBar));
```
ttannis: Let's go with `expect(wrapper.find(SearchBar).exists()).toBeTruthy();` instead.

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -323,14 +322,14 @@ export class TableDetail extends React.Component<TableDetailProps & RouteCompone
     } else if (this.state.statusCode === 500) {
       innerContent = (
         <div className="container error-label">
-          <Breadcrumb path='/' text='Search Results'/>
+          <Breadcrumb path="/" text="Home" />
```
ttannis: Was this a find/replace mistake? On the `TableDetail` page we want the logic of whether or not to go back to `Home` or `SearchPage` to be executed.

amundsen_application/static/js/components/common/Breadcrumb/index.tsx
ttannis: ^ not sure I understand this, but can we chat about it on slack or in person

amundsen_application/static/js/components/common/Breadcrumb/index.tsx
```diff
@@ -1,14 +1,30 @@
...
 }
 
-const Breadcrumb: React.SFC<BreadcrumbProps> = ({ path, text }) => {
+export interface StateFromProps {
+  searchTerm: string;
+}
+
+export type BreadcrumbProps = OwnProps & StateFromProps;
+
+export const Breadcrumb: React.SFC<BreadcrumbProps> = (props) => {
```
ttannis: nit: If you do:
`export const Breadcrumb: React.SFC<BreadcrumbProps> = ({ path, text, searchTerm ) => {`
then variables `path` and `text` and `searchTerm` will automatically be defined and pulled from the `props`. Then, you can also set the default `path` and `text` in the `defaultProps` on line 40. This will make it such that all you have to have in this section is:
```
if (searchTerm) {
    path = `/search?searchTerm=${props.searchTerm}&selectedTab=table&pageIndex=0`
    text = 'Search Results'
}
```

ryanlieu: :+1: yeah that's way cleaner

ryanlieu:
@ttannis just pushed that breadcrumb cleanup - going to 🚢after tests pass

PR #191, opened 2019-06-05
Use table 'key' in Metadata APIs

danwom:
### Summary of Changes
Previously all of our metadata APIs sent a set of fields `cluster`, `database`, `schema`, and `table` to identify the table record in metadata service. Those parameters were replaced by a single 'key'.

### Tests
Modified metadata API tests to include `key` rather than the other fields.

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/api/metadata/v0.py
+++ b/amundsen_application/api/metadata/v0.py
@@ -79,7 +79,8 @@ def get_table_metadata() -> Response:
     TODO: Define an interface for envoy_client
     """
     try:
-        table_key = get_table_key(request.args)
+        # table_key = get_table_key(request.args)
+        table_key = get_query_param(request.args, 'key')
         list_item_index = get_query_param(request.args, 'index')
         list_item_source = get_query_param(request.args, 'source')
 
@@ -155,7 +156,9 @@ def _update_table_owner(*, table_key: str, method: str, owner: str) -> Dict[str,
 def update_table_owner() -> Response:
     try:
         args = request.get_json()
-        table_key = get_table_key(args)
+        # table_key = get_table_key(args)
+        table_key = get_query_param(args, 'key')
+
         owner = get_query_param(args, 'owner')
 
         payload = jsonify(_update_table_owner(table_key=table_key, method=request.method, owner=owner))
@@ -197,7 +200,8 @@ def get_last_indexed() -> Response:
 def get_table_description() -> Response:
     try:
         table_endpoint = _get_table_endpoint()
-        table_key = get_table_key(request.args)
+        # table_key = get_table_key(request.args)
+        table_key = get_query_param(request.args, 'key')
 
         url = '{0}/{1}/description'.format(table_endpoint, table_key)
 
@@ -222,7 +226,8 @@ def get_table_description() -> Response:
 def get_column_description() -> Response:
     try:
         table_endpoint = _get_table_endpoint()
-        table_key = get_table_key(request.args)
+        # table_key = get_table_key(request.args)
+        table_key = get_query_param(request.args, 'key')
 
         column_name = get_query_param(request.args, 'column_name')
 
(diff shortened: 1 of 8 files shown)


amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -97,7 +98,9 @@ export class TableDetail extends React.Component<TableDetailProps & RouteCompone
     this.database = params ? params.db : '';
     this.schema = params ? params.schema : '';
     this.tableName = params ? params.table : '';
+
     this.displayName = params ? `${this.schema}.${this.tableName}` : '';
+    this.key = params ? `${this.database}://${this.cluster}.${this.schema}/${this.tableName}` : '';
```
danwom: Because of the way React Router's params work, the `/` is a delimiter so I could not just put the key as a URL param. I could use `encodeURIComponent` but that results in ugly URLs and potentially broken links in Slack.

As a compromise, this is the only place where a key is jerry rigged.

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -39,3 +26,74 @@ def marshall_table_partial(table: Dict) -> Dict:
...
+        'schema',
+        'source',
+        'table_description',
+        'table_name',
+        'table_readers',
+        'table_writer',
+        'tags',
+        'watermarks',
+        # 'last_updated_timestamp' Exists on the response from metadata but is not used.
+        # This should also be consolidated with 'last_updated_epoch' to have the same name and format.
```
danwom: Should clean up `last_updated_timestamp` and `last_updated_epoch`.
Most notably, this value is missing from `popular_tables` schema.

amundsen_application/api/metadata/v0.py
```diff
@@ -210,7 +155,8 @@ def _update_table_owner(*, table_key: str, method: str, owner: str) -> Dict[str,
 def update_table_owner() -> Response:
     try:
         args = request.get_json()
-        table_key = get_table_key(args)
+        table_key = get_query_param(args, 'key')
+
```
ttannis: nit nit: extra empty space?

amundsen_application/static/js/ducks/tableMetadata/api/helpers.ts
```diff
@@ -6,8 +6,8 @@ import { filterFromObj, sortTagsAlphabetical } from 'ducks/utilMethods';
  * Generates the query string parameters needed for requests that act on a particular table resource.
  */
 export function getTableParams(tableDataObject: TableMetadata | GetTableDataRequest): string {
```
ttannis: nit: We can rename this `getTableQueryParams` for clarity, or any other better name you can think of. Not sure if you feel the same way but looking at this method after some time I found this to be a confusing name

amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
```diff
@@ -104,17 +98,14 @@ export function metadataTableOwners(tableData: TableMetadata) {
 }
 
 /* TODO: Typing this method generates redux-saga related type errors that need more dedicated debugging */
-export function metadataUpdateTableOwner(action, tableData) {
+export function metadataUpdateTableOwner(action, tableData: TableMetadata) {
```
ttannis: nit: Can you add a `TODO` for putting the `key` on the action and using `action.key`?

amundsen_application/static/js/components/TableDetail/index.tsx
ttannis: Can you add a comment and link to the line of code in the python files that this has to match? Anything that would inform someone not to change this hardcoded pattern.

feng-tao:
Didn't look into much detail, but two questions:
1. Is it backward compatible?
2. Does it require API response changed in metadata?
3. Does it work with Atlas as the proxy for metadata?

danwom:
@feng-tao These changes only affect the interface between frontend-python and frontend-react.

The metadata service APIs always used `key` as the identifier. There are some instances metadata returns an object without `key` so frontend-python manually adds it in.

feng-tao:
I see. Thanks for the info @danwom 👍

amundsen_application/static/js/components/TableDetail/index.tsx
danwom: @feng-tao having trouble finding where this format is defined. Is it in databuilder or metadata?

PR #193, opened 2019-06-06
Create Generic Paginated Resource List

danwom:
### Summary of Changes
- Moved 'SearchList' to a generic 'ResourceList'
- Renamed 'BookmarkList' to 'MyBookmarks'
- Added pagination to 'ResourceList'
- Converted usages in 'PopularTables', 'SearchPage', and 'MyBookmarks'

### Tests
Updated tests to reflect changes.

### Documentation

### CheckList


amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -584,34 +584,23 @@ describe('SearchPage', () => {
...
-            paginationStartIndex: 0,
-          },
-        });
-      });
-
-      it('does not render Pagination if total_results <= RESULTS_PER_PAGE', () => {
-        expect(content.children().find(Pagination).exists()).toBeFalsy()
-      });
-
-      it('renders Pagination with correct props if total_results > RESULTS_PER_PAGE', () => {
```
danwom: Moved to `ResourceList` tests.

danwom:
@ttannis ready for review.

amundsen_application/static/js/components/common/ResourceList/index.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from 'react';
+import Pagination from 'react-js-pagination';
+import ResourceListItem from 'components/common/ResourceListItem';
+import { Resource } from 'components/common/ResourceListItem/types';
+import { ITEMS_PER_PAGE, PAGINATION_PAGE_RANGE } from "./constants";
+
+
+export interface ResourceListProps {
+  activePage: number;
```
danwom: One thing I've been mulling about was where to keep track of the `activePage` state. In some cases it needs to be managed externally (SearchPage) and in others it can managed internally in the `ResourceList` component (MyBookmarks).

I could make the `activePage` and `onPagination` props as optional. Any thoughts on this?

amundsen_application/static/js/components/common/PopularTables/constants.ts
```diff
@@ -2,3 +2,4 @@
 export const POPULAR_TABLES_INFO_TEXT = 'These are some of the most commonly accessed tables within your organization.';
 export const POPULAR_TABLES_LABEL = 'Popular Tables';
 export const POPULAR_TABLES_SOURCE_NAME = 'popular_tables';
+export const POPULAR_TABLES_PER_PAGE = 4;
```
ttannis: Why not a property on the component itself with a default value of 4?

amundsen_application/static/js/components/common/PopularTables/index.tsx
```diff
@@ -3,15 +3,20 @@ import * as React from 'react';
...
+} from './constants';
 import InfoButton from 'components/common/InfoButton';
-import SearchList from 'components/SearchPage/SearchList';
 
 import { getPopularTables } from 'ducks/popularTables/reducer';
 import { GetPopularTablesRequest, TableResource } from 'ducks/popularTables/types';
 import { GlobalState } from 'ducks/rootReducer';
 import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
+import ResourceList from "components/common/ResourceList";
```
ttannis: nit: single quotes

amundsen_application/static/js/components/common/Bookmark/MyBookmarks/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+  isLoaded: boolean;
+}
+
+export type MyBookmarksProps = StateFromProps;
+
+interface MyBookmarksState {
+  activePage: number;
+}
+
+export class MyBookmarks extends React.Component<MyBookmarksProps, MyBookmarksState> {
```
ttannis: Why the rename? Is there some other component that you imagine being called `BookmarkList`? I think `BookmarkList` is a more appropriate name because it names the component simply based on what it is, without any further implications. 

Is the rename because we anticipate having logic that will apply based on whether or not this bookmark list is for the current user vs another user? If so, then I'd propose that the bookmark list component have whatever boolean properties it needs to do any conditional rendering. Let me know your thoughts.

amundsen_application/static/js/components/common/ResourceList/index.tsx
danwom: Yeah, I think I'll make all of the pagination stuff optional in addition to providing some default state management. Will push a PR shortly.

amundsen_application/static/js/components/common/PopularTables/constants.ts
danwom: This follows the pattern established for the search page and bookmarks. This might also be easier to import and test.

ttannis: I see, also just remembered that since this isn't a reusable component and only appears in one use case, its fine to be a constant and if anything would be changed from the application config.

amundsen_application/static/js/components/common/Bookmark/MyBookmarks/index.tsx
danwom: In my head the `ProfilePage` will be responsible for fetching bookmarks, owned, and frequently used resources and put them into a `ResourceList` without making separate components for each of them.

```
class ProfilePage {
  ... 
  componentDidMount() {
    this.props.getBookmarksForUser(userId);
    this.props.getFrequentlyUsedForUser(userId);
    this.props.getOwnedForUser(userId);
  }

  getTabsInfo() {
    return [ 
       { content: <ResouceList items={ this.props.owned } ...>, title: "owned", }
       { content: <ResouceList items={ this.props.bookmarks } ...>, title: "bookmarks" }
       { content: <ResouceList items={ this.props.frequentlyUses } ...>, title: "frequently uses" }
    ]
  }
}
```

In this case we only need one `MyBookmarks` component which connects the `globalState.bookmarks.myBookmarks`. I'm also removing the state management from this component so it should be a relatively light weight.

danwom: In an alternative version of this, we would create 3 very similar components that hit different endpoints and are responsible for fetching their own data. At this point I don't think it's worth creating a separate component for each tab, but I may change my mind.

amundsen_application/static/js/components/common/PopularTables/index.tsx
danwom: I think my IDE auto imports with double quotes. I'll have to look at settings.

ttannis: If that doesn't work out, we can also check to see if our linter can auto fix.

amundsen_application/static/js/components/common/Bookmark/MyBookmarks/index.tsx
ttannis: I see. This resolves my concern of whether or not there would be some duplicate bookmarks component. Still not attached to the name `MyBookmarks`, but we have some other questionable names and at least this one is tied to the piece of globalState that it uses. Thanks for the explanation.

amundsen_application/static/js/components/common/Bookmark/MyBookmarks/constants.ts
```diff
@@ -1,5 +1,4 @@
-export const ITEMS_PER_PAGE = 4;
-export const PAGINATION_PAGE_RANGE = 10;
-
 export const BOOKMARK_TITLE = "My Bookmarks";
+export const BOOKMARKS_PER_PAGE = 4;
 export const EMPTY_BOOKMARK_MESSAGE = "You don't have any bookmarks. Use the star icon to save a bookmark.";
+export const MY_BOOKMARKS_SOURCE_NAME = "Bookmarks";
```
ttannis: nit: I think we're currently using `snake_case` for these, so it would be `bookmarks`.

amundsen_application/static/package-lock.json
```diff
@@ -1074,13 +1074,15 @@
           "version": "1.0.0",
```
ttannis: Did you change your local npm/node versions recently?

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -142,7 +139,7 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
   };
 
   onPaginationChange = (pageNumber: number): void => {
-    const index = pageNumber - 1;
+    const index = pageNumber;
```
ttannis: nit: just pass in `pageNumber`

amundsen_application/static/js/components/common/ResourceList/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { Resource } from 'interfaces';
+import { ITEMS_PER_PAGE, PAGINATION_PAGE_RANGE } from './constants';
+
+
+export interface ResourceListProps {
+  items: Resource[];
+  source: string;
+
+  // The following props are only used if 'paginate' is set to true
+  paginate?: boolean;
```
ttannis: nit: Another common practice for flexible, re-useable components is introduce a property that takes an object for all of the properties for a particular feature. i.e. we'd have 
`paginate: boolean`
`paginateOptions: { ...all those other props }`. 

Fine to merge without it, and perhaps the next time we end up wanting to leverage for of the 3rd party `Pagination` options, we can make that switch.

amundsen_application/static/js/components/common/ResourceList/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+interface ResourceListState {
+  activePage: number;
+}
+
+class ResourceList extends React.Component<ResourceListProps, ResourceListState> {
+  public static defaultProps: Partial<ResourceListProps> = {
+    paginate: false,
+    activePage: 0,
+    itemsPerPage: ITEMS_PER_PAGE,
+    isFullList: true,
```
ttannis: Can you clarify the purpose of this prop? It looks like it's a boolean for whether or not the `items` list is the full list of items to show, but I got confused here:
```
 if (paginate && isFullList) { <-- full list yet we only show a portion?
      itemsToRender = items.slice(startIndex, startIndex + itemsPerPage);
 }
```

Is the `paginate` boolean not enough?

amundsen_application/static/js/components/common/ResourceList/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+      this.setState({ activePage: this.props.activePage });
+    }
+  }
+
+  onPagination = (rawPageNum: number) => {
+    const activePage = rawPageNum - 1;
+    if (this.props.onPagination !== undefined) {
+      this.props.onPagination(activePage);
+    } else {
+      this.setState({ activePage });
```
ttannis: Does this not cause another `setState` top trigger in `componentDidUpdate`?

amundsen_application/static/js/components/common/ResourceList/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+    }
+  };
+
+
+  render() {
+    const { isFullList, items, itemsPerPage, paginate, source } = this.props;
+    const itemsCount = this.props.itemsCount || items.length;
```
ttannis: Why is `items.length` not sufficient?

amundsen_application/static/js/components/common/ResourceList/tests/index.spec.tsx
```diff
@@ -0,0 +1,108 @@
+import * as React from 'react';
+import Pagination from 'react-js-pagination';
+
+import { shallow } from 'enzyme';
+
+import { ResourceType } from 'interfaces';
+import ResourceListItem from 'components/common/ResourceListItem/index';
+import ResourceList, { ResourceListProps } from '../';
+
+describe('ResourceList', () => {
```
ttannis: Missing a test for `componentDidUpdate`

amundsen_application/static/package-lock.json
danwom: I upgraded to `node v10.15.3` and `npm 6.9.0` a while ago when there were some github issues related node compatibility. node 8 is on maintenance mode while node 10 is active.

amundsen_application/static/js/components/common/ResourceList/index.tsx
danwom: There's actually a difference between `paginate` and `isFullList` which adds a bit to the complexity. I considered creating two separate components because of this.

For `MyBookmarks` the `items` prop is the full list of bookmarks. Displaying page 2 means we render bookmarks 5 - 8, so we would slice off the appropriate bookmarks from our full list and render them.

On the `SearchPage` the `items` prop is limited to 10 since that is what we get from the API response. To display page 2, we render all 10 `items` to render results 11 - 20 of X,XXX.

I guess the term `isFullList` is somewhat unclear. I also considered `sliceItems` or something similar.

amundsen_application/static/js/components/common/ResourceList/index.tsx
danwom: It does, but the condition on 39 stops it from doing anything. This is another weird thing were the `activePage` can either be managed internally inside this component or externally a la `SearchPage`. This is another reason I considered splitting this into multiple components, but that ended up being pretty messy as well.

amundsen_application/static/js/components/common/ResourceList/index.tsx
danwom: For `SearchPage` we `items` is an array of at most 10 items, when there might be 3000 items to paginate over.

amundsen_application/static/js/components/common/ResourceList/index.tsx
danwom: I do like this pattern, but how do the default props work for the nested options? If you specify the `paginateOptions` with just one or two options, does it overwrite all of the fields?

```
defaultProps = {
  paginate: true,
  paginationOptions: {
    activePage: 0,
    isFullList: true,
    itemsPerPage: 4,
  }
}
```

```
<ResourceList 
  paginate={ true }
  paginateOptions={
    itemsPerPage: 5
  }
```

ttannis: ick, looks like React only supports defaultProps for top level properties. Here's a ticket for that request but it was closed, however some folks have posted workarounds: https://github.com/facebook/react/issues/2568. Wouldn't suggest wasting time on this though.

amundsen_application/static/js/components/common/ResourceList/index.tsx
ttannis: I'd go with `sliceItems`.

amundsen_application/static/js/components/common/ResourceList/index.tsx
ttannis: The check on line 39 compares the current props to previous props, given that on this line we update the state and not the props, how does the check stop `setState` from firing again?

amundsen_application/static/js/components/common/ResourceList/index.tsx
ttannis: I see. So if I understand correctly we have ~3 different cases for the ways we can display a ResourceList:
1. Total items given, all items rendered with no pagination (some future use case)
2. Total items given, slice of items rendered with pagination (MyBookmarks)
3. Slice of items given with total number of items, slice of items rendered with pagination (search results)

And what's confusing me a bit is that the flexibility is achieved by creating a set of optional properties, and different combinations of those properties render the desired results. I don't want to block progress if this might take too long to address but now I'm leaning towards something like:
`paginationType?:  valid values "none" (default), "fullList", "partialList"`
`fullListOptions?: { all the props needed to render this case }`
`partialListOptions?: { all the props needed to render this case }`
And then unfortunately all the nested props would have to be required due to how React handles nested default props. 

The names in my examples can probably be better, and what I've proposed may need to be tweaked in some way to account for all the use cases, but my thought process here is to see how we can make this component flexible yet easy to work with and less prone to bugs. 

It feels like if we set up our properties in a slightly less consolidated way, it will be easy to place our logic in 3 clear chunks for each case and render accordingly. We would lose some consolidation and be forced to have some nested props of the same name on the different options objects, but it will be clearer for the next person who comes along what needs to be modified if they want to make a change for a particular case. I'm fine tabling this conversation and filing a ticket.

amundsen_application/static/js/components/common/ResourceList/index.tsx
danwom: So there are two pathways for pagination and `activePage`.

1. The `activePage` is managed internally only (MyBookmarks, PopularTables). `onPagination` triggers `setState` on line 49. The check on line 39 doesn't run because `this.props.activePage` should be `undefined`.
2. `activePage` is managed externally (SearchPage). `onPagination` fires `this.props.onPagination` which updates the `activePage` on the external component. This in turn should update `this.prop.activePage` and trigger the `setState` in `componentDidUpdate`


I agree this behavior is unclear and convoluted. I'll try to see if we can simplify this somehow. Ideally I would like all page management to be done internally.

danwom:
Did some cleanup, hopefully this should be easier to understand.

For the rendered items, the user can pass in `allItems` or `slicedItems` (and `slicedItemsCount`). The component should automatically detect which of these fields is filled and should behave accordingly. I also made the state management a bit more straightforward. If `activePage` is passed in as a prop, it will always use the prop directly. Otherwise it will use `state.activePage` internally.

PR #195, opened 2019-06-13
Cleanup ducks folder

ttannis:
### Summary of Changes

In order to write unit tests for the `ducks` folder, we wanted to resolve some inconsistencies in patterns and improve our organization of common types. This PR applies the organizational changes introduced and discussed in https://github.com/lyft/amundsenfrontendlibrary/pull/185 to the rest of the `ducks` directory.

### Tests

Unit tests for `ducks` will be added in an upcoming PR. Jest thresholds for `components` were increased to reflect recent coverage improvements.

### Documentation

N/A

### CheckList


amundsen_application/static/js/components/ProfilePage/index.tsx
```diff
@@ -10,12 +10,14 @@ import Tabs from 'components/common/Tabs';
 
 import { GlobalState } from 'ducks/rootReducer';
 import { getUserById } from 'ducks/user/reducer';
-import { User, GetUserRequest } from 'ducks/user/types';
+import { GetUserRequest } from 'ducks/user/types';
+
+import { PeopleUser } from 'interfaces';
```
ttannis: `PeopleUser` is a temporary name until we completely reconcile whether or not we can make the switch to `User`.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -5,25 +5,25 @@ import * as DocumentTitle from 'react-document-title';
 import * as qs from 'simple-query-string';
 import { RouteComponentProps } from 'react-router';
 
-import SearchBar from './SearchBar';
 import LoadingSpinner from 'components/common/LoadingSpinner';
-import { ResourceType } from 'interfaces';
-
+import PopularTables from 'components/common/PopularTables';
```
ttannis: Rebase error, I will delete before merging.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -5,25 +5,25 @@ import * as DocumentTitle from 'react-document-title';
...
   DashboardSearchResults,
-  SearchAllOptions,
   SearchAllRequest,
   SearchResourceRequest,
   TableSearchResults,
-  UserSearchResults
+  UserSearchResults,
 } from 'ducks/search/types';
 
+import { ResourceType, SearchAllOptions, TableResource } from 'interfaces';
```
ttannis: Rebase error, I will delete the import of `TableResource`.

amundsen_application/static/js/ducks/feedback/types.ts
```diff
@@ -1,21 +1,21 @@
...
 export enum ResetFeedback {
-  ACTION = 'amundsen/feedback/RESET_FEEDBACK',
-}
+  REQUEST = 'amundsen/feedback/RESET_FEEDBACK_REQUEST',
+};
 export interface ResetFeedbackRequest {
-  type: ResetFeedback.ACTION;
-}
+  type: ResetFeedback.REQUEST;
+};
```
danwom: I'm not sure we need the `;` for defining interfaces. We should probably define an explicit convention for this.

amundsen_application/static/js/ducks/search/types.ts
```diff
@@ -1,74 +1,64 @@
...
-}
-
+};
 export interface SearchAllRequest {
   options: SearchAllOptions;
   term: string;
-  type: SearchAll.ACTION;
-}
-
+  type: SearchAll.REQUEST;
```
danwom: Any thoughts on adding `payload` to requests?

amundsen_application/static/js/interfaces/Enums.ts
```diff
@@ -0,0 +1,17 @@
...
+export enum ResourceType {
+  table = "table",
+  user = "user",
+  dashboard = "dashboard",
+};
+
+export enum UpdateMethod {
+  PUT = 'PUT',
+  DELETE = 'DELETE',
+};
```
danwom: Not sure about grouping by enums. I feel like they would be better situated inside their respective files (e.g. putting `ResourceType` in the `Resources` file.

amundsen_application/static/js/interfaces/User.ts
```diff
@@ -0,0 +1,25 @@
...
+  last_name: string;
+  manager_fullname: string;
+  profile_url: string;
+  role_name?: string;
+  slack_id: string;
+  team_name: string;
+  user_id: string;
+};
+
+export type LoggedInUser = PeopleUser & {};
```
danwom: What does the `& {}` do? I also noticed this on the `Bookmark` type. 

I'm also thinking about just using `Resource` instead of `Bookmark`, which is no different from `own` and `read` resources from a data perspective.

amundsen_application/static/js/components/ProfilePage/index.tsx
ttannis: I think it should be `User` as well, however I'm waiting to be sure about what's going on in our backend.

amundsen_application/static/js/components/Feedback/FeedbackForm/BugReportFeedbackForm/tests/index.spec.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from 'react';
 import { shallow } from 'enzyme';
 
 import AbstractFeedbackForm, { FeedbackFormProps } from 'components/Feedback/FeedbackForm';
-import { SendingState } from 'components/Feedback/types';
+import { SendingState } from 'interfaces';
```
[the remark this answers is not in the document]
ttannis: I find there's a benefit for components to only need to know that they are imported a shared interface from `interfaces` without needing to know what file it lives in. I think this is especially helpful as we figure out what we're doing and what is best -- we have free range to restructure and rearrange `interfaces` without needing to make component changes. 

The way I see us resolving our spaghetti import problem is to move away from importing from file by file and think about how we can better bundle those files and import from that bundle. I'm not convinced what I've set up is perfect, however it is set up in a way that let's us experiment and switch things up with less burden.

amundsen_application/static/js/ducks/search/types.ts
ttannis: Yeah called that out in https://github.com/lyft/amundsenfrontendlibrary/pull/196. Will move forward with that. (still in a separate PR though)

amundsen_application/static/js/ducks/feedback/types.ts
ttannis: I've been using it because this is what I believe is supposed to be the result of our current tslint configurations, but we have an open issue to figure out why things aren't getting applied.

amundsen_application/static/js/interfaces/Enums.ts
ttannis: `UpdateMethod` is a shared enum (used for updating tags, owners, and anything else that will have to use our "batch editing" hack) and I needed a place to put it for now.

I'm not opposed to putting the other enums back within the context in which they are used though. Sound good?

amundsen_application/static/js/interfaces/User.ts
ttannis: It's something that helps for iterating quickly during development. Given that we were at one point less sure about what params we'd be getting from the backend, this structure would let you freely modify `LoggedInUser` by doing:
```
export type LoggedInUser = PeopleUser & { 
   paramFromDummyDataUntilBackendIsReady: ...
}
```
Please feel free to clean up as you work towards completing the Profile Page feature, it's my understanding that the APIs are pretty much set in stone right now.

amundsen_application/static/js/ducks/search/types.ts
ttannis: Addressed in #197.

amundsen_application/static/js/ducks/feedback/types.ts
danwom: I don't totally agree that all the semi-colons are necessary, but we should table the discussion for another time.

amundsen_application/static/js/components/Feedback/FeedbackForm/BugReportFeedbackForm/tests/index.spec.tsx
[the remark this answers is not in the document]
danwom: Personally I would prefer to be more explicit with my imports, but I think our codebase is small enough to avoid any collisions in terms of ambiguous naming and collisions. I'm okay with trying this out and revisiting if it ever becomes an issue.

amundsen_application/static/js/ducks/feedback/types.ts
ttannis: Sounds good. In general my plan was always to rely on the most popular tslint configurations as opposed to personal preference for linting, because unless we make an objective choice we will have these kinds of discussions over and over as more folks get involved.

Until our lint problem is fixed all of my lint comments are pretty much nits and I don't plan on manually enforcing them on anybody. However once we have that fixed, my hope with that we can take a purely objective approach to code style where the discussions are based on whether or not a particular configuration is a hinderance or a benefit, and why.

amundsen_application/static/js/components/ProfilePage/index.tsx
ttannis: I'll leave this for you to clean up as you work on the Profile Page.

PR #196, opened 2019-06-14
Add owned/frequently uses/bookmarks to profile page

danwom:
### Summary of Changes
Added APIs for user owned and read resources. Added owned, read, and bookmarked resources to the profile page.

### Tests
Modified tests for the profile page, added tests to the python metadata APIs

### Documentation

### CheckList


danwom:
Tests are not complete but main functionality should be done.
@ttannis @feng-tao @ryanlieu

amundsen_application/static/js/ducks/user/types.ts
```diff
@@ -52,3 +57,42 @@ export interface GetUserResponse {
...
+export interface GetUserOwnRequest {
+  type: GetUserOwn.REQUEST;
+  payload: {
+    userId: string;
+  }
+}
+
+export interface GetUserOwnResponse {
+  type: GetUserOwn.SUCCESS | GetUserOwn.FAILURE;
+  payload?: Resource[];
```
ttannis: Since we decided we wanted `payload`s to be objects for consistency, this should be:
```
payload?: { 
  someName: Resource[] ;
};
```

amundsen_application/static/js/ducks/user/types.ts
```diff
@@ -52,3 +57,42 @@ export interface GetUserResponse {
...
+export interface GetUserReadRequest {
+  type: GetUserRead.REQUEST;
+  payload: {
+    userId: string;
+  }
+}
+
+export interface GetUserReadResponse {
+  type: GetUserRead.SUCCESS | GetUserRead.FAILURE;
+  payload?: Resource[];
```
ttannis: Since we decided we wanted `payload`s to be objects for consistency, this should be:
```
payload?: { 
  someName: Resource[] ;
};
```

amundsen_application/static/js/ducks/user/types.ts
```diff
@@ -52,3 +57,42 @@ export interface GetUserResponse {
...
+/* getUserOwn */
+export enum GetUserOwn {
+  REQUEST = 'amundsen/user/own/GET_REQUEST',
+  SUCCESS = 'amundsen/user/own/GET_SUCCESS',
+  FAILURE = 'amundsen/user/own/GET_FAILURE',
+}
+
+export interface GetUserOwnRequest {
+  type: GetUserOwn.REQUEST;
+  payload: {
```
ttannis: worthy mention: Seeing this reminded me that in my ducks refactor work I didn't consolidate the request actions that take params to use `payload` -- only did the response actions. Thanks for using it here. I'll clean that up for everything else.

amundsen_application/static/js/ducks/user/sagas.ts
```diff
@@ -29,3 +37,29 @@ export function* getUserWorker(action: GetUserRequest): SagaIterator {
 export function* getUserWatcher(): SagaIterator {
   yield takeEvery(GetUser.ACTION, getUserWorker);
 }
+
+export function* getUserOwnWorker(action: GetUserOwnRequest): SagaIterator {
+  try {
+    const userOwn = yield call(getUserOwn, action.payload.userId);
+    yield put({ type: GetUserOwn.SUCCESS, payload: userOwn.own });
```
ttannis: See comment in `ducks/user/types.ts` about the payload being an object that contains the data we wanted, and not just the data itself.

amundsen_application/static/js/ducks/user/sagas.ts
```diff
@@ -29,3 +37,29 @@ export function* getUserWorker(action: GetUserRequest): SagaIterator {
...
+}
+
+export function* getUserOwnWatcher(): SagaIterator {
+  yield takeEvery(GetUserOwn.REQUEST, getUserOwnWorker);
+}
+
+export function* getUserReadWorker(action: GetUserReadRequest): SagaIterator {
+  try {
+    const userRead = yield call(getUserRead, action.payload.userId);
+    yield put({ type: GetUserRead.SUCCESS, payload: userRead.read });
```
ttannis: See comment in `ducks/user/types.ts` about the payload being an object that contains the data we wanted, and not just the data itself.

amundsen_application/static/js/ducks/user/reducer.ts
```diff
@@ -53,9 +73,28 @@ export default function reducer(state: UserReducerState = initialState, action:
       return { ...state, loggedInUser: action.payload };
     case GetUser.ACTION:
     case GetUser.FAILURE:
-      return { ...state, profileUser: defaultUser };
+      state.profile.user= defaultUser;
```
ttannis: lint nit: space before `=`

amundsen_application/static/js/ducks/user/reducer.ts
```diff
@@ -53,9 +73,28 @@ export default function reducer(state: UserReducerState = initialState, action:
...
+      return { ...state };
```
ttannis: Much prefer this here:
```
{
  ...state,
  profile: {
    ...state.profile,
    user: defaultUser,
  }
}
```
 The state is supposed to be immutable and I wouldn't want this to temp others into recklessly starting to use `state.blah = foo` elsewhere. Also sure the above is more lines of code, but this is also one of the reason the redux state is supposed to either be kept as flat as possible or use nested reducers so that each reducer is pretty much just handling one level of the state.
https://github.com/reduxjs/redux/issues/432

amundsen_application/static/js/ducks/user/api/v0.ts
```diff
@@ -13,10 +13,28 @@ export function getLoggedInUser() {
...
+    })
+    .catch((error: AxiosError) => {
+      return {};
+    });
+}
+
+export function getUserOwn(userId: string) {
+  return axios.get(`/api/metadata/v0/user/own?user_id=${userId}`)
+    .then((response: AxiosResponse<UserOwnResponse>) => {
+      return response.data
```
ttannis: Do we have type safety here, I've assumed yes but recently I haven't been so sure with where TypeScript succeeds with its inferences versus not. In `ducks/user/sagas.ts` if we try to access `userOwn.notValid` do we get a type error?

amundsen_application/static/js/ducks/bookmark/reducer.ts
```diff
@@ -20,25 +21,25 @@ export function addBookmark(resourceKey: string, resourceType: string): AddBookm
     resourceType,
     type: AddBookmark.REQUEST,
   }
-};
+}
```
ttannis: Did the removal of semicolons come from your editor? If so you may want to look into that setting or if there are extensions for your IDE for the tslint configs we uses. Our tslint configs technically use `semicolon: [true, 'always']`. (however there's been an open task to revisit our lint configs to see why rules that I believed would be applied aren't getting applied).  

Just a nit though, as our linter will eventually enforce consistency.

amundsen_application/static/js/components/ProfilePage/index.tsx
```diff
@@ -37,41 +49,52 @@ export class ProfilePage extends React.Component<ProfilePageProps> {
...
   }
 
   getUserId = () => {
     return this.userId;
   };
 
-  // TODO: consider moving logic for empty content into Tab component
-  createEmptyTabMessage = (message: string) => {
+
+  getTabContent = (resource: Resource[], label: string, source: string) => {
```
ttannis: I think the `//TODO` should be left in. While we don't have to do it now, we will save ourselves the trouble of having to have this code chunk on every page that uses `Tabs`. In addition it will make it such that the empty message logic/markup will only have to be unit tested once in `Tabs` component.

amundsen_application/static/js/ducks/user/types.ts
danwom: I wasn't sure where we landed on this discussion. Will fix.

amundsen_application/static/js/ducks/bookmark/reducer.ts
danwom: My IDE highlights these semicolons as unnecessary, which means I just do a drive-by cleanup manually. I'll probably have to tinker with my settings to get them to read the tslint config.

amundsen_application/static/js/ducks/user/api/v0.ts
danwom: Doesn't seem like it-- probably because TS can't infer the type based on the `catch` part of these. Will investigate to see if there's a better pattern.

danwom:
Should be ready for re-review

amundsen_application/static/js/ducks/user/api/v0.ts
ttannis: Yeah I'll capture a ticket for this.

PR #197, opened 2019-06-14
Cleanup ducks folder: Address a few more inconsistencies.

ttannis:
### Summary of Changes

1. Consistent names for our `AxiosResponse` data types, especially ones that would not conflict with our action types that are named `SomethingResponse`. I've chosen the naming `SomethingAPI`.

2. There were some leftover cases in `ducks/tableMetadata` that were still passing the action to the api call. Instead we want to pass in explicit parameters to help decouple the logic in that file from our actions and their types. 
**_Worthy Mention_**: The `call()` method from `redux-saga` throws type errors if we try to pass variables that are potentially undefined, such as the optional parameters used for logging like `searchIndex` and `source`. A workaround is documented in the code.

3. Simplified ` getTableQueryParams` which used to take an object that was allowed to be either of type `TableMetadata` or type `GetTableDataRequest`. All this method needs is the table's key, so we can change the signature to `getTableQueryParams(tableKey: string)`. 

4. Use `payload` for request action params.

### Tests

Unit tests for ducks will be added in an upcoming PR.

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/js/ducks/search/reducer.ts
+++ b/amundsen_application/static/js/ducks/search/reducer.ts
@@ -24,16 +24,20 @@ export interface SearchReducerState {
 /* ACTIONS */
 export function searchAll(term: string, options: SearchAllOptions = {}): SearchAllRequest {
   return {
-    options,
-    term,
+    payload: {
+      options,
+      term,
+    },
     type: SearchAll.REQUEST,
   };
 };
 export function searchResource(resource: ResourceType, term: string, pageIndex: number): SearchResourceRequest {
   return {
-    pageIndex,
-    term,
-    resource,
+    payload: {
+      pageIndex,
+      term,
+      resource,
+    },
     type: SearchResource.REQUEST,
   };
 };
@@ -72,7 +76,7 @@ export default function reducer(state: SearchReducerState = initialState, action
       // updates search term to reflect action
       return {
         ...state,
-        search_term: (<SearchAllRequest>action).term,
+        search_term: (<SearchAllRequest>action).payload.term,
         isLoading: true,
       };
     case SearchResource.REQUEST:
--- a/amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
+++ b/amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
@@ -9,37 +9,37 @@ import { PreviewData, PreviewQueryParams, TableMetadata, User, Tag } from 'inter
 
 const API_PATH = '/api/metadata/v0';
 
-type MessageResponse = { msg: string };
+type MessageAPI = { msg: string };
 type TableData = TableMetadata & {
   owners: User[];
   tags: Tag[];
 };
-export type DescriptionResponse = { description: string; } & MessageResponse;
-export type LastIndexedResponse = { timestamp: string; } & MessageResponse;
-export type PreviewDataResponse = { previewData: PreviewData; } & MessageResponse;
-export type TableDataResponse = { tableData: TableData; } & MessageResponse;
+export type DescriptionAPI = { description: string; } & MessageAPI;
+export type LastIndexedAPI = { timestamp: string; } & MessageAPI;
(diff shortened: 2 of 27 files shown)


ttannis:
@danwom please take look only after any concerns on #195 are settled.

ttannis:
@danwom PTAL

amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
```diff
@@ -9,50 +9,48 @@ import { PreviewData, PreviewQueryParams, TableMetadata, User, Tag } from 'inter
 
 const API_PATH = '/api/metadata/v0';
 
-type MessageResponse = { msg: string };
+type MessageAPI = { msg: string };
```
danwom: Seems like a common interface used in most APIs. Maybe this should be shared in interfaces?

amundsen_application/static/js/ducks/tableMetadata/types.ts
```diff
@@ -15,9 +15,11 @@ export enum GetTableData {
 };
 export interface GetTableDataRequest {
   type: GetTableData.REQUEST;
-  key: string;
-  searchIndex?: string;
-  source?: string;
+  payload: {
+    key: string;
+    searchIndex?: string;
+    source?: string;
```
danwom: If you use default values for the `getTableData` you should be able to make these fields required.

amundsen_application/static/js/ducks/search/reducer.ts
```diff
@@ -72,7 +76,7 @@ export default function reducer(state: SearchReducerState = initialState, action
       // updates search term to reflect action
       return {
         ...state,
-        search_term: (<SearchAllRequest>action).term,
+        search_term: (<SearchAllRequest>action).payload.term,
```
danwom: If you add an explicit type to the action parameter on line 71, you _should_ be able to remove the cast here. 

In `ducks/user/reducer.ts` my IDE is able to correctly detect the action sub-type within each case block.

amundsen_application/static/js/ducks/tableMetadata/sagas.ts
```diff
@@ -23,8 +23,13 @@ import {
 
 export function* getTableDataWorker(action: GetTableDataRequest): SagaIterator {
   try {
-    /* TODO: Pass explicit params into api method and not action */
-    const { data, owners, statusCode, tags } = yield call(metadataGetTableData, action);
+    /* Errors will be throw if we use redux-saga 'call()' with undefined or null values.
+       As a workaround, set appropriate defaults for the optional values on the action. */
+    const { payload } = action;
+    const searchIndex = payload.searchIndex ? payload.searchIndex.toString() : '';
+    const source = payload.source || '';
```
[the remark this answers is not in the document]
ttannis: Will look into it.

amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
ttannis: Ah I see. But it's only used in `ducks`, and right now the current concept behind `interfaces` is that it would contain shared interfaces between `ducks` and `components`. That being said, if we identify any other things shared in `ducks`, I'm alright to start separating stuff out into a place for shared ducks types. I'll put a comment here.

amundsen_application/static/js/ducks/tableMetadata/types.ts
ttannis: Will look into it.

amundsen_application/static/js/ducks/search/reducer.ts
ttannis: That was removed in the previous cleanup PR because it's technically incorrect. All actions get passed through reducers, it's just that they don't have any effect if the case isn't covered. We were essentially adding a type definition based on what we knew we'd be processing, and not based on what Redux actually does. I have a ticket filed for looking into what's the best pattern for having type saftey on our actions.

PR #198, opened 2019-06-18
Made google analytics disabled by default

danwom:
### Summary of Changes
- Refactored `index.html` to use template fragments that are optionally included
- Disabled google analytics by default
- Added 'test' folder to templates for API tests to use

### Tests

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/static/templates/index.html
+++ b/amundsen_application/static/templates/index.html
@@ -1,15 +1,8 @@
 <html>
   <head>
-    <!-- Global site tag (gtag.js) - Google Analytics -->
-    <script async src="https://www.googletagmanager.com/gtag/js?id=<%= htmlWebpackPlugin.options.config.google.key%>"></script>
-    <script>
-      const google = <%= JSON.stringify(htmlWebpackPlugin.options.config.google) %>;
-      window.dataLayer = window.dataLayer || [];
-      function gtag() { dataLayer.push(arguments); }
-      gtag('js', new Date());
-      gtag('config', google.key);
-      gtag('create', google.key, { 'siteSpeedSampleRate': google.sampleRate });
-    </script>
+{% if <%= htmlWebpackPlugin.options.config.google.enabled%> %}
+{% include 'fragments/google-analytics-loader.html' %}
+{% endif %}
     <meta charset="utf-8">
     <link rel="shortcut icon" href="/static/images/favicon.png">
     <link href="/static/dist/main.css" type="text/css" rel="stylesheet"/>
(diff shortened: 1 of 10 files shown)


amundsen_application/static/templates/index.html
```diff
@@ -1,15 +1,9 @@
...
-      window.dataLayer = window.dataLayer || [];
-      function gtag() { dataLayer.push(arguments); }
-      gtag('js', new Date());
-      gtag('config', google.key);
-      gtag('create', google.key, { 'siteSpeedSampleRate': google.sampleRate });
-    </script>
+
+{% if <%= htmlWebpackPlugin.options.config.google.enabled%> %}
+{% include 'fragments/google-analytics-loader.html' %}
+{% endif %}
```
danwom: @ttannis any idea why this causes the following Python FE test to fail? I'm at a loss.

https://github.com/lyft/amundsenfrontendlibrary/blob/a5368433b44220982726608483c54fb7b7607e7c/tests/unit/api/search/test_v0.py#L164-L177

danwom:
CC: @bolkedebruin

amundsen_application/static/templates/index.html
danwom: It looks like something in this API causes Flask to parse and render `index.html`, which has obviously bad syntax because it hasn't been processed yet while running tests. But what triggers this?

amundsen_application/static/webpack.common.ts
```diff
@@ -1,10 +1,29 @@
...
+};
+const templatesList = walkSync('templates');
+const htmlWebpackPluginConfig = templatesList.map(file => {
+  return new HtmlWebpackPlugin({
+      filename: file,
+      template: file,
+      config: appConfig,
+      inject: false,
+    });
+});
```
danwom: Because HtmlWebpackPlugin can't process entire directories we have to process them all individually. `walkSync` will return all files in a directory and then create an `HtmlWebpackPlugin` config for each file.

ttannis: Can you clarify what purpose this serves in regards to the fix that needs to made?

amundsen_application/static/templates/index.html
ttannis: My only idea now is that some of the tests require creating a an application instance because we cannot test certain things outside of the application context. What happens if we tweak the following in some way:
https://github.com/lyft/amundsenfrontendlibrary/blob/a5368433b44220982726608483c54fb7b7607e7c/tests/unit/api/search/test_v0.py#L10

Perhaps removes the use of 'static/templates'? I see the metadata tests work fine with just the `TestConfig`

danwom: Removed the reference to `static/templates` and fixed the tests.

amundsen_application/static/webpack.common.ts
danwom: By default you would need to add one block per template. We currently have four.

```
new HtmlWebpackPlugin({	      ...htmlWebpackPluginConfig,
        template: path.join(__dirname, 'templates/index.html'),	
        filename: 'templates/index.html',	
        config: appConfig,	
        inject: false,	
      }),
```

amundsen_application/static/templates/index.html
danwom: Nvm it actually broke more tests. Locally it was defaulting to `static/dist/templates` which I had created from a previous build but those don't exist on the test framework.

PR #199, opened 2019-06-19
Browse on home page

ryanlieu:
### Summary of Changes

Adds the browse tags component on to the home page. This was done by moving the main logic for tags into a new BrowseTags component which then could be reused on both the BrowsePage and the HomePage.

### Tests

Moved the tests for BrowsePage to BrowseTags

### Documentation

### CheckList


ryanlieu:
@ttannis @danwom this should be good to go

amundsen_application/static/js/components/BrowsePage/index.tsx
```diff
@@ -1,116 +1,22 @@
...
-    } else {
-      innerContent = (
+    return (
+      <DocumentTitle title="Browse - Amundsen">
         <div className="container">
           <div className="row">
             <div className="col-xs-12">
-              <h3 id="browse-header">Browse Tags</h3>
-              <hr className="header-hr"/>
-              <div id="browse-body" className="browse-body">
```
ttannis: I believe that part that needed to be broken out was only the portion that showed the actual tags. i.e. what was contained in `browse-body`. We want this page to still look exactly the same. Why? We shouldn't make adhoc changes to existing pieces of the product, only what is required for the feature.

That being said, I think it is correct for `BrowseTags` to still be the owner of the logic that connects to the store and fetches the tags.

Thoughts?

ryanlieu: Yeah that would make sense to just make the component the list of tags. The reason I didn't go down that path was I wanted to keep our components on the home page consistent - both bookmarks and populartables have titles as part of their respective components. Either way we go we should make it consistent across the board.

ttannis: That would be a case for having an existing `BrowsePage`, the nested `BrowseTags` component connected to the redux store -- which by the way would be a common component now -- and if you want the title as part of the component on the landing page then some third component. 

The three components may seem overkill since two of them will be really simple -- but I think that's a symptom of us having a whole page (`BrowsePage`) that is just one component in the first place. If `BrowsePage` ever gets more content or a different layout, we will be glad we made the 3 components.

danwom: `PopularTables` fetches and connects the list of table resources. It also creates the title and places a `ResourceList` which takes care of displaying elements and pagination. For these components, there is no analogous full page component that displays only `PopularTables. Instead the `HomePage` is what houses and displays them.

If we're using `PopularTables/MyBookmarks` as an example, then...

- A `TagsList` component that would be responsible for displaying tags similarly to `ResourceList`. Instead of pagination, maybe it would have `[ + more]` `[ - less]` buttons (as per Philippe's mocks).
- A `BrowseTags` component would fetch and connect tags to the `TagsList`, perhaps with some logic about deciding which tags to show or hide. Maybe it has options to show top 20 tags, then click to show the rest.
- A `BrowsePage` would be like the `HomePage` that puts items together. Right now it only has tags, but this could be expanded to have other sections. Maybe there is an overall page title "Browse" and then a sub-section with "Tags", but that can all come once we have more things to show here. To be completely honest, the entire `BrowsePage` is sort of useless once the tags get added to the `HomePage`.

ryanlieu:
@ttannis created that third component, this should be good to look at again

amundsen_application/static/js/components/common/TagsList/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+export type TagsListProps = StateFromProps & DispatchFromProps;
+
+export class TagsList extends React.Component<TagsListProps, TagsListState> {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      curatedTags: [],
+      otherTags: [],
+      isLoading: this.props.isLoading,
```
danwom: I know I wrote this code originally, but we can remove `isLoading` from `this.state` and just use `this.props.isLoading` instead.

ryanlieu:
@ttannis this should be good to look over again, just made one TagsList component that is used both on HomePage and BrowsePage (which have their own unique titles)

amundsen_application/static/js/components/common/TagsList/index.tsx
ttannis: +1. Don't forget to remove it from the `TagsListState` interface as well

ryanlieu:
@ttannis @danwom should be good to look at again, addressed comments

amundsen_application/static/js/components/BrowsePage/tests/index.spec.tsx
```diff
@@ -26,6 +26,14 @@ describe('BrowsePage', () => {
       expect(wrapper.find(DocumentTitle).props().title).toEqual('Browse - Amundsen');
     });
     
+    it('renders correct header', () => {	
+      expect(wrapper.find('#browse-header').text()).toEqual('Browse Tags');	
+    });	
+
+    it('renders <hr> in if curatedTags.length > 0 & otherTags.length > 0 ', () => {
```
ttannis: Copy-paste? This title should just be `renders <hr>`

PR #201, opened 2019-06-20
Redux unit tests: feedback, log, popularTables, search

ttannis:
### Summary of Changes

This PR further increases the unit test coverage of our Redux code:
1. Coverage thresholds increased -- however only large pieces of new untested code should cause failures.
2. Actions creators now exist not for for request actions, but for the responses as well. The reason is that manually typing in the objects for response actions in multiple places, e.g. `{ type: GetAllTags.SUCCESS, payload: { tags }}` creates extra work any time we want to update or change the response action's type's name or the payload, and is more prone to human error. This change was applied to directories with existing coverage -- `allTags`, `announcements`, and `bookmarks` -- as well as the directories that gained coverage in this PR.
3. Updated `redux-saga-test-plan` for a bug fix, which required also updating `redux-saga` and `typescript` for compatibility. There are slight syntax changes in existing tests and sagas because of this, namely the usage of `takeEveryEffect` in tests is now just `takeEvery`, and the usage of `delay` in our sagas has been updated.

### Tests

1. There is some cleanup in existing tests due to the changes described above.
2. Added unit test coverage for `feedback`, `log`, `popularTables`, and `search` directories.

### Documentation

N/A

### CheckList


ttannis:
@danwom PTAL

amundsen_application/static/js/ducks/bookmark/reducer.ts
```diff
@@ -33,11 +41,25 @@ export function removeBookmark(resourceKey: string, resourceType: string): Remov
...
+  return { type: RemoveBookmark.SUCCESS, payload: { resourceKey, resourceType } };
+}
+
 export function getBookmarks(): GetBookmarksRequest {
   return {
     type: GetBookmarks.REQUEST,
   }
 }
+export function getBookmarksFailure(): GetBookmarksResponse {
+  return { type: GetBookmarks.FAILURE };
```
danwom: Why not return a `payload: { bookmarks: [] }` here?

amundsen_application/static/js/ducks/bookmark/reducer.ts
```diff
@@ -46,6 +68,12 @@ export function getBookmarksForUser(userId: string): GetBookmarksForUserRequest
     type: GetBookmarksForUser.REQUEST,
   }
 }
+export function getBookmarksForUserFailure(): GetBookmarksForUserResponse {
+  return { type: GetBookmarksForUser.FAILURE };
```
danwom: Same as above for this response.

amundsen_application/static/js/ducks/feedback/tests/index.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+          .next()
+          .call(feedbackSubmit, formData)
+          .next()
+          .put(submitFeedbackSuccess())
+          .next()
+          .delay(2000)
+          .next()
+          .put(resetFeedback())
+          .next()
+          .isDone();
```
danwom: This type of test seems very cryptic and tightly coupled with the code structure. I'm not sure how much value this adds overall.

amundsen_application/static/js/ducks/bookmark/types.ts
```diff
@@ -14,7 +14,7 @@ export interface AddBookmarkRequest {
 }
 export interface AddBookmarkResponse {
   type: AddBookmark.SUCCESS | AddBookmark.FAILURE;
-  payload: {
+  payload?: {
```
danwom: `?` could be removed if a default `[]` is returned for bookmark failures.

amundsen_application/static/js/ducks/allTags/tests/index.spec.ts
```diff
@@ -3,17 +3,37 @@ import * as matchers from 'redux-saga-test-plan/matchers';
...
+    it('getAllTags - returns the action to get all tags', () => {
+      expect(getAllTags()).toEqual({ type: GetAllTags.REQUEST });
+    });
+
+    it('getAllTagsFailure - returns the action to process failure', () => {
+      expect(getAllTagsFailure()).toEqual({
+        type: GetAllTags.FAILURE,
+        payload: {
+          tags: [],
+        },
```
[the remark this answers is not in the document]
ttannis: I think I can get behind that. What are your thoughts on the perspective that unit tests should only have one assertion? This will end up being slightly more code to separate the assertions.

amundsen_application/static/js/ducks/bookmark/reducer.ts
ttannis: We didn't use the value at all in the reducer so I deleted it. Seems like if we included it in the payload it should have been intentionally used in the reducer. Let me know what your intent was when writing this and I'll update accordingly.

amundsen_application/static/js/ducks/bookmark/reducer.ts
ttannis: We didn't use the value at all in the reducer so I deleted it. Seems like if we included it in the payload it should have been intentionally used in the reducer. Let me know what your intent was when writing this and I'll update accordingly.

amundsen_application/static/js/ducks/feedback/tests/index.spec.ts
ttannis: Sagas are iterators, so the way I understand that they should be tested is to verify that each iteration step proceeds as expected. This is also how redux-saga has documented to test sagas:
https://redux-saga.js.org/docs/advanced/Testing.html
> There are two main ways to test Sagas: testing the saga generator function step-by-step or running the full saga and asserting the side effects.

I've chosen the step by step testing as this seemed to be the approach that required the least amount of burden in terms of learning curve.

ttannis: It's also worth noting our sagas are currently extremely straight forward, they do a thing then return a thing. Our logic is purely linear and doesn't have any complicated branching logic yet or leverage some of the more complicated saga features.

amundsen_application/static/js/ducks/bookmark/types.ts
ttannis: Up for it depending on where we end up in regards to whether or not this needs to be included in the payload.

amundsen_application/static/js/ducks/allTags/tests/index.spec.ts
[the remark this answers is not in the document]
danwom: I don't think we need to limit one assertion per block. I think they can be grouped in logical blocks according to the scenario/case that's being tested. For example you could test that an HTTP response is correct and group the related assertions together. 
```
... setup
expect(response.status).toEqual(200);
expect(response.data.tags).toEqual(expectedTags);
```
If either test fails the error log will tell you which particular assertion failed, even if they share the same test block. In contrast doing a `expect(complexObj).equals(expectedComplexObject)` you will get a large diff printed out which may be difficult to decipher.

amundsen_application/static/js/ducks/bookmark/reducer.ts
danwom: The bookmarks _should_ behave like other reducers and set the value to default/empty array. If there is a discrepancy it's probably just an oversight or changed pattern that wasn't updated. 


That being said, I do think we need to revisit failure scenarios from a holistic perspective. Our current pattern is to fail quietly, but perhaps we should surface some error message to the user or do an automatic retry? I'm not sure what the best approach is, but we should standardize it. (Out of scope)

ttannis: There's an ancient ticket to revisit our error handling. We've been making these decisions on a case by case basis, but some guidelines would be nice. I will update the failure payload back to `{ bookmarks: [] }`.

amundsen_application/static/js/ducks/allTags/tests/index.spec.ts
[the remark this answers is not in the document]
ttannis: Also can get behind that reasoning, I will update the actions tests to assert on specific values that we expect to be on the action as opposed to the entire action itself.

ttannis:
@danwom PTAL

amundsen_application/static/js/ducks/search/tests/index.spec.ts
```diff
@@ -0,0 +1,211 @@
+import { expectSaga, testSaga } from 'redux-saga-test-plan';
+import * as matchers from 'redux-saga-test-plan/matchers';
+import { throwError } from 'redux-saga-test-plan/providers';
+
+import { ResourceType } from 'interfaces';
+
+import globalState from 'fixtures/globalState';
+
+import { searchResource as srchResource } from '../api/v0';
```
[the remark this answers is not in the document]
ttannis: i plan on renaming so that we don't have these conflicts, but doing that outside of PRs -- didn't want to cause any further confusion in the large diffs.

Issue #170, opened 2019-05-21
Frontend configurations to be set using API


verdan:
In order to support all the customizations of the frontend i.e., the logo path etc (mentioned [here](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/config/config-default.ts)) when installed using a pip package (already compiled), we need to pass those configs from backend and store them in React's store and use them instead of hardcoding them on frontend .js file. 

@ttannis @danwom FYI.

PR #206, opened 2019-06-21
Enable Amundsen People/User feature

danwom:
### Summary of Changes

- Added user search API with tests
- Added feature flag which enables/disables user search and links to the profile page
- Polish styles on ProfilePage

### Tests

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/api/search/v0.py
+++ b/amundsen_application/api/search/v0.py
@@ -10,6 +10,7 @@ from flask.blueprints import Blueprint
 
 from amundsen_application.log.action_log import action_logging
 from amundsen_application.api.utils.request_utils import get_query_param, request_search
+from amundsen_application.models.user import load_user, dump_user
 
 LOGGER = logging.getLogger(__name__)
 
@@ -25,6 +26,7 @@ valid_search_fields = {
 }
 
 SEARCH_ENDPOINT = '/search'
+SEARCH_USER_ENDPOINT = '/search_user'
 
 
 def _create_error_response(*, message: str, payload: Dict, status_code: int) -> Response:
@@ -69,7 +71,11 @@ def search_table() -> Response:
 
 @search_blueprint.route('/user', methods=['GET'])
 def search_user() -> Response:
-    return make_response(jsonify({'msg': 'Not implemented'}), HTTPStatus.NOT_IMPLEMENTED)
+    search_term = get_query_param(request.args, 'query', 'Endpoint takes a "query" parameter')
+    page_index = get_query_param(request.args, 'page_index', 'Endpoint takes a "page_index" parameter')
+    results_dict = _search_user(search_term=search_term, page_index=page_index)
+
+    return make_response(jsonify(results_dict), results_dict.get('status_code', HTTPStatus.INTERNAL_SERVER_ERROR))
 
 
 def _create_url_with_field(*, search_term: str, page_index: int) -> str:
@@ -117,25 +123,9 @@ def _search_user(*, search_term: str, page_index: int) -> Dict[str, Any]:
         """
 
     def _map_user_result(result: Dict) -> Dict:
-        return {
-            'type': 'user',
-            'active': result.get('active', None),
-            'birthday': result.get('birthday', None),
-            'department': result.get('department', None),
-            'email': result.get('email', None),
-            'first_name': result.get('first_name', None),
-            'github_username': result.get('github_username', None),
-            'id': result.get('id', None),
-            'last_name': result.get('last_name', None),
-            'manager_email': result.get('manager_email', None),
-            'name': result.get('name', None),
-            'offboarded': result.get('offboarded', None),
-            'office': result.get('office', None),
-            'role': result.get('role', None),
-            'start_date': result.get('start_date', None),
-            'team_name': result.get('team_name', None),
-            'title': result.get('title', None),
-        }
+        user_result = dump_user(load_user(result))
+        user_result['type'] = 'user'
+        return user_result
 
     users = {
         'page_index': int(page_index),
@@ -150,7 +140,31 @@ def _search_user(*, search_term: str, page_index: int) -> Dict[str, Any]:
         'users': users,
     }
 
-    return results_dict
+    try:
+        url = '{0}?query_term={1}&page_index={2}'.format(app.config['SEARCHSERVICE_BASE'] + SEARCH_USER_ENDPOINT,
+                                                         search_term,
+                                                         page_index)
+
+        response = request_search(url=url)
+        status_code = response.status_code
+
+        if status_code == HTTPStatus.OK:
+            results_dict['msg'] = 'Success'
+            results = response.json().get('results')
+            users['results'] = [_map_user_result(result) for result in results]
+            users['total_results'] = response.json().get('total_results')
+        else:
+            message = 'Encountered error: Search request failed'
+            results_dict['msg'] = message
+            logging.error(message)
+
+            results_dict['status_code'] = status_code
+        return results_dict
+    except Exception as e:
+        message = 'Encountered exception: ' + str(e)
+        results_dict['msg'] = message
+        logging.exception(message)
+        return results_dict
 
 
 @action_logging
--- a/amundsen_application/api/search/v0.py
+++ b/amundsen_application/api/search/v0.py
@@ -116,9 +116,8 @@ def _search_user(*, search_term: str, page_index: int) -> Dict[str, Any]:
     call the search service endpoint and return matching results
     :return: a json output containing search results array as 'results'
 
-    Schema Defined Here: https://github.com/lyft/
-    amundsensearchlibrary/blob/master/search_service/api/search.py
-
+    Schema Defined Here:
+    https://github.com/lyft/amundsensearchlibrary/blob/master/search_service/api/user.py
     TODO: Define an interface for envoy_client
     """
 
@@ -173,8 +172,8 @@ def _search_table(*, search_term: str, page_index: int) -> Dict[str, Any]:
     call the search service endpoint and return matching results
     :return: a json output containing search results array as 'results'
 
-    Schema Defined Here: https://github.com/lyft/
-    amundsensearchlibrary/blob/master/search_service/api/search.py
+    Schema Defined Here:
+    https://github.com/lyft/amundsensearchlibrary/blob/master/search_service/api/search.py
 
     TODO: Define an interface for envoy_client
     """
--- a/amundsen_application/api/search/v0.py
+++ b/amundsen_application/api/search/v0.py
@@ -113,14 +113,14 @@ def _create_url_with_field(*, search_term: str, page_index: int) -> str:
 @action_logging
 def _search_user(*, search_term: str, page_index: int) -> Dict[str, Any]:
     """
-        call the search service endpoint and return matching results
-        :return: a json output containing search results array as 'results'
+    call the search service endpoint and return matching results
+    :return: a json output containing search results array as 'results'
 
-        Schema Defined Here: https://github.com/lyft/
-        amundsensearchlibrary/blob/master/search_service/api/search.py
+    Schema Defined Here: https://github.com/lyft/
+    amundsensearchlibrary/blob/master/search_service/api/search.py
 
-        TODO: Define an interface for envoy_client
-        """
+    TODO: Define an interface for envoy_client
+    """
 
     def _map_user_result(result: Dict) -> Dict:
         user_result = dump_user(load_user(result))
--- a/amundsen_application/models/user.py
+++ b/amundsen_application/models/user.py
@@ -21,6 +21,7 @@ class User:
                  github_username: str = None,
                  is_active: bool = True,
                  last_name: str = None,
+                 manager_email: str = None,
                  manager_fullname: str = None,
                  profile_url: str = None,
                  role_name: str = None,
@@ -35,6 +36,7 @@ class User:
         self.github_username = github_username
         self.is_active = is_active
         self.last_name = last_name
+        self.manager_email = manager_email
         self.manager_fullname = manager_fullname
         self.profile_url = profile_url
         self.role_name = role_name
@@ -53,6 +55,7 @@ class UserSchema(Schema):
     github_username = fields.Str(allow_none=True)
     is_active = fields.Bool(allow_none=True)
     last_name = fields.Str(allow_none=True)
+    manager_email = fields.Str(allow_none=True)
     manager_fullname = fields.Str(allow_none=True)
     profile_url = fields.Str(allow_none=True)
     role_name = fields.Str(allow_none=True)
--- a/amundsen_application/models/user.py
+++ b/amundsen_application/models/user.py
@@ -78,6 +78,9 @@ class UserSchema(Schema):
             if app.config['GET_PROFILE_URL']:
                 data['profile_url'] = app.config['GET_PROFILE_URL'](data['user_id'])
 
+        # Fallback since search and metadata use a different key for 'full_name'
+        data['full_name'] = data.get('full_name', data.get('name'))
+
         if self. _str_no_value(data.get('display_name')):
             if self._str_no_value(data.get('full_name')):
                 data['display_name'] = data.get('email')
--- a/amundsen_application/static/css/_typography-default.scss
+++ b/amundsen_application/static/css/_typography-default.scss
@@ -15,6 +15,7 @@ h1,
 .h1 {
   font-size: 36px;
   font-weight: $font-weight-header-regular;
+  line-height: 34px;
 }
 
 h2,
--- a/amundsen_application/static/css/_typography-default.scss
+++ b/amundsen_application/static/css/_typography-default.scss
@@ -22,12 +22,14 @@ h2,
 .h2 {
   font-size: 26px;
   font-weight: $font-weight-header-bold;
+  line-height: 34px;
 }
 
 h3,
 .h3 {
   font-size: 20px;
   font-weight: $font-weight-header-bold;
+  line-height: 34px;
 }
 
 body {
@@ -47,7 +49,7 @@ body {
 .title-1 {
   font-size: 20px;
   font-weight: $font-weight-body-bold;
-  line-height: 24px;
+  line-height: 28px;
 }
 
 .title-2 {
@@ -64,7 +66,7 @@ body {
 .subtitle-1 {
   font-size: 20px;
   font-weight: $font-weight-body-semi-bold;
-  line-height: 20px;
+  line-height: 28px;
 }
 
 .subtitle-2 {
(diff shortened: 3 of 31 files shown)


amundsen_application/static/js/components/ProfilePage/index.tsx
```diff
@@ -46,25 +45,42 @@ interface DispatchFromProps {
...
-export type ProfilePageProps = StateFromProps & DispatchFromProps;
+interface RouteProps {
+  userId: string;
+}
+
+
+export type ProfilePageProps = StateFromProps & DispatchFromProps & RouteComponentProps<RouteProps>;
 
 export class ProfilePage extends React.Component<ProfilePageProps> {
   private userId: string;
```
danwom: Is there any reason why we use `private userId: string;` rather than using the `props` or `state`?

amundsen_application/static/js/components/ProfilePage/tests/index.spec.tsx
```diff
@@ -23,7 +23,7 @@ import {
 
 describe('ProfilePage', () => {
   const setup = (propOverrides?: Partial<ProfilePageProps>) => {
-    const props: ProfilePageProps = {
+    const props: Partial<ProfilePageProps> = {
```
danwom: Since I added `RouteComponentProps` to the profile page the props object requires a bunch of nested objects. Is it worth using a shortcut or just defining everything in `RouteComponentProps` or just using a shortcut?

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -155,8 +157,14 @@ export class SearchPage extends React.Component<SearchPageProps, SearchPageState
...
       },
-      // TODO PEOPLE - Add users tab
     ];
+    if (AppConfig.userFeature.enabled) {
+      tabConfig.push({
+        title: `Users (${ this.props.users.total_results })`,
+        key: ResourceType.user,
+        content: this.getTabContent(this.props.users, USER_RESOURCE_TITLE),
+      })
+    }
```
danwom: I thought about how to do feature flags, but this is what I came up with... littering the codebase with if-statements. Anyone have other ideas or suggestions?

CC: @jornh @verdan

amundsen_application/static/js/components/common/Breadcrumb/index.tsx
```diff
@@ -23,17 +23,15 @@ export const Breadcrumb: React.SFC<BreadcrumbProps> = (props) => {
     path = '/';
     text = 'Home';
     if (props.searchTerm) {
-      path = `/search?searchTerm=${props.searchTerm}&selectedTab=table&pageIndex=0`
-      text = 'Search Results'
+      path = `/search?searchTerm=${props.searchTerm}&selectedTab=table&pageIndex=0`;
```
danwom: Probably need to fix the hard-coded `selectedTab` and `pageIndex`. 

FYI: @ryanlieu

amundsen_application/models/user.py
```diff
@@ -78,6 +81,9 @@ def preprocess_data(self, data: Dict) -> Dict:
             if app.config['GET_PROFILE_URL']:
                 data['profile_url'] = app.config['GET_PROFILE_URL'](data['user_id'])
 
+        # Fallback since search and metadata use a different key for 'full_name'
+        data['full_name'] = data.get('full_name', data.get('name'))
```
danwom: Ideally, there should be a consistent data format among FE, Metadata, and Search.

amundsen_application/api/search/v0.py
```diff
@@ -107,35 +113,19 @@ def _create_url_with_field(*, search_term: str, page_index: int) -> str:
...
     """
-        call the search service endpoint and return matching results
-        :return: a json output containing search results array as 'results'
+    call the search service endpoint and return matching results
+    :return: a json output containing search results array as 'results'
 
-        Schema Defined Here: https://github.com/lyft/
-        amundsensearchlibrary/blob/master/search_service/api/search.py
+    Schema Defined Here: https://github.com/lyft/
+    amundsensearchlibrary/blob/master/search_service/api/search.py
```
feng-tao: the file should be https://github.com/lyft/amundsensearchlibrary/blob/master/search_service/api/user.py

amundsen_application/api/search/v0.py
```diff
@@ -107,35 +113,19 @@ def _create_url_with_field(*, search_term: str, page_index: int) -> str:
...
-            'manager_email': result.get('manager_email', None),
-            'name': result.get('name', None),
-            'offboarded': result.get('offboarded', None),
-            'office': result.get('office', None),
-            'role': result.get('role', None),
-            'start_date': result.get('start_date', None),
-            'team_name': result.get('team_name', None),
-            'title': result.get('title', None),
-        }
+        user_result = dump_user(load_user(result))
```
feng-tao: do you have any exception handling inside load_user method?

amundsen_application/static/js/components/ProfilePage/index.tsx
```diff
@@ -93,16 +109,16 @@ export class ProfilePage extends React.Component<ProfilePageProps> {
...
-      content: this.getTabContent(read, READ_SOURCE, READ_LABEL),
-      key: READ_TAB_KEY,
-      title: `${READ_TAB_TITLE} (${read.length})`,
-    });
     tabInfo.push({
       content: this.getTabContent(bookmarks, BOOKMARKED_SOURCE, BOOKMARKED_LABEL),
       key: BOOKMARKED_TAB_KEY,
       title: `${BOOKMARKED_TAB_TITLE} (${bookmarks.length})`,
     });
+    tabInfo.push({
```
feng-tao: nice! So bookmark will be first tab?

amundsen_application/static/js/components/ProfilePage/index.tsx
```diff
@@ -204,4 +224,4 @@ export const mapDispatchToProps = (dispatch) => {
   return bindActionCreators({ getUserById, getUserOwn, getUserRead, getBookmarksForUser }, dispatch);
 };
 
-export default connect<StateFromProps, DispatchFromProps>(mapStateToProps, mapDispatchToProps)(ProfilePage);
+export default connect<StateFromProps, DispatchFromProps>(mapStateToProps, mapDispatchToProps)(withRouter(ProfilePage));
```
feng-tao: what is `withRouter` doing here?

amundsen_application/static/js/components/ProfilePage/tests/index.spec.tsx
```diff
@@ -230,7 +227,7 @@ describe('ProfilePage', () => {
     it('renders Tabs w/ correct props', () => {
       expect(wrapper.find('#profile-tabs').find(Tabs).props()).toMatchObject({
         tabs: wrapper.instance().generateTabInfo(),
-        defaultTab: READ_TAB_KEY,
+        defaultTab: BOOKMARKED_TAB_KEY,
```
feng-tao: +1

amundsen_application/static/js/components/SearchPage/index.tsx
jornh: I don’t have anything against those if-statements - I think there easy enough to read. (I just googled for javascript feature flags and a React related lib came up. Not sure if it adds enough value to add that extra dependency though.

One thing I would like to ask though is if the default for the user feature should be that it’s turned on? Speaking *for* that would be that the default Docker frontend can be used to demo the User/People feature with just a hard coded user (Roald Amundsen comes to mind) without requiring a local rebuild first. The other alternative of course is to implement #170

amundsen_application/static/js/components/ProfilePage/index.tsx
danwom: Yep. It's that easy to rearrange.

amundsen_application/static/js/components/ProfilePage/index.tsx
danwom: The specific case where it's needed is when you link a profile page while the profile page is already loaded. E.g. routing from `/user/tfeng@lyft.com` -> `/user/dwon@lyft.com` wasn't loading. `withRouter` is a `react-router` utility allows us to read updates when the route changes.

amundsen_application/static/js/components/SearchPage/index.tsx
danwom: That's a very good point about whether the default should be on or off. I do see value in having it enabled in the default Docker version. My assumption, and I may be totally wrong, would be that most companies would be launching Amundsen with tables only. Integration with users would come later, if at all. What I am trying to avoid is for a company to integrate just tables and have a bunch of broken links to missing profile pages.

#170 is an interesting solution, although there would be non-trivial work to invest there. Some of the configurations are really build configurations (enabling google analytics will change the processed root `index.html` file) while others could be runtime configs (like this one).

amundsen_application/api/search/v0.py
danwom: There's a simple try-catch block inside `load_user` and `dump_user`. 

Seems like there should be a single function to decorate and cleanup a user dictionary. I'm not very familiar with this particular pattern in Python, but do you think this is the best pattern for this?

amundsen_application/static/js/components/ProfilePage/tests/index.spec.tsx
danwom: There's some code that does this in `homepage/tests/index.spec.ts` but I feel like that's a bit onerous just to get some URL match params.

I'll leave this as is and possibly revisit later.

amundsen_application/static/js/components/common/Breadcrumb/index.tsx
danwom: @ttannis this was the other quality-of-life issue that I wanted to fix separately.

amundsen_application/models/user.py
danwom: Done.

amundsen_application/static/css/_typography-default.scss
```diff
@@ -15,6 +15,7 @@ h1,
 .h1 {
   font-size: 36px;
   font-weight: $font-weight-header-regular;
+  line-height: 34px;
```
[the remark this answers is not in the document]
danwom: Vertical spacing on the H1s didn't look quite the same as mocks. Turns out some of the elements have explicitly defined line-heights. I can go in and add them for other type faces.

amundsen_application/static/js/components/ProfilePage/index.tsx
```diff
@@ -137,51 +152,56 @@ export class ProfilePage extends React.Component<ProfilePageProps> {
...
                       {
                         (!user.is_active) &&
-                        <Flag caseType="sentenceCase" labelStyle="label-danger" text="Alumni"/>
+                        <Flag caseType="sentenceCase" labelStyle="danger" text="Alumni"/>
                       }
                     </div>
                     {
                       user.role_name && user.team_name &&
-                      <label id="user-role">{ `${user.role_name} on ${user.team_name}` }</label>
+                      <div id="user-role" className="body-2">{ `${user.role_name} on ${user.team_name}` }</div>
```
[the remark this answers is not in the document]
danwom: There's probably a better tag for text, but I'm not sure it's `<label>`. Generally labels are used to describe form inputs (radio, input text, checkbox). Additionally these have some default styles that we don't want to inherit `font-weight: bold; display: inline-block; margin-bottom: 5px;`


By default I tend to use `<span>` or `<div>` since they are generally pretty neutral and don't come with too many default styles that need to be overwritten.

amundsen_application/static/js/components/ProfilePage/index.tsx
```diff
@@ -137,51 +152,56 @@ export class ProfilePage extends React.Component<ProfilePageProps> {
...
                     }
                     {
                       user.manager_fullname &&
-                      <label id="user-manager">{ `Manager: ${user.manager_fullname}` }</label>
+                      <div id="user-manager" className="body-2">{ `Manager: ${user.manager_fullname}` }</div>
                     }
                     <div className="profile-icons">
                       {
-                        user.is_active &&
+                        user.is_active && user.slack_id &&
```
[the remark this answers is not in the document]
danwom: Might be better to just comment this block out as well.

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -127,35 +128,31 @@ export class TableDetail extends React.Component<TableDetailProps & RouteCompone
...
     }
+
     return (
-      <OverlayTrigger key={fullName} trigger={['hover', 'focus']} placement="top" overlay={popoverHoverFocus}>
-        <div style={{display: 'inline-block', marginLeft: '-5px',
-          backgroundColor: 'white', borderRadius: '90%'}}>
-          <Avatar name={fullName} size={25} round={true} style={{zIndex, position: 'relative'}}/>
-        </div>
+      <OverlayTrigger key={user.display_name} trigger={['hover', 'focus']} placement="top" overlay={popoverHoverFocus}>
+        <Link
```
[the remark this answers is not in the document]
danwom: The `<Link>` will work with both cases. For external links the `target` attribute is set to `_blank`

amundsen_application/static/js/ducks/search/api/v0.ts
```diff
@@ -16,15 +17,30 @@ interface SearchAPI {
...
-  return axios.all([
-      axios.get(`${BASE_URL}/table?query=${term}&page_index=${options.tableIndex || 0}`),
-      // TODO PEOPLE - Add request for people here
-    ]).then(axios.spread((tableResponse: AxiosResponse<SearchAPI>) => {
-      return {
-        search_term: tableResponse.data.search_term,
-        tables: tableResponse.data.tables,
-      }
-  }));
+  const searchAllResponse = {
```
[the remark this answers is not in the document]
danwom: I also prefer the previous structure. Unfortunately it does not work when we need to optionally include or exclude API calls, especially with the `spread` operator whose parameter is a function that needs to know what parameters it should expect, and in what order.

The new structure each promise will independently modify the response object and `axiox.all` will wait for everything to be ready.

amundsen_application/static/js/interfaces/User.ts
```diff
@@ -1,6 +1,8 @@
 export interface User {
   display_name: string;
+  email: string;
```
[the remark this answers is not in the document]
danwom: Table readers are loaded and serialized by the `User` class. The code in there guarantees `display_name`, `email`, `profile_url`, and `user_id`.

```
    for reader_object in readers:
        reader_object['reader'] = _map_user_object_to_schema(reader_object['reader'])
```

```
def _map_user_object_to_schema(u: Dict) -> Dict:
    return dump_user(load_user(u))
```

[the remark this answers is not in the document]
danwom: I wonder how hard it would be to include the full user info in readers/owners

CC: @feng-tao @jinhyukchang

amundsen_application/static/js/config/config-default.ts
```diff
@@ -40,6 +40,9 @@ const configDefault: AppConfig = {
       return `https://DEFAULT_EXPLORE_URL?schema=${schema}&cluster=${cluster}&db=${database}&table=${table}`;
     }
   },
+  userFeature: {
```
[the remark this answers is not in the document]
danwom: Let's go with `indexUserResources`?

amundsen_application/static/js/ducks/search/api/v0.ts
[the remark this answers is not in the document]
ttannis: Occurs to me we might be doing this incorrectly -- I ask myself for what reason would we included the logic here as opposed to leveraging our sagas. It appears what we might actually need in this file is just as `searchResource` api call. The `searchAllWorker` saga could call `searchResource` multiple times to get the data, then pass those results into a utility method that massages those responses into what we need. `searchResource` could also just return the response data as well, and `searchResourceWorker` would also have a utility method would handle that logic on lines 50-55.

In short, I think in `ducks/search` we've been letting the api call take some of the middleware's responsibilities. Thoughts on the proposed change?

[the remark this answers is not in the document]
danwom: I think the previous version with a simple `all` and `spread` was simple enough, but reworking each optional API call to resolve individually changes things. We could move the aggregation step from APIs to the Saga and I think could simplify things again. 

Alternatively we could move the aggregation to our Flask server and combine our browser network calls, but this refactor would take some time.

amundsen_application/static/js/interfaces/User.ts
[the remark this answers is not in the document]
verdan: Please keep this issue in mind while making any decision here: 
https://github.com/lyft/amundsenmetadatalibrary/issues/23

At ING, we don't rely on email field, but on the Corporate Key, and expect it to be unique. For us email is not required, but good to have.

amundsen_application/static/js/ducks/search/api/v0.ts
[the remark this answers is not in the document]
ttannis: I think what you have in the most recent upload is perfectly fine. We introduced this redux-saga middleware in the first place to better handle these api calls. 

Though I think you've realized that redux-saga and Flask are both capable of taking care of the responsibility to massage the data before returning it to the application, this is further down the line but at some point we should make a clear decision of what lives in Flask (likely anything that needs the python configs or is tightly coupled with the metadata and search apis), and what lives in Redux (likely anything we do with the data that affects the presentation, filtering, sorting, combining, etc). :)

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -127,35 +128,31 @@ export class TableDetail extends React.Component<TableDetailProps & RouteCompone
...
+          to={ link }
+          target={ target }
+          style={{ display: 'inline-block', marginLeft: '-5px', backgroundColor: 'white', borderRadius: '90%'}}
```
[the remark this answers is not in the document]
danwom: Not sure what this was for... upon closer inspection these styles are redundant with `Avatar` component styles. Will delete.

amundsen_application/static/js/components/ProfilePage/tests/index.spec.tsx
```diff
@@ -230,18 +250,18 @@ describe('ProfilePage', () => {
...
 
     describe('if user.is_active', () => {
-      it('renders slack link with correct href', () => {
-        expect(wrapper.find('#slack-link').props().href).toEqual('www.slack.com');
-      });
-
-      it('renders slack link with correct text', () => {
-        expect(wrapper.find('#slack-link').find('span').text()).toEqual('Slack');
-      });
+      // it('renders slack link with correct href', () => {
```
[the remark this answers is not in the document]
danwom: We have a slack integration feature that we're planning to roll out. There were some problems with data ingestion so we commented out the html and tests. Is it better to delete or comment out?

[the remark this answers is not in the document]
verdan: A `ToDo/FixMe` with commented out perhaps.

amundsen_application/static/js/components/TableDetail/index.tsx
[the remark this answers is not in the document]
ttannis: We need these styles. It creates the "bubble overlap" look. 
image

Though yes it's more appropriate to be in a class

PR #208, opened 2019-06-21
Redux unit tests: tableMetadata, user

ttannis:
### Summary of Changes

This PR brings our coverage for our Redux layer logic to 80%+ and completes our current best effort initiative to increase the unit test coverage of Amundsen's application.

### Tests

1. Added tests to cover `ducks/user`
2. Added tests for best effort coverage of `ducks/tableMetadata`. 
3. Jest config thresholds updated. Current thresholds still undershoot the current coverage within a range of 5% to 8% for each measurement, as to not be too troublesome during development.

_Note: `ducks/tableMetadata/api ` still stands out as the one area that still needs coverage in the near future._

### Documentation

N/A.

### CheckList


amundsen_application/static/js/components/ProfilePage/index.tsx
```diff
@@ -9,7 +9,7 @@ import Flag from 'components/common/Flag';
 import Tabs from 'components/common/Tabs';
 
 import { GlobalState } from 'ducks/rootReducer';
-import { getUserById, getUserOwn, getUserRead } from 'ducks/user/reducer';
+import { getUser, getUserOwn, getUserRead } from 'ducks/user/reducer';
```
ttannis: Method name changed to prevent naming conflicts in tests.

amundsen_application/static/js/ducks/user/api/tests/index.spec.ts
```diff
@@ -0,0 +1,170 @@
+import axios, { AxiosResponse } from 'axios';
+
+import globalState from 'fixtures/globalState';
+
+import { LoggedInUser, PeopleUser, Resource } from 'interfaces';
+
+import {
+  loggedInUser, userById, userOwn, userRead,
+  LoggedInUserAPI, UserAPI, UserOwnAPI, UserReadAPI
```
danwom: Based on recent naming changes, I think these become less clear.
`UserAPI` is the response type, `userById` is the API function, and `getUser` is the connected dispatch function. On top of this we have the redux action types such as `GetUserRequest` and `GetUserResponse`. 


Maybe something like the following would work better
```
// Action Type
GetUser

// Actions
GetUserRequest
GetUserResponse

// Dispatch function
getUser

// API function
userById -> getUserApi

// API response
UserAPI -> GetUserApiResponse
```

danwom: We could also use the `import * as` syntax to for grouping and just call everything the same thing.
```
import * as UserAPI from '../v0';
import * as User from '../../reducer'

UserApi.getUser(...);
User.getUser
```

amundsen_application/static/js/ducks/tableMetadata/owners/tests/index.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+          .next()
+          .takeEvery(UpdateTableOwner.REQUEST, updateTableOwnerWorker);
+      });
+    });
+
+    describe('updateTableOwnerWorker', () => {
+      describe('executes flow for updating owners and returning up to date owner dict', () => {
+        let sagaTest;
+        beforeAll(() => {
+          sagaTest = (action) => {
```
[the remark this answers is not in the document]
ttannis: That's correct.

amundsen_application/static/js/ducks/tableMetadata/owners/tests/index.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+            return testSaga(updateTableOwnerWorker, action)
+                .next()
+                .select()
+                .next(globalState)
+                .all(apis.metadataUpdateTableOwner(updatePayload, globalState.tableMetadata.tableData.key))
+                .next()
+                .call(apis.metadataTableOwners, globalState.tableMetadata.tableData.key)
+                .next(expectedOwners)
+                .put(updateTableOwnerSuccess(expectedOwners));
```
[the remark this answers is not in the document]
ttannis: No problem. I went with the hybrid approach:
```
testSaga(getLastIndexedWorker, getLastIndexed())
          .next().call(metadataGetLastIndexed)
          .next(testEpoch).put(getLastIndexedSuccess(testEpoch))
          .next().isDone();
```

I will also back-propagate this syntax for existing tests in a syntax-only PR after this is merged.

amundsen_application/static/js/ducks/tableMetadata/reducer.ts
```diff
@@ -27,6 +27,24 @@ export function getTableData(key: string, searchIndex: string = '', source: stri
     type: GetTableData.REQUEST,
   };
 };
+export function getTableDataFailure(): GetTableDataResponse {
+  return {
+    type: GetTableData.FAILURE,
+    payload: { data: initialTableDataState, owners: {}, statusCode: 500, tags: [] }
+  }
+}
+export function getTableDataSuccess(data: TableMetadata, owners: { [id: string] : User }, statusCode: number, tags: Tag[]): GetTableDataResponse {
```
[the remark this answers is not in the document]
ttannis: 👍 Must have erased that during the merge (using merge instead of rebase now)

amundsen_application/static/js/ducks/user/api/tests/index.spec.ts
ttannis: **Re: Naming**
I don't intend to do any renaming in this PRs, only what is necessary and affected the unit tests. However I'm not opposed to taking a pass to have better and more consistent names -- it seems we're good on the actions and dispatch methods but things start feeling like "this could be better" when it comes to what's in those `api` directories, correct? I would very much like to clean those up.

For the naming patterns proposed, let me know how you feel about it when not looking a short simple method like `getUser`. 😸 I'm fine to make those changes in a renaming-only PR as long as we're sure we won't want to change it later solely because of long names, or get tempted to use abbreviations instead of being explicit.
Examples:
`updateTableDescriptionApi`
`UpdateTableDescriptionApiResponse`
My vote is for consistency and clarity so I'm fine with `SuperLongNameApiResponse` if that's what it has to be.

ttannis: **Re: Imports**
I'm a fan of the `import * as blah` to get rid of the crazy imports at the beginning of each test and having to add the new method each time. Only took until the monster that is `ducks/tableMetadata` to realize it. Some concerns:
1. The aliases will have to be more specific because naming conflicts, e.g. `import * as User from '../../reducer'`. Not sure if something like `SomethingReducer` is preferable, makes the imports nicer but we have to call `SomethingReducer.getSomethingFailure()`
2. Do you feel the same about the imports from '..types'?

danwom: I mentioned the names because we did change `getUserById` -> `userById` (API) and `getUserById` -> `getUser`(Reducer). We should 1) have some consistency about the `ById` part and 2) make these more clear as to what they are. I'm fine with revisiting in a separate PR to cleanup names.

For import formats `User` may be a bad example since there is a `User` interface already (or is that `PeopleUser` now?). In general I know you shouldn't use the * import pattern on large packages, but it should be okay `import *` from our own source files. It's not going to increase our bundle size. I don't have a concrete idea on when to use `import *` and when to use named imports, but we could employ it on a per-need basis in test files for now.

ttannis: For my previous comments what I'm getting at is purely whether or not you stand by the naming pattern proposals even if it results in long names. Seems like we're good and I'll use those patterns in the cleanup PR.

Let me know if you have any other concerns on this PR.

PR #211, opened 2019-06-25
Fixed empty search bar to go back to home

ryanlieu:
### Summary of Changes

Fixed issue where search wouldn't go back to home if search term was empty.

### Tests

Added test to make sure user is redirected home if search bar executes a search with an empty searchTerm

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
@@ -60,7 +60,7 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
   handleValueSubmit = (event: React.FormEvent<HTMLFormElement>) : void => {
     const searchTerm = this.state.searchTerm.trim();
     event.preventDefault();
-    if (this.isFormValid()) {
+    if (this.isFormValid(searchTerm)) {
       let pathName = '/';
       if (searchTerm !== '') {
         pathName = `/search?searchTerm=${searchTerm}`;
@@ -69,9 +69,7 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
     }
   };
 
-  isFormValid = () : boolean => {
-    const searchTerm = this.state.searchTerm;
-
+  isFormValid = (searchTerm: string) : boolean => {
     const hasAtMostOneCategory = searchTerm.split(':').length <= 2;
     if (!hasAtMostOneCategory) {
       this.setState({
--- a/amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
@@ -58,11 +58,12 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
   };
 
   handleValueSubmit = (event: React.FormEvent<HTMLFormElement>) : void => {
+    const searchTerm = this.state.searchTerm.trim();
     event.preventDefault();
     if (this.isFormValid()) {
       let pathName = '/';
-      if (this.state.searchTerm !== '') {
-        pathName = `/search?searchTerm=${this.state.searchTerm}&selectedTab=table&pageIndex=0`;
+      if (searchTerm !== '') {
+        pathName = `/search?searchTerm=${searchTerm}`;
       }
       this.props.history.push(pathName);    
     }
--- a/amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
@@ -60,7 +60,10 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
   handleValueSubmit = (event: React.FormEvent<HTMLFormElement>) : void => {
     event.preventDefault();
     if (this.isFormValid()) {
-      const pathName = `/search?searchTerm=${this.state.searchTerm}&selectedTab=table&pageIndex=0`;
+      let pathName = '/';
+      if (this.state.searchTerm !== '') {
+        pathName = `/search?searchTerm=${this.state.searchTerm}&selectedTab=table&pageIndex=0`;
+      }
       this.props.history.push(pathName);    
     }
   };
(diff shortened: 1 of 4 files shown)


ryanlieu:
@ttannis @danwom this fix should be good to check out

amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
```diff
@@ -60,7 +60,10 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
   handleValueSubmit = (event: React.FormEvent<HTMLFormElement>) : void => {
     event.preventDefault();
     if (this.isFormValid()) {
-      const pathName = `/search?searchTerm=${this.state.searchTerm}&selectedTab=table&pageIndex=0`;
+      let pathName = '/';
+      if (this.state.searchTerm !== '') {
```
danwom: Does this account for cases where a blank space is passed to the search bar? Not sure if there is a `trim()` somewhere along the way.

amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
```diff
@@ -60,7 +60,10 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
...
+        pathName = `/search?searchTerm=${this.state.searchTerm}&selectedTab=table&pageIndex=0`;
```
danwom: What happens if you don't hard-code the `selectedTab` and `pageIndex`? I think the `SearchPage` should be in charge of automating those default parameters.

amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
ryanlieu: no it does not, empty spaces returns with: "Your search - - did not match any tables result". Do we want the behavior to be that or should the user be redirected back to home if the bar is all white space?

amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
ryanlieu: the page will redirect automatically to the defaults (selectedTab=table and pageIndex=0), can change this

amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
```diff
@@ -58,9 +58,13 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
   };
 
   handleValueSubmit = (event: React.FormEvent<HTMLFormElement>) : void => {
+    const searchTerm = this.state.searchTerm.trim();
     event.preventDefault();
     if (this.isFormValid()) {
```
ttannis: As a cleanup here, since we are now grabbing the `searchTerm` in this method and trimming it, we let `isFormValid` take that string, and perhaps at this point a better name is `isSearchTermValid`.

amundsen_application/static/js/ducks/search/reducer.ts
```diff
@@ -104,6 +104,7 @@ export default function reducer(state: SearchReducerState = initialState, action
     case SearchResource.FAILURE:
       return {
         ...initialState,
+        search_term: (<SearchAllRequest>action).payload.term,
```
ttannis: This doesn't look quite right. Two different cases are now executing this logic, yet they don't have the same types. The action types are `SearchAllResponse` and `SearchResourceResponse`.  Both of these have a payload that is will have to contains all of these terms as well:
```
payload: {
    search_term: string;
    isLoading: boolean;
    dashboards: DashboardSearchResults;
    tables: TableSearchResults;
    users: UserSearchResults;
  };
```

I think your changes in the `SearchBar` look good, however now that we've discovered what the necessary changes are for the `category:fake` case I think it makes more sense to add it in as part of #201. You're free to invest the effort, but because I will still have to tweak it anyway it could save you some time.

ryanlieu: I've changed the call here, lmk if that looks correct - I feel like it makes sense to grab the search_term from SearchAllResponse's payload

amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
ryanlieu: :+1: cleaned this up

amundsen_application/static/js/ducks/search/reducer.ts
```diff
@@ -104,6 +104,7 @@ export default function reducer(state: SearchReducerState = initialState, action
     case SearchResource.FAILURE:
       return {
         ...initialState,
+        search_term: (<SearchAllResponse>action).payload.search_term,
```
ttannis: This does make sense, but let me clarify what I was trying to get at:
1. This current executes for `SearchAll.FAILURE` and `SearchResource.FAILURE`. However you haven't modified the `searhResearchWorker` saga to return a payload for `SearchResource.FAILURE`. It will be a better practice to separate the two cases regardless, as right now we've introduced a side effect for `SearchResource.FAILURE` as well. If `SearchResource` actually does fail, our app will current break because `payload` will be undefined. Let's fix that before we merge.
2. I referenced #201 because it makes some changes that has to modify this code regardless, so you had the option to not do throw-away work if you wanted. 
The issue that will have to be fixed/re-written regardless of what we merge in here has to do with this line in the sagas
```
yield put({ type: SearchAll.FAILURE, payload: {search_term: term} });
```
 Notice that our current code did not complain about the fact that you created a payload that only had `search_term` even though the `SearchAllResponse` type requires more on the payload according to the type we specified: https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/ducks/search/types.ts#L34. We're missing some type safety here and #201 cleans that up.

ryanlieu: ok, so in that case I should just roll back the changes related to the faulty query?

ttannis: yeah feel free

ryanlieu: :+1:

PR #215, opened 2019-07-02
Initial start to notifications API

ryanlieu:
### Summary of Changes

Adds a new /notification endpoint on the flask backend that currently doesn't do anything besides return a 200. Adds axios middleware so that endpoint can be called from frontend cleanly.

### Tests

None currently

### Documentation

N/A

### CheckList


ttannis:
@rlieu let's create that feature branch and update this PR to merge into that branch instead

ryanlieu:
@ttannis just made that new branch

amundsen_application/api/mail/v0.py
```diff
@@ -75,3 +75,14 @@ def _feedback(*,
               subject: str) -> None:
     """ Logs the content of the feedback form """
     pass  # pragma: no cover
+
+
+@mail_blueprint.route('/notification', methods=['POST'])
+def notification() -> Response:
```
ttannis: Something that I find very helpful for partial/dummy implementations is to leave "TODO" comments in the code. This helps me not nitpick about pieces of implementation that will be able to be addressed later. It will also serve a reminder to you to not forget to come back to certain things.

For example a "TODO" that could be added in the try block would be "When making a real api call, check the status code and return appropriate response". Example of what I mean:
https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/api/mail/v0.py#L50-L59

ryanlieu:
@ttannis assuming my changes pass the pipeline this should be good to look at again

amundsen_application/static/js/ducks/notification/api/tests/index.spec.ts
```diff
@@ -0,0 +1,34 @@
+import axios from 'axios';
+
+import { sendNotification } from '../v0';
```
ttannis: nit: We're using a new pattern here to make imports nicer for files that will have a lot of imports. This one doesn't but we should be consistent.

`import * as API from '../v0'`

and then call

`API.sendNotification(...)`

amundsen_application/static/js/ducks/notification/api/v0.ts
```diff
@@ -0,0 +1,11 @@
+import axios from 'axios';
+import { NotificationType, SendNotificationOptions } from 'interfaces/Notifications'
```
ttannis: Checkout `interfaces/index.ts`. You can export the `Notifications` module from there and just be able to use `... from 'interfaces'`.

amundsen_application/static/js/interfaces/Notifications.ts
```diff
@@ -0,0 +1,15 @@
+// TODO: Remove notification types that can be triggered in flask layer if necessary
+export enum NotificationType {
+  ADDED = 'amundsen/notification/OWNER_ADDED',
```
ttannis: I think we should still use simpler strings here. This is following the naming patterns of action enums (which is necessary to make sure we don't have any collisions, as every action needs a unique string), and I'd rather not start mixing in naming patterns associated with one area of code, into another area of code. These could easily be:
```
export enum NotificationType {
  ADDED = 'added',
  REMOVED = 'removed',
  EDITED = 'edited',
  REQUESTED = 'requested',
}
```

ryanlieu: gotcha, will do

ryanlieu:
@ttannis addressed comments

amundsen_application/static/js/interfaces/index.ts
```diff
@@ -6,3 +6,4 @@ export * from './Search';
 export * from './TableMetadata';
 export * from './Tags';
 export * from './User';
+export * from './Notifications'
```
ttannis: nit: alphabetical order

ryanlieu: :+1:

PR #217, opened 2019-07-08
Test Improvements: Date() issues, ReactRouterProps, & updated doc

ttannis:
### Summary of Changes
This PR addresses some low hanging items from our unit test backlog to help improve the unit test experience:
1. Our unit test scripts in `package.json` have been updated to run tests in the UTC timezone. This is so that unit tests for logic that uses JavaScript's `Date` object will be consistent for all users. Added a note about this in the doc as well.
2. Created `fixtures/mockRouter`, which has a `getMockRouterProps()` utility that can be reused for easier testing of components that use `ReactRouterProps`. It currently supports our use cases of overriding `match.params` and `history.location` for unit tests, and can be modified in the future if new use cases arise. 
3. Updated our `recommended_practices` doc to be organized by general recommendations, React specific recommendations, and added Redux recommendations. Also added some new recommendations based on recent learnings of better test patterns.

### Tests

1. Added back in commented out tests for `getDateLabel()` in `TableListItem`.  Tests should pass consistently for everyone.
2. Updated unit tests of components with ReactRouterProps to use the `getMockRouterProps()` utility.

### Documentation

Updated our `recommended_practices` doc.

### CheckList


ttannis:
@danwom PTAL

amundsen_application/static/js/components/HomePage/tests/index.spec.tsx
```diff
@@ -8,31 +8,23 @@ import SearchBar from 'components/SearchPage/SearchBar';
...
+
 describe('HomePage', () => {
   const setup = (propOverrides?: Partial<HomePageProps>) => {
+    const mockLocation = {
+      search: '/search?searchTerm=testName&selectedTab=table&pageIndex=1',
+      pathname: 'mockstr',
+      state: jest.fn(),
+      hash: 'mockstr',
+    };
+    const routerProps = getMockRouterProps<any>(null, mockLocation);
```
danwom: It looks like we only ever need `location.search`. What if instead of overwriting the entire `location` object we only did `location.search`?

amundsen_application/static/js/components/HomePage/tests/index.spec.tsx
```diff
@@ -8,31 +8,23 @@ import SearchBar from 'components/SearchPage/SearchBar';
...
-        pathname: 'mockstr',
-        state: jest.fn(),
-        hash: 'mockstr',
-      },
-      match: jest.fn() as any,
-      staticContext: jest.fn() as any,
+      history: routerProps.history,
+      location: routerProps.location,
+      match: routerProps.match,
+      staticContext: routerProps.staticContext,
```
danwom: Could we do `props =  { ...routerProps }`?

amundsen_application/static/js/fixtures/mockRouter.ts
```diff
@@ -0,0 +1,40 @@
...
+      replace: null,
+      go: null,
+      goBack: null,
+      goForward: null,
+      block: null,
+      createHref: null,
+      listen: null,
+    },
+    staticContext: {
+    }
```
danwom: I see ` staticContext: jest.fn() as any` elsewhere in code, but that might be wrong. Looking at the type info it is 

```
extends StaticContext {
  statusCode?: number;
}
```

amundsen_application/static/js/components/HomePage/tests/index.spec.tsx
danwom: If we really need to overwrite the whole `location` object, we can do 

```
props = {
  ...routerProps,
  location: {
    // custom overrides
  } 
}
```

amundsen_application/static/js/fixtures/mockRouter.ts
ttannis: Yeah the `jest.fn()` is incorrect. It's an object not a method, and I left it empty because I don't quite know what it does. 

Notice most things that we don't use I've set as sort of the bare minimum needed to pass the typechecker, because I'm concerned about adding in anything that would unknowingly have side effects. The way I currently see it, we add more meat to this only when that value is relevant for our tests.

amundsen_application/static/js/components/HomePage/tests/index.spec.tsx
ttannis: I can also make it such that `getMockRouterProps` takes `Partial<Location>`, and anything passed in will be used, else use the default.

ttannis:
@danwom made some changes

PR #222, opened 2019-07-12
Notifs post email functionality

ryanlieu:
### Summary of Changes

### Tests

### Documentation

### CheckList


amundsen_application/api/utils/notification_utils.py
```diff
@@ -0,0 +1,37 @@
...
+
+def get_notification_content(notification_type: str, data: Dict) -> Dict:
+    """
+    Forms a short version of a table Dict, with selected fields and an added 'key'
+    :param table: Dict of partial table object
+    :return: partial table Dict
+
+    TODO - Unify data format returned by search and metadata.
+    """
+    notification_type_dict = {
```
ttannis: Let's protect ourselves from a bug that would happen if we try to access `notification_type_dict` with an invalid `notification_type`. We want to make sure we don't try to access the `'html'` key on an undefined variable. There's a couple ways to do this, you could write a helper method, or you could use `.get` and return a default value. i.e. `notification_type_dict.get(notification_type, {}).get('html')`

The second argument of `.get()` on python dictionaries is what will get returned if the value doesn't exist, else it  returns `None`.

amundsen_application/static/templates/notification_added.html
```diff
@@ -0,0 +1,3 @@
+<!doctype html>
```
ttannis: nit: let's create a `templates/notifications` directory for these.

amundsen_application/static/js/components/Feedback/index.tsx
```diff
@@ -17,6 +17,8 @@ import {
 
 // TODO: Use css-modules instead of 'import'
 import './styles.scss';
+import { sendNotification } from 'ducks/notification/api/v0';
```
ttannis: I see this is for testing, let's make sure to add a `TODO: Delete after xyz` where `xyz` is whatever task allows you test this out some other way, I assume it'll be when we have the first iteration of thee "Requesting" component. Same for the comment chunk added in `toggle`.

amundsen_application/api/mail/v0.py
```diff
@@ -83,7 +89,33 @@ def notification() -> Response:
...
+
+        if not mail_client:
+            message = 'An instance of BaseMailClient client must be configured on MAIL_CLIENT'
+            logging.exception(message)
+            return make_response(jsonify({'msg': message}), HTTPStatus.NOT_IMPLEMENTED)
+
+        notification_content = get_notification_content(data['notificationType'], data['options'])
+
+        response = mail_client.send_email(
+            recipients=data['recipients'],
```
ttannis: We can add some safety here by using `get_query_param` from `request_utils`.  You can see examples of how we use it `metadata/v0`. We want to treat the React application as a black box (i.e. pretend we are not writing the logic or that we know the logic) to protect against the potential for something to go wrong there and the values don't get passed. 

If we use something like `get_query_param` (which could be renamed to something more applicable now but don't worry about that), if something goes wrong at least that method will raise an exception with a helpful error message about what parameter is missing.

ryanlieu:
@ttannis back up for review, should be good to look at again. Addressed comments, only large change outside of that was introducing a new exception type when the mail client is not implemented as well as removing typing for get_query_param (could return Dict, List, String, etc., not just String)

amundsen_application/api/mail/v0.py
```diff
@@ -7,24 +7,35 @@
...
 
 LOGGER = logging.getLogger(__name__)
 
 mail_blueprint = Blueprint('mail', __name__, url_prefix='/api/mail/v0')
 
 
-@mail_blueprint.route('/feedback', methods=['POST'])
-def feedback() -> Response:
-    """ An instance of BaseMailClient client must be configured on MAIL_CLIENT """
+class MailClientNotImplemented(Exception):
```
ttannis: We have logic for raising a "not implemented" error in other places. How about we put this in a new `amundsen_application/api/exceptions.py` can call it `NotImplementedException`. After it's merged down to master we can then use it other places as well.

amundsen_application/api/mail/v0.py
```diff
@@ -63,6 +74,10 @@ def feedback() -> Response:
             logging.error(message)
 
         return make_response(jsonify({'msg': message}), status_code)
+    except MailClientNotImplemented as e:
```
ttannis: nit: I'm not sure if giving this the same alias `e` will cause any obscure python bugs in the future, but we could always follow the existing convention in our other repos of using `e1`, `e2`, etc

amundsen_application/api/utils/request_utils.py
```diff
@@ -6,7 +6,7 @@
 REQUEST_SESSION_TIMEOUT_SEC = 3
 
 
-def get_query_param(args: Dict, param: str, error_msg: str = None) -> str:
+def get_query_param(args: Dict, param: str, error_msg: str = None):  # type: ignore
```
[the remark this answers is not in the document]
ryanlieu: MyPy was giving a few errors as get_query_params was returning non-str objects that were valid

amundsen_application/api/mail/v0.py
ryanlieu: :+1:

amundsen_application/api/utils/request_utils.py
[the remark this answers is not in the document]
ttannis: Ah just noticed your comment on this. I'm not a fan of loosening the typing on existing code. Perhaps the mistake was recommending that you use this method instead of creating a new one for your purposes given that your use case is different. 

Another option instead of using `get_query_param` could just be the simpler approach of using `.get()` and passing in some safe defaults. But how about I pose this instead -- what do you think will be best to do in the case where the React app has some issue and doesn't pass the right arguments? Let's agree on the behavior before I make any other recommendation.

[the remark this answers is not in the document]
ryanlieu: I feel like if any data's malformed or it doesn't exist then an exception should be thrown and a 500 should be returned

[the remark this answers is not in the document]
ryanlieu: @ttannis if that's the case, then directly accessing the dict would be fine as a KeyError would be thrown if the data is in fact malformed.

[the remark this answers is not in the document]
ttannis: I see let's go with that then. Apologies for the bit of extra churn.

ryanlieu:
@ttannis should be good to look over again

PR #230, opened 2019-07-18
Add typescript live-reload for local development 

kathawthorne:
### Summary of Changes

Add typescript live-reload for local development so that it is no longer necessary to rebuild and restart the python server when developing

### Tests

No tests were modified. This is specifically for development environment setup. 

### Documentation

I added documentation of how to use this feature. 

### CheckList

### DIFF
--- a/amundsen_application/static/webpack.dev.ts
+++ b/amundsen_application/static/webpack.dev.ts
@@ -1,4 +1,27 @@
+import * as path from 'path';
 import merge from 'webpack-merge';
-import commonConfig from './webpack.common'
+import commonConfig, {htmlWebpackPluginConfig} from './webpack.common';
+import MiniCssExtractPlugin from 'mini-css-extract-plugin';
 
-export default merge(commonConfig, {});
+export default merge(commonConfig, {
+  devServer: {
+      contentBase: path.join(__dirname, '../build/lib/amundsen_application/static/dist'),
+      publicPath: '/static/dist/',
+      watchContentBase: true,
+      port: 60480,
+      proxy: {
+          '!(/static/dist/**.*)': {
+              target: 'http://127.0.0.1:60477',
+              secure: false,
+          },
+      },
+  },
+  output: {
+    path: path.join(__dirname, '/dist'),
+    filename: '[name].js',
+  },
+  plugins: [
+      new MiniCssExtractPlugin(),
+      ...htmlWebpackPluginConfig("http://localhost:60480"), // should match port number above for devserver
+  ]
+});
--- a/amundsen_application/static/package-lock.json
+++ b/amundsen_application/static/package-lock.json
@@ -1276,6 +1276,23 @@
         "@types/react": "*"
       }
     },
+    "@types/events": {
+      "version": "3.0.0",
+      "resolved": "https://artifactory.global.square/artifactory/api/npm/square-npm/@types/events/-/events-3.0.0.tgz",
+      "integrity": "sha1-KGLz9Yqaf3w+eNefEw3U1xwlwqc=",
+      "dev": true
+    },
+    "@types/glob": {
+      "version": "7.1.1",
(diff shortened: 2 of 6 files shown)


kathawthorne:
Hm, I signed the CLA however it seems to still be failing.

ttannis:
@kathawthorne what version of node and npm are you running in development? Wondering why the `package-lock.json` has changed so much.

amundsen_application/static/webpack.dev.ts
```diff
@@ -1,4 +1,27 @@
+import * as path from 'path';
 import merge from 'webpack-merge';
-import commonConfig from './webpack.common'
+import commonConfig, {htmlWebpackPluginConfig} from './webpack.common';
+import MiniCssExtractPlugin from 'mini-css-extract-plugin';
 
-export default merge(commonConfig, {});
+export default merge(commonConfig, {
+  devServer: {
+      contentBase: path.join(__dirname, '../build/lib/amundsen_application/static/dist'),
```
ttannis: Where does the path `'../build/lib/amundsen_application/static/dist'` come from and what behavior does it contribute to? I'm looking to understand the difference between this and the `publicPath`.

amundsen_application/static/webpack.dev.ts
```diff
@@ -1,4 +1,27 @@
...
+      publicPath: '/static/dist/',
+      watchContentBase: true,
+      port: 60480,
+      proxy: {
```
ttannis: What behavior does this contribute to, and what exists at the specified target?

amundsen_application/static/webpack.dev.ts
```diff
@@ -1,4 +1,27 @@
...
+          '!(/static/dist/**.*)': {
+              target: 'http://127.0.0.1:60477',
+              secure: false,
+          },
+      },
+  },
+  output: {
```
ttannis: This configuration for `output` already exists in the common config, do we need it here?

amundsen_application/static/webpack.dev.ts
```diff
@@ -1,4 +1,27 @@
...
+    path: path.join(__dirname, '/dist'),
+    filename: '[name].js',
+  },
+  plugins: [
+      new MiniCssExtractPlugin(),
+      ...htmlWebpackPluginConfig("http://localhost:60480"), // should match port number above for devserver
```
ttannis: Does this mean folks will be accessing their local development version of the UI at `http://localhost:60480` instead of `http://localhost:5000`?

kathawthorne: Whoops! Internally we use different ports.

amundsen_application/static/webpack.dev.ts
kathawthorne: It's because I'm overwriting the plugins configuration from the common config so that it will attach a different port.

amundsen_application/static/webpack.dev.ts
kathawthorne: The proxy says that if you get a request for anything other than these urls, send it back to the original application port. This port is incorrect as is, however, and I need to switch it back to the port you use.

amundsen_application/static/webpack.dev.ts
kathawthorne: `contentBase` is where to serve content from, and this is the output content, including static files that seems to be in the repo. `publicPath` is what the prefix will be from the port. 
https://webpack.js.org/configuration/dev-server/#devserverpublicpath-
https://webpack.js.org/configuration/dev-server/#devservercontentbase

Reading through the documentation again, theoretically this could be just 
`contentBase: ../build/lib/amundsen_application` and omit the public path, but I'd need to play around with it.

kathawthorne:
Closing since there's a remaining unaddressed issue, in addition to the wrong version of npm.

PR #236, opened 2019-07-25
Added auto-select feature for search tabs

danwom:
### Summary of Changes

Added an auto-select tab feature.
- Added a 'submit' url param to force a search action from search bar
- Added 'default' resource type to indicate auto-select should activate

### Tests
Added tests for auto-select and modified existing tests related to search.

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
@@ -60,16 +60,13 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
   handleValueSubmit = (event: React.FormEvent<HTMLFormElement>) : void => {
     event.preventDefault();
     const searchTerm = this.state.searchTerm.trim();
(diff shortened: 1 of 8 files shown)


danwom:
The current approach is to update the URL and then sync the component state with the URL. Initially I put the auto-select logic into the saga, but this did not mesh well with the existing system and I encountered an exploding number of edge cases that were not handled.

I'll try again with removing the `default` tab type. I think I can simplify this to an acceptable level if I play around with it some more.

ttannis:
> The current approach is to update the URL and then sync the component state with the URL. Initially I put the auto-select logic into the saga, but this did not mesh well with the existing system and I encountered an exploding number of edge cases that were not handled.
> 
> I'll try again with removing the `default` tab type. I think I can simplify this to an acceptable level if I play around with it some more.

I see. Do you think it's worth going with putting the auto-select logic in the saga in the near future? Re: How we're considering changing up our approach for maintaining search state as part of filtering.

ttannis:
👍

PR #238, opened 2019-07-26
Request notifications component

ryanlieu:
### Summary of Changes

New request information component, pretty much has full functionality (sending a POST, grabbing list of owners, close button), need to work on validation and revealing components

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/Feedback/styles.scss
+++ b/amundsen_application/static/js/components/Feedback/styles.scss
@@ -5,7 +5,7 @@
   bottom: 75px;
   position: fixed;
   right: 25px;
-  z-index: 999;
+  z-index: 2;
 
   &.expanded {
     background-color: white;
--- a/amundsen_application/static/js/components/Feedback/styles.scss
+++ b/amundsen_application/static/js/components/Feedback/styles.scss
@@ -5,7 +5,7 @@
   bottom: 75px;
   position: fixed;
   right: 25px;
-  z-index: 2;
+  z-index: 999;
 
   &.expanded {
     background-color: white;
--- a/amundsen_application/static/js/components/TableDetail/RequestMetadataForm/constants.ts
+++ b/amundsen_application/static/js/components/TableDetail/RequestMetadataForm/constants.ts
@@ -1,7 +1,7 @@
 /* TODO: harcoded string that should be translatable/customizable */
 export const TITLE_TEXT = 'Amundsen Resource Request';
-export const FROM = 'From';
-export const TO = 'To';
+export const FROM_LABEL = 'From';
+export const TO_LABEL = 'To';
 export const REQUEST_TYPE = 'Request Type';
 export const TABLE_DESCRIPTION = 'Table Description';
 export const COLUMN_DESCRIPTIONS = 'Column Descriptions';
--- a/amundsen_application/static/js/components/TableDetail/RequestMetadataForm/constants.ts
+++ b/amundsen_application/static/js/components/TableDetail/RequestMetadataForm/constants.ts
@@ -1,0 +1,9 @@
+/* TODO: harcoded string that should be translatable/customizable */
+export const TITLE_TEXT = 'Amundsen Resource Request';
+export const FROM = 'From';
+export const TO = 'To';
+export const REQUEST_TYPE = 'Request Type';
+export const TABLE_DESCRIPTION = 'Table Description';
+export const COLUMN_DESCRIPTIONS = 'Column Descriptions';
+export const ADDITIONAL_DETAILS = 'Additional Details';
+export const SEND_BUTTON = 'Send Verification Request'
--- a/amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
+++ b/amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
@@ -14,14 +14,15 @@ import {
   ADDITIONAL_DETAILS,
   SEND_BUTTON,
 } from './constants'
-import { SubmitNotificationRequest } from 'ducks/notification/types';
-import { submitNotification } from 'ducks/notification/reducer';
+import { OpenRequestAction, SubmitNotificationRequest } from 'ducks/notification/types';
+import { openRequest, submitNotification } from 'ducks/notification/reducer';
 import { bindActionCreators } from 'redux';
 
 interface StateFromProps {
   userEmail: string;
   tableName: string;
   tableOwners: Array<string>;
+  requestIsOpen: boolean;
 }
 
 export interface DispatchFromProps {
@@ -31,28 +32,22 @@ export interface DispatchFromProps {
     notificationType: NotificationType,
     options?: SendNotificationOptions
   ) => SubmitNotificationRequest;
+  openRequest: () => OpenRequestAction;
 }
 
-type RequestMetadataProps = StateFromProps & DispatchFromProps;
+export type RequestMetadataProps = StateFromProps & DispatchFromProps;
 
-interface RequestMetadataState {
-  isOpen: boolean,
-}
+interface RequestMetadataState {}
 
 export class RequestMetadataForm extends React.Component<RequestMetadataProps, RequestMetadataState> {
   public static defaultProps: Partial<RequestMetadataProps> = {};
 
-
   constructor(props) {
     super(props);
-
-    this.state = {
-      isOpen: true,
-    };
   }
 
   toggle = () => {
-    this.setState({ isOpen: !this.state.isOpen });
+    this.props.openRequest();
   }
 
   submitForm = (event) => {
@@ -80,7 +75,7 @@ export class RequestMetadataForm extends React.Component<RequestMetadataProps, R
   };
 
   render() {
-    const expandedClass = this.state.isOpen ? 'expanded' : 'collapsed';
+    const expandedClass = this.props.requestIsOpen ? 'expanded' : 'collapsed';
     return (
       <div className={`request-component ${expandedClass}`}>
         <div className="form-section request-header">
@@ -118,15 +113,17 @@ export const mapStateToProps = (state: GlobalState) => {
   const userEmail = state.user.loggedInUser.email;
   const tableName = state.tableMetadata.tableData.schema + '.' + state.tableMetadata.tableData.table_name;
   const ownerObj = state.tableMetadata.tableOwners.owners;
+  const requestIsOpen = state.notification.requestIsOpen;
   return {
     userEmail,
     tableName,
+    requestIsOpen,
     tableOwners: Object.keys(ownerObj),
   };
 };
 
 export const mapDispatchToProps = (dispatch: any) => {
-  return bindActionCreators({ submitNotification } , dispatch);
+  return bindActionCreators({ submitNotification, openRequest } , dispatch);
 };
 
 export default connect<StateFromProps, DispatchFromProps>(mapStateToProps, mapDispatchToProps)(RequestMetadataForm);
--- a/amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
+++ b/amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
@@ -1,0 +1,28 @@
+import * as React from 'react';
+import './styles.scss';
+
+export class RequestMetadataForm extends React.Component{
+  render() {
+    return (
+      <div className="request-component expanded">
+        <form>
+          <input type="hidden" name="feedback-type" value="Bug Report"/>
+          <div className="form-group">
+            <input type="text" name="subject" className="form-control" required={ true } />
+          </div>
+          <div className="form-group">
+            <textarea name="bug-summary" className="form-control" required={ true }
+                      rows={3} maxLength={ 2000 }/>
+          </div>
+          <div className="form-group">
+            <textarea name="repro-steps" className="form-control" rows={5} required={ true }
+                      maxLength={ 2000 }/>
+          </div>
+          <button className="btn btn-default submit" type="submit">HEY</button>
+        </form>
+      </div>
+    );
+  }
+}
+
+export default RequestMetadataForm
--- a/amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
+++ b/amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
@@ -52,11 +52,12 @@ export class RequestMetadataForm extends React.Component<RequestMetadataProps, R
     event.preventDefault();
     const form = document.getElementById("RequestForm") as HTMLFormElement;
     const formData = new FormData(form);
-    const recipients = String(formData.get('recipients')).split(",");
-    const sender = String(formData.get('sender'));
+    const recipientString = formData.get('recipients') as string
+    const recipients = recipientString.split(",")
+    const sender = formData.get('sender') as string;
     const descriptionRequested = formData.get('table-description') === "on" ? true : false;
     const fieldsRequested = formData.get('column-description') === "on" ? true : false;
-    const comment = String(formData.get('details'));
+    const comment = formData.get('details') as string;
     sendNotification(
       recipients,
       sender,
--- a/amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
+++ b/amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
@@ -1,24 +1,80 @@
 import * as React from 'react';
 import './styles.scss';
 
-export class RequestMetadataForm extends React.Component{
+import { sendNotification } from 'ducks/notification/api/v0';
+import { NotificationType } from 'interfaces/Notifications';
+
+interface RequestMetadataProps {
+}
+
+interface RequestMetadataState {
+  isOpen: boolean,
+}
+
+export class RequestMetadataForm extends React.Component<RequestMetadataProps, RequestMetadataState> {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isOpen: true,
+    };
+  }
+
+  toggle = () => {
+    this.setState({ isOpen: !this.state.isOpen });
+  }
+
+  submitForm = (event) => {
+    event.preventDefault();
+    const form = document.getElementById("RequestForm") as HTMLFormElement;
+    const formData = new FormData(form);
+    const recipients = String(formData.get('recipients')).split(",");
+    const sender = String(formData.get('sender'));
+    const description_requested = formData.get('table-description') === "on" ? true : false;
+    const fields_requested = formData.get('column-description') === "on" ? true : false;
+    const comment = formData.get('details');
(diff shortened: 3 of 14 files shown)


amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    userEmail: '',
+    tableName: '',
+    tableOwners: [],
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isOpen: true,
```
danwom: Looks like you're initializing to open by default for testing.

Consider that an external action is going to trigger this component to open and close. You will likely need to redux to change TableMetadata state and pass that value into this component in `mapStateToProps`

amundsen_application/static/js/components/Feedback/styles.scss
```diff
@@ -5,7 +5,7 @@
   bottom: 75px;
   position: fixed;
   right: 25px;
-  z-index: 2;
+  z-index: 999;
```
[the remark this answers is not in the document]
ryanlieu: Ah, yeah there was an issue with the overlapping user avatars (had to modify z-indexes to get the avatars to render in the correct order). Should I just make this value smaller?

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+        </form>
+      </div>
+    );
+  }
+}
+
+export const mapStateToProps = (state: GlobalState) => {
+  const userEmail = state.user.loggedInUser.email;
+  const tableName = state.tableMetadata.tableData.schema + '.' + state.tableMetadata.tableData.table_name;
+  const ownerObj = state.tableMetadata.tableOwners.owners;
```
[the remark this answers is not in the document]
ryanlieu: :+1:

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+  submitForm = (event) => {
+    event.preventDefault();
+    const form = document.getElementById("RequestForm") as HTMLFormElement;
+    const formData = new FormData(form);
+    const recipients = String(formData.get('recipients')).split(",");
+    const sender = String(formData.get('sender'));
+    const descriptionRequested = formData.get('table-description') === "on" ? true : false;
+    const fieldsRequested = formData.get('column-description') === "on" ? true : false;
+    const comment = String(formData.get('details'));
+    sendNotification(
```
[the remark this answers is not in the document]
ryanlieu: Is there a reason why we should be using redux-saga for API calls if there's no change to the global state?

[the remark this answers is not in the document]
ryanlieu: Totally fine with switching it over, just wondering if there was a rationale here besides keeping API calls consistent

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
ryanlieu: ^ yeah that would make sense. I'll address this in the follow up PR where I create the component to open this component

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  render() {
+    const expandedClass = this.state.isOpen ? 'expanded' : 'collapsed';
+    return (
+      <div className={`request-component ${expandedClass}`}>
+        <div className="form-section request-header">
+          <h3 className="title">Amundsen Resource Request</h3>
+          <button type="button" className="btn btn-close" aria-label={"Close"} onClick={this.toggle}/>
+        </div>
+        <form onSubmit={ this.submitForm } id="RequestForm">
```
ttannis: You should separate this out into a separate component for testing when you get to that point, and even futher consider creating subcomponents for each form section. 

The reason is that unit testing the render methods of components with so much pure html makes updating unit tests very tedious. Why? Because when we test that each item is rendered we tend to write tests that equate to "Check if the nth child is this thing". With so much pure html combined into one render method, in the future if we decide we want to add some other component in, it will push the indices of all the components after it up by one, and then we have to update the indices for all of those tests as well.

I will let you decide if you want to do this now, or as a cleanup task when the majority of work is complete. If later, be sure to file a ticket so we don't forget.

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+      }
+    )
+  };
+
+  render() {
+    const expandedClass = this.state.isOpen ? 'expanded' : 'collapsed';
+    return (
+      <div className={`request-component ${expandedClass}`}>
+        <div className="form-section request-header">
+          <h3 className="title">Amundsen Resource Request</h3>
```
ttannis: We'll want to separate all hardcoded english text out into a `constants.ts` file, the import the constants here and use them. 

This will help us in the future when we implement some system for folks to configure custom text, either for their own UI needs or to support translating the application's text into another language.

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  toggle = () => {
+    this.setState({ isOpen: !this.state.isOpen });
+  }
+
+  submitForm = (event) => {
+    event.preventDefault();
+    const form = document.getElementById("RequestForm") as HTMLFormElement;
+    const formData = new FormData(form);
+    const recipients = String(formData.get('recipients')).split(",");
```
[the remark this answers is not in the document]
ryanlieu: Pretty sure I need to cast these as strings because formData.get() returns a FormDataEntryValue object, not a string

amundsen_application/static/js/components/Feedback/styles.scss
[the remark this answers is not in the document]
ryanlieu: yeah the bubbles would just go over the component on a small page, figured I'd put in a quick fix in this PR.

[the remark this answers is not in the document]
ttannis: I see, good eye catching the bug! Though as it stands it will make more sense to fix this either on master or as part of the update to put this in the NavBar.

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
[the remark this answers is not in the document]
ttannis: A `FormDataEntryValue` is documented as a "A string which represents a single value from a set of FormData key-value pairs" https://developer.mozilla.org/en-US/docs/Web/API/FormDataEntryValue...I wonder what's going on.

Do you get a type error when using `formData.get('recipients').split(",")` or an actual JS error?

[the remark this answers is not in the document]
ryanlieu: Im getting a type error, without the casting I get the error: Property 'split' does not exist on type 'FormDataEntryValue'.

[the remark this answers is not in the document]
ttannis: Did a bit more searching. That doc is misleading -__-, I looked at some code and it seems TypeScript defines `FormDataEntryValue` as being either `string | File`. So because it can be either, that would be why the code won't let you treat it as a string. 

However...we know it's going to be a string because we control the logic that generated the value downcast them as `string`.
```
...
const recipients = formData.get('recipients') as string;
const recipientsList = recipients.split(",");
const sender = formData.get('sender') as string;
....
```

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
ryanlieu: ^ will do this as a followup task

ryanlieu:
@ttannis @danwom  should be good to look at again, added redux-saga boilerplate

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+}
+
+type RequestMetadataProps = StateFromProps & DispatchFromProps;
+
+interface RequestMetadataState {
+  isOpen: boolean,
+}
+
+export class RequestMetadataForm extends React.Component<RequestMetadataProps, RequestMetadataState> {
+  public static defaultProps: Partial<RequestMetadataProps> = {};
```
ttannis: nit: You don't necessarily need this, as `defaultProps` only need to be assigned for optional props. No need to take action, just mentioning for understanding.

amundsen_application/static/js/ducks/notification/api/v0.ts
```diff
@@ -2,10 +2,15 @@ import axios from 'axios';
...
+  return axios({
+    data: {
+      recipients,
+      sender,
+      notificationType,
+      options,
+    },
+    method: 'post',
+    url: `/api/mail/v0/notification`,
+    timeout: 5000
```
ttannis: Let's remove `timeout: 5000` before merge. It's not needed and could cause problems if the mail client suffers some latency issue and happens to take more than 5 seconds to compete the request.

amundsen_application/static/js/ducks/notification/sagas.ts
```diff
@@ -0,0 +1,19 @@
+import { SagaIterator } from 'redux-saga';
+import { call, put, takeEvery } from 'redux-saga/effects';
+import { sendNotification } from './api/v0';
+
+import { submitNotificationFailure, submitNotificationSuccess } from './reducer';
+import { SubmitNotification, SubmitNotificationRequest } from './types';
+
+export function* submitNotificationWorker(action: SubmitNotificationRequest): SagaIterator {
+  try {
+    yield call(sendNotification, action.payload.recipients, action.payload.sender, action.payload.notificationType, action.payload.options);
```
ttannis: nit:
```
const { notificationType, options, recipients, sender } = action.payload;
yield call(sendNotification, recipients, sender, notificationType, options);
```

ryanlieu:
@ttannis sounds good, going to write unit tests for the component and get one last review before merging

ryanlieu:
@ttannis @danwom should be good to look at again - added a requestIsOpen state to the global redux store so that the component could be open/closed from another non-parent component. Also added tests for pretty much everything besides the render() function

amundsen_application/static/js/fixtures/globalState.ts
```diff
@@ -30,7 +30,10 @@ const globalState: GlobalState = {
     bookmarksForUser: [],
   },
   feedback: {
-      sendState: SendingState.IDLE,
+    sendState: SendingState.IDLE,
+  },
+  notification: {
```
danwom: nit: I'm not sure that `notification.requestIsOpen` is the right path for this value, since this isn't really a `notification` itself-- I would reserve for in-app notifications. This is a mechanism that triggers other users to receive notifications.

I suppose the `feedback` is the most similar pattern here, with the difference that `feedback` lives on every page and `requestData` only exists on `table_detail`. You _could_ make the argument that state should be organized on a per page basis (with the exception of global things like logged-in-user, feedback, and all-tags). In that case `requestIsOpen` would be added to the `table_detail` state (currently named `tableMetadata`).

Alternatively, if this `requestIsOpen` exists in multiple pages, perhaps it should be similar to `feedback` and have its own state branch at the root. Even then we could change this to something like 
```
modals: {
  requestTableMetadata: {
     isOpen: boolean;
     sendState: SendingState;
  },
  feedback: {
    isOpen: boolean;
    sendingState: SendingState;
  }
}
```

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/constants.ts
```diff
@@ -0,0 +1,9 @@
+/* TODO: harcoded string that should be translatable/customizable */
+export const TITLE_TEXT = 'Amundsen Resource Request';
+export const FROM = 'From';
+export const TO = 'To';
```
danwom: This variable name is probably too short. `FROM_LABEL` and `TO_LABEL` might be less ambiguous. That being said, it's also possible this is overkill.

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+            <label>{FROM}</label>
+            <input type="email" name="sender" className="form-control" required={true} value={this.props.userEmail}/>
+          </div>
+          <div className="form-section">
+            <label>{TO}</label>
+            <input type="email" name="recipients" className="form-control" required={true} multiple={true} defaultValue={this.props.tableOwners.join(",")}/>
+          </div>
+          <div className="form-section">
+            <label>{REQUEST_TYPE}</label>
+            <p><input type="checkbox" name="table-description"/> {TABLE_DESCRIPTION}</p>
```
danwom: Generally we want the checkbox description to be clickable. There are two main ways of doing this. Either way we want to use the `<label>` element and not the `<p>`

1) The `for` attribute should match the input box's `id` attribute
2) The `label` element wraps the `input` element and is automatically connected (recommended).

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/styles.scss
```diff
@@ -0,0 +1,48 @@
...
+  .title {
+    flex-grow: 1;
+  }
+
+  .request-header {
+    display: flex;
+  }
+
+  label {
+    display: block;
```
danwom: Why is this needed? There should be default styles associated with bootstrap forms and label elements.

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/styles.scss
```diff
@@ -0,0 +1,48 @@
...
+    border-radius: 0 0 6px 6px;
+    height: auto;
+    min-height: 450px;
+    padding: 32px;
+    width: 400px;
+    display: block;
+  }
+
+  &.collapsed {
+    display: none;
```
danwom: nit: You could put `display:none` on the base `.request-component` styles then you would only need `.expanded`

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/styles.scss
```diff
@@ -0,0 +1,48 @@
...
+  z-index: 2;
+
+  &.expanded {
+    background-color: white;
+    border-radius: 0 0 6px 6px;
+    height: auto;
+    min-height: 450px;
+    padding: 32px;
+    width: 400px;
+    display: block;
```
danwom: nit: Except `display:block` you could move all of these to the base `.request-component`.

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/styles.scss
```diff
@@ -0,0 +1,48 @@
...
+  .submit-button {
+    background-color: $brand-color-4;
+    color: white;
+    border-radius: 5px;
+    padding: 10px;
+    font-weight: $font-weight-body-bold;
+    float: right;
+  }
+
+  .form-section {
```
danwom: Why not use bootstrap `.form-group` and `.form-control` which were meant to be used together?

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+            <p><input type="checkbox" name="column-description"/> {COLUMN_DESCRIPTIONS}</p>
+          </div>
+          <div className="form-section">
+            <label>{ADDITIONAL_DETAILS}</label>
+            <textarea className="form-control" name="comment" rows={ 8 } maxLength={ 2000 } />
+          </div>
+          <button className="submit-button" type="submit">
```
danwom: We should use `className="btn btn-primary"` No need to reinvent button styles each time.

ryanlieu:
render if !isOpen then return null

ryanlieu:
@danwom addressed all comments besides the location of requestIsOpen in redux (waiting on @ttannis comments)

amundsen_application/static/js/fixtures/globalState.ts
ttannis: I think it's worth rearranging this state for sure. Though strong preference for doing this after the project is complete, as well as not adding `isOpen` for feedback in Redux because it's not needed based on how that component is used. I think consistency is great but only when the functionality is needed.

ryanlieu: :+1: if that's the case I'm just going to leave the redux state as is

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/tests/index.spec.tsx
```diff
@@ -0,0 +1,106 @@
...
+      wrapper.instance().submitNotification({ preventDefault: jest.fn() });
+      expect(submitNotificationSpy).toHaveBeenCalledWith(
+        mockFormData['recipients'].split(','),
+        mockFormData['sender'],
+        NotificationType.REQUESTED,
+        {
+          comment: mockFormData['comment'],
+          resource_name: props.tableName,
+          resource_url: window.location.href,
+          description_requested: mockFormData['table-description'] === "on" ? true : false,
```
ttannis: nit: your tests really shouldn't include boolean logic like this, a unit test is expected to be set up in such a way that given a particular input, we get a particular output. So we should just be expecting `true` or `false` here based on what condition we have set up.

amundsen_application/static/js/components/TableDetail/RequestMetadataForm/tests/index.spec.tsx
```diff
@@ -0,0 +1,106 @@
...
+    let result;
+    beforeAll(() => {
+      result = mapStateToProps(globalState);
+    });
+
+    it('sets userEmail on the props', () => {
+      expect(result.userEmail).toEqual(globalState.user.loggedInUser.email);
+    });
+    it('sets tableName on the props', () => {
+      expect(result.tableName).toEqual(globalState.tableMetadata.tableData.schema + '.' + globalState.tableMetadata.tableData.table_name);
```
ttannis: nit: ideally we'd have a method that generates this value so that the test isn't dependent on executing extra logic. Not need to address this now but in the future it will be best to have a test like:
```
expect(result.tableName).toEqual(methodThatOwnsThisLogic(...));
```

That way, if we ever change how the tableName is generated, we won't have to change the test itself.

PR #239, opened 2019-07-26
Add better logging for profile page views

danwom:
### Summary of Changes
- Added optional logging fields `index` and `source` to the GetUser API.
- Removed mandatory inclusion of `index` and `source` to GetTableData API

### Tests
- Added tests for Profile Page logging params
- Added logging params to various tests.

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/api/metadata/v0.py
+++ b/amundsen_application/api/metadata/v0.py
@@ -1,7 +1,7 @@
 import logging
 
 from http import HTTPStatus
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 from flask import Response, jsonify, make_response, request
 from flask import current_app as app
@@ -389,7 +389,7 @@ def update_table_tags() -> Response:
 def get_user() -> Response:
 
     @action_logging
-    def _log_get_user(*, user_id: str, index: int, source: str) -> None:
+    def _log_get_user(*, user_id: str, index: Optional[int], source: Optional[str]) -> None:
         pass  # pragma: no cover
 
     try:
--- a/amundsen_application/api/metadata/v0.py
+++ b/amundsen_application/api/metadata/v0.py
@@ -80,8 +80,8 @@ def get_table_metadata() -> Response:
     """
     try:
         table_key = get_query_param(request.args, 'key')
-        list_item_index = get_query_param(request.args, 'index')
-        list_item_source = get_query_param(request.args, 'source')
+        list_item_index = request.args.get('index', None)
+        list_item_source = request.args.get('source', None)
 
         results_dict = _get_table_metadata(table_key=table_key, index=list_item_index, source=list_item_source)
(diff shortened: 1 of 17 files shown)


amundsen_application/api/metadata/v0.py
```diff
@@ -389,14 +389,16 @@ def _log_update_table_tags(*, table_key: str, method: str, tag: str) -> None:
 def get_user() -> Response:
 
     @action_logging
-    def _log_get_user(*, user_id: str) -> None:
+    def _log_get_user(*, user_id: str, index: int, source: str) -> None:
```
danwom: @jinhyukchang Are there any consequences of logging a value of `None`? I've changed the default value for `index` and `source` to `None` on line 83-84.

amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
```diff
@@ -48,9 +48,16 @@ export function updateTableTags(tagArray, tableKey: string) {
...
-  const tableParams = getTableQueryParams(tableKey);
-  return axios.get(`${API_PATH}/table?${tableParams}&index=${searchIndex}&source=${source}`)
+export function getTableData(tableKey: string, searchIndex?: string, source?: string ) {
+  let url = `${API_PATH}/table?${getTableQueryParams(tableKey)}`;
+  if (searchIndex !== undefined) {
+    url = `${url}&index=${searchIndex}`;
+  }
+  if (source !== undefined) {
+    url = `${url}&source=${source}`;
+  }
```
danwom: Will update to use the `simple-query-string.stringify` function.

ttannis:
@jinhyukchang can you take a look at the question regarding changes in `amundsen_application/api/metadata/v0.py`?

amundsen_application/api/metadata/v0.py
jinhyukchang: I don't think so as it will serialized as a python `dict`. What was the default value before?

Nit: btw, I think the type of index `int` should be `Optional[int]` as it's now nullable.

amundsen_application/api/metadata/v0.py
```diff
@@ -80,8 +80,8 @@ def get_table_metadata() -> Response:
     """
     try:
         table_key = get_query_param(request.args, 'key')
-        list_item_index = get_query_param(request.args, 'index')
-        list_item_source = get_query_param(request.args, 'source')
+        list_item_index = request.args.get('index', None)
```
jinhyukchang: What would be the case when either `index` or `source` null?

danwom: In React FE `index` and `source` are optional parameters but in Flask they are required parameters. To get around this we added `index` and `source` with no value to the URL. E.g. `v0/table?key=hive...fact_driver&index=&source=` This ends up being read as empty string (vs None in new code).

I changed these to optional so React can make requests like `v0/table?key=hive...fact_driver` and not fail.

PR #242, opened 2019-07-31
Email Notification Actual Copy

ryanlieu:
### Summary of Changes

Adds actual content needed in notification emails. New calls to metadata service were needed to retrieve table and user metadata. 

### Tests

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/api/metadata/v0.py
+++ b/amundsen_application/api/metadata/v0.py
@@ -168,13 +168,18 @@ def update_table_owner() -> Response:
         else:
             raise Exception('method not handled')
 
+        results_dict = _get_table_metadata(table_key=table_key, index=0, source="")
+        table_data = results_dict['tableData']
+
         send_notification(
             notification_type=notification_type,
-            options={},
+            options={
+                'resource_name': '{}.{}'.format(table_data['schema'], table_data['table_name']),
+                'resource_url': 'https://amundsen.lyft.net/table_detail/gold/{}/{}/{}'.format(table_data['database'], table_data['schema'], table_data['table_name']),
+            },
             recipients=[owner],
             sender=user.email
         )
-
         payload = jsonify(_update_table_owner(table_key=table_key, method=request.method, owner=owner))
         return make_response(payload, HTTPStatus.OK)
     except Exception as e:
@@ -417,7 +422,6 @@ def get_user() -> Response:
 
         response = request_metadata(url=url)
         status_code = response.status_code
-
         if status_code == HTTPStatus.OK:
             message = 'Success'
         else:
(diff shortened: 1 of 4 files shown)


amundsen_application/api/metadata/v0.py
```diff
@@ -168,13 +169,18 @@ def update_table_owner() -> Response:
         else:
             raise Exception('method not handled')
 
+        results_dict = _get_table_metadata(table_key=table_key, index=0, source="")
```
ttannis: We should not be making a request here. I see this information is being used just to get the table data, for the purpose of getting the schema, tablename, and database. I also see a hardcoded Lyft specific url. You should be able to do the following:
1. It seems we should put the `table_name` and `schema` into the request json so that we can grab it from the there.
2. The `table_key` should already what we need to create the `resource_url`
3. Create a config variable for the url with a default value of `localhost:5000`, and in our internal logic will be assigned accordingly. This logic should use that config variable instead of hardcoding anything.

amundsen_application/api/utils/notification_utils.py
```diff
@@ -21,14 +23,15 @@ def send_notification(*, notification_type: str, options: Dict, recipients: List
...
             notification_type=notification_type,
-            options=options
+            options=options,
+            recipients=recipients,
+            sender=sender
         )
 
-        if sender in recipients:
-            recipients.remove(sender)
+        # if sender in recipients:
```
ttannis: You can still leave this in, for testing purposes if you want to make sure you get the email so can do that in our custom client. I prefer that option so that we can still check that the email list is what it should be and that all logic is working correctly. And then just append yourself to the list for testing purposes, or temporarily replace the recipients list with yourself.

amundsen_application/api/utils/notification_utils.py
```diff
@@ -80,34 +84,42 @@ def get_mail_client():  # type: ignore
...
     Returns a subject and a rendered html email template based off
     the input notification_type and data provided
     :param notification_type: type of notification
     :param options: data necessary to render email template content
     :return: subject and html Dict
     """
+    user_id = recipients[0]
+    url = '{0}{1}/{2}'.format(app.config['METADATASERVICE_BASE'], USER_ENDPOINT, user_id)
+
+    recipient_name = request_metadata(url=url).json()['first_name']
```
ttannis: I'm not a fan of having these extra requests here just to have personalized emails, it changes our call patterns in a way that we did not document.

I'd prefer to not shoot for personalized emails in this initial implementation anymore (or perhaps revisit next week if we have). The reason is that now we will want to consider calling all of these requests form the react application to begin with and passing in all parameters that are needed, or holding off on personalized emails until we have the support to do it the right way. (e.g. the owner information should ideally have first name and last name in the future not just email)

PR #243, opened 2019-07-31
Prevent Self-Notifications

ryanlieu:
### Summary of Changes

Adds a check in so that owners adding/removing themselves won't be notified that they have added themselves

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/api/metadata/v0.py
+++ b/amundsen_application/api/metadata/v0.py
@@ -168,13 +168,12 @@ def update_table_owner() -> Response:
         else:
             raise Exception('method not handled')
 
-        if owner != user.email:
-            send_notification(
-                notification_type=notification_type,
-                options={},
-                recipients=[owner],
-                sender=user.email
-            )
+        send_notification(
+            notification_type=notification_type,
+            options={},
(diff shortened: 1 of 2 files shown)


amundsen_application/api/metadata/v0.py
```diff
@@ -167,12 +167,14 @@ def update_table_owner() -> Response:
...
         else:
             raise Exception('method not handled')
-        send_notification(
-            notification_type=notification_type,
-            options={},
-            recipients=[owner],
-            sender=user.email
-        )
+
+        if owner != user.email:
```
ttannis: What do you think of having this logic in `send_notification` instead? The way I'm looking at this, the real use case here is that we want to remove the `sender` from `recipients` if they exist in that list. 

Note that if we do that we should then also check if `recipients` is empty and not send the email if that is the case. And if so perhaps use `logging.info()` to log a message about this happening. The `logging.info()` isn't necessary, but I think it will be helpful in future if we ever have to debug this feature to log information about whether or not a particular edge case was hit.

Thoughts?

PR #245, opened 2019-07-31
[WIP] Added auto-select tab for search

danwom:
### Summary of Changes

### Tests

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/SearchPage/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/index.tsx
@@ -93,6 +93,15 @@ export class SearchPage extends React.Component<SearchPageProps> {
     const nextUrlParams = this.getUrlParams(this.props.location.search);
     const prevUrlParams = this.getUrlParams(prevProps.location.search);
 
+    if (!this.props.isLoading && prevProps.isLoading) {
+      console.log('loading finished here!')
+      console.log(nextUrlParams);
+      const globalStateParams = this.getGlobalStateParams();
+      console.log(globalStateParams);
+      this.updatePageUrl(globalStateParams.term, globalStateParams.tab, globalStateParams.index, true);
+
+    }
+
     // If urlParams and globalState are synced, no need to update
     if (this.isUrlStateSynced(nextUrlParams)) return;
 
@@ -117,11 +126,12 @@ export class SearchPage extends React.Component<SearchPageProps> {
   }
 
   shouldUpdateSearchTerm(nextUrlParams, prevUrlParams): boolean {
-    return nextUrlParams.term !== prevUrlParams.term;
+    return nextUrlParams.term !== prevUrlParams.term ||
+      (nextUrlParams.tab == null && nextUrlParams.index == null);
   }
 
   shouldUpdateTab(nextUrlParams, prevUrlParams): boolean {
-    return nextUrlParams.tab !== prevUrlParams.tab;
+    return nextUrlParams.tab !== prevUrlParams.tab && nextUrlParams.tab !== null;
   }
 
   shouldUpdatePageIndex(nextUrlParams, prevUrlParams): boolean {
@@ -135,7 +145,7 @@ export class SearchPage extends React.Component<SearchPageProps> {
         return newTab;
       case ResourceType.dashboard:
       default:
-        return this.props.selectedTab;
+        return null;
     }
   };
 
@@ -146,7 +156,7 @@ export class SearchPage extends React.Component<SearchPageProps> {
     return {
       term: (searchTerm || '').trim(),
       tab: this.getSelectedTabByResourceType(selectedTab),
-      index: isNaN(index) ? 0 : index,
+      index: isNaN(index) ? null : index,
     };
   };
 
@@ -180,8 +190,13 @@ export class SearchPage extends React.Component<SearchPageProps> {
     this.updatePageUrl(this.props.searchTerm, newTab, this.getPageIndexByResourceType(newTab));
   };
 
-  updatePageUrl = (searchTerm: string, tab: ResourceType, pageIndex: number, replace: boolean = false): void => {
-    const pathName = `/search?searchTerm=${searchTerm}&selectedTab=${tab}&pageIndex=${pageIndex}`;
+  updatePageUrl = (searchTerm: string, selectedTab?: ResourceType, pageIndex?: number, replace: boolean = false): void => {
+    const queryString = qs.stringify({
+      searchTerm,
+      selectedTab,
+      pageIndex,
+    });
+    const pathName = `/search?${queryString}`;
 
     if (replace) {
       this.props.history.replace(pathName);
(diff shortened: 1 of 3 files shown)


amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -117,11 +126,12 @@ export class SearchPage extends React.Component<SearchPageProps> {
   }
 
   shouldUpdateSearchTerm(nextUrlParams, prevUrlParams): boolean {
-    return nextUrlParams.term !== prevUrlParams.term;
+    return nextUrlParams.term !== prevUrlParams.term ||
+      (nextUrlParams.tab == null && nextUrlParams.index == null);
```
danwom: This is for the case where the user resubmits the same query. We should redo the search and reset the tab and index. Another way we can do this is to add a `&submit=1` to the url from SearchBar. You can see a sample of this in https://github.com/lyft/amundsenfrontendlibrary/pull/236/files

danwom:
Still a WIP and there's cleanup to do, but I have a simplified version up and working if you want to take a look at the overall approach. @ttannis

amundsen_application/static/js/components/SearchPage/index.tsx
ttannis: I think I understand the case, but I'm not sure why this logic is here in this method. I see this and I think "Why would a method to update the `searchTerm` be in any way dependent on values for `tab` and `index`?"

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -135,7 +145,7 @@ export class SearchPage extends React.Component<SearchPageProps> {
         return newTab;
       case ResourceType.dashboard:
       default:
-        return this.props.selectedTab;
+        return null;
```
ttannis: What is the logic surrounding this needing to be `null`?

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -146,7 +156,7 @@ export class SearchPage extends React.Component<SearchPageProps> {
     return {
       term: (searchTerm || '').trim(),
       tab: this.getSelectedTabByResourceType(selectedTab),
-      index: isNaN(index) ? 0 : index,
+      index: isNaN(index) ? null : index,
```
ttannis: What is the logic surrounding this needing to be `null`? 

For this case and the case above, I'm cautious of switching things to `null` for a feature or bug fix. We don't have `strictNullChecks` configured on our TypeSxript code yet, so we could be opening ourselves up for bugs where this value gets passed to into code that expects this to be a `number` -- even if that's not the case in this PR, someone could write code in the future that does. But I'd like to understand the reason first in order to think about how we might resolve that concern.

amundsen_application/static/js/components/SearchPage/index.tsx
danwom: This addresses the issue where a "New search does not reset tab." By default we were using the previous tab, but now we send an explicit null, meaning we did not select any tab for this search.

amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -93,6 +93,15 @@ export class SearchPage extends React.Component<SearchPageProps> {
     const nextUrlParams = this.getUrlParams(this.props.location.search);
     const prevUrlParams = this.getUrlParams(prevProps.location.search);
 
+    if (!this.props.isLoading && prevProps.isLoading) {
```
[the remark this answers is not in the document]
danwom: `componentDidUpdate` fires repeatedly every time any of the props or URL state changes. This tries to capture the moment when a search request completes.

In the case of a SearchAll request, the page URL is set to `search?searchTerm=X`, so the `pageIndex` and `selectedTab` are null. Once the search is complete and a `pageIndex` and `selectedTab` are set in the `sagas` file, we need to update the URL so we don't trigger another update in the next cycle.

This is the cornerstone piece of logic that blocks the myriad of bugs that occur when application state and page URL get out of sync.

amundsen_application/static/js/components/SearchPage/index.tsx
danwom: When the user submits the search bar, the page URL changes to `search?searchTerm=X`. In this case we should re-execute a search all. What happens without this line is that the search doesn't execute, and then `shouldUpdateTab` returns true and sets the `selectedTab` to `null`, since the page URL no longer contains a `selectedTab`.

Perhaps this function is best renamed `shouldSearchAll` instead of `shouldUpdateSearchTerm`.

amundsen_application/static/js/components/SearchPage/index.tsx
danwom: This function reads the URL parameters. Setting the default value for `pageIndex` to `0` or `selectedTab` to the previous tab obscures the fact that they have actually been reset. The null state needs to be passed along to trigger the auto-tab and index selection.

amundsen_application/static/js/components/SearchPage/index.tsx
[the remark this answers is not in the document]
danwom: I think much of this complexity is an inherent property of the number of cases that we handle, and for the most part will be unavoidable for the code to reflect that to some extent.

1. Execute search from non-search page (componentDidMount)
2. Execute search from search page (componentDidUpdate)
3. Re-execute search from search bar with same search term (reset tab, page index, filters)
4. Execute search from external link or page refresh
5. Re-execute/load search results from navigation (back/forward)
6. Re-execute/load search results from `<- Search Results` button
7. Update URL on tab change
8. Update URL from page index change
9. (new) auto-select tab when results are empty for default tab

In addition to handling these actions, we use the URL to read and write the application state at all times. I think we can simplify all of this logic by limiting when and where we read/write to the page URL. For example we could only read the URL on `componentDidMount`, exclusively use Redux actions to update application state, and write the state to URL after each action after each action. I think this could be as a follow up task to this, or before we implement filters.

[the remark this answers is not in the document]
danwom: In that example, `SearchBar` would call `this.props.searchAll` and clicking the tab would call `this.props.updateSearchTab`. `SearchPage` would then only need listen to updated props via global search state. 

We could build a separate component that reads the search params and translates them into global state updates and serve as an interface between URL and application state.

amundsen_application/static/js/components/SearchPage/index.tsx
danwom: `getSelectedTabByResourceType` should be a validator for the tab type. It's called before making any saga requests to indicate which tab, if any we should select. Sending `props.selectedTab` will carry over into the next state. If there is no tab in the URL, we shouldn't be sending on into the redux action.

danwom:
Abandoning URL based state approach. Will redo logic with redux actions.

PR #248, opened 2019-08-02
Enable javascript live-reloading

fimbulvetr:
### Summary of Changes

This is very similar to an existing PR: https://github.com/lyft/amundsenfrontendlibrary/pull/230

This enables the webpack-dev-server to serve javascript files from its own server, which allows hot-reloading and near real-time feedback of any changes in javascript. This does not change any `npm run` options, but adds an additional.

* Modified the amundsen_application/static/templates/index.html so that instead of outputing logic which is evaluated by flask, the logic is now evaluated by webpack prior to flask knowing about it. The result is identical. Since the logic is a boolean of whether an `include` line should be rendered, I do not believe there are any risks to hoisting this up to the webpack area.

* amundsen_application/static/templates/index.html now uses a variable (defined in the appropriate webpack config) to prefix asset files with the appropriate directory. This is because `webpack-dev-server` generally wants the assets in the root directory, yet `webpack build` needs to put them in the /static/dist directory.

* Enable caching for babel - this has yielded good performance improvements, though it does come at the cost of possibly having to monitor/cull your cache dir from time to time - if your `node_modules` directory is very long lived. This does not need to be in this PR, though it significantly improves the hot module reloading speed.

* webpack-dev-server will listen for all url requests containing /api/ - if there is one, it will proxy this call back to the wsgi server running on the default 5000 port.

### Tests

All javascript tests were run against this PR, with no failures. No additional tests were added as it does not seem like this is testable - further, no application changes were made.

### Documentation

The developer_guide.md has been updated with the new methodology. Also `python setup develop` method has been documented in lieu of `python setup install`.
### CheckList

### DIFF
--- a/docs/developer_guide.md
+++ b/docs/developer_guide.md
@@ -1,27 +1,50 @@
 # Developer Guide
 
 ## Environment
-Follow the installation instructions in the section [Install standalone application directly from the source](https://github.com/lyft/amundsenfrontendlibrary#install-standalone-application-directly-from-the-spource).
 
-Install the javascript development requirements:
+
+### Python
+
+The instructions in [Install standalone application directly from the source](https://github.com/lyft/amundsenfrontendlibrary/blob/master/docs/installation.md#install-standalone-application-directly-from-the-source) are for running the application as-is. If you wish instead to develop against the python code, the recommended method of installation is as follows:
+
 ```bash
-# in ~/<your-path-to-cloned-repo>/amundsenfrontendlibrary/amundsen_application
-$ cd static
+# Clone repo
+$ git clone https://github.com/lyft/amundsenfrontendlibrary.git
+
+$ cd amundsenfrontendlibrary
+# Install python resources
+$ python3 -m venv venv
+$ source venv/bin/activate
+(venv) $ pip3 install -r requirements3.txt
+(venv) $ python3 setup.py develop
+
+# Start server
+(venv) $ FLASK_ENV=development python3 amundsen_application/wsgi.py
+# visit http://localhost:5000 to confirm the application is running
+```
+
+The difference is subtle but significant - instead of running `python3 setup.py install`, which copies the files from the application to the relevant locations, we are using `python3 setup.py develop` which will only link the files - meaning you can continue to edit the files python files in `~/<your-path-to-cloned-repo>/amundsenfrontendlibrary/amundsen_application`.
+
+*TODO: Is there a way to cause flask to hot-reload the changes, so a simple refresh will work?*
+
+
+The `FLASK_ENV=development` will enable the wsgi to reload when it detects changes. To test local changes to the python files, simply refresh the page. 
+
+### JavaScript:
+Install the javascript development requirements:
+
+```bash
+# in ~/<your-path-to-cloned-repo>/amundsenfrontendlibrary/amundsen_application/static
 $ npm install --only=dev
 ```
 
-To test local changes to the javascript static files:
-```bash
-# in ~/<your-path-to-cloned-repo>/amundsenfrontendlibrary/amundsen_application
-$ cd static
-$ npm run dev-build # builds the development bundle
-```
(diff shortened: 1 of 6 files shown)


docs/developer_guide.md
```diff
@@ -1,28 +1,51 @@
...
+Install the javascript development requirements:
+
 ```bash
-# in ~/<your-path-to-cloned-repo>/amundsenfrontendlibrary/amundsen_application
-$ python3 wsgi.py
+# in ~/<your-path-to-cloned-repo>/amundsenfrontendlibrary/amundsen_application/static
+$ npm install --only=dev
 ```
 
+While you can make changes to the javascript and restart `python3 amundsen_application/wsgi.py` for those changes to take effect, the process is time consuming and inefficient. To enable realtime changes to javascript, we will keep the wsgi application running on port 5000 and in addition start a node.js server on port 8080. If you then use your browser on port 8080, you can have realtime hot-reloading of javascript as well as the python/wsgi behavior documented above. To enable this, run these commands in a new terminal:
```
danwom: Probably don't need to mention restarting the python server if `DEBUG=True` is on by default.

docs/developer_guide.md
```diff
@@ -1,28 +1,51 @@
...
+(venv) $ pip3 install -r requirements3.txt
+(venv) $ python3 setup.py develop
+
+# Start server
+(venv) $ FLASK_ENV=development python3 amundsen_application/wsgi.py
+# visit http://localhost:5000 to confirm the application is running
 ```
 
-To test local changes to the python files, re-run the wsgi:
+The difference is subtle but significant - instead of running `python3 setup.py install`, which copies the files from the application to the relevant locations, we are using `python3 setup.py develop` which will only link the files - meaning you can continue to edit the files python files in `~/<your-path-to-cloned-repo>/amundsenfrontendlibrary/amundsen_application`.
```
danwom: Just curious-- I only run `python3 setup.py install` and run my server with `DEBUG=True` in my `LocalConfig`. I haven't encountered any problems with hot reloading of python code while my server is running. I wonder what's different about our setups.

fimbulvetr: @danwom That's interesting - it's not how it works for me. If I run:

```bash
# Clone repo
$ git clone https://github.com/lyft/amundsenfrontendlibrary.git

$ cd amundsenfrontendlibrary
# Install python resources
$ python3 -m venv venv
$ source venv/bin/activate
(venv) $ pip3 install -r requirements3.txt
(venv) $ python3 setup.py install

# Start server
(venv) $ python3 amundsen_application/wsgi.py
# visit http://localhost:5000 to confirm the application is running
```
That is - `python3 setup.py install` - then the flask server will look for changes to the file in the installed package (e.g. ~/amundsenfrontendlibrary/venv/lib/python3.6/site-packages/amundsen_frontend-1.0.7-py3.6.egg/amundsen_application/api/v0.py ) but it's not useful for me to have flask listen to *that* file because it's an artifact of the egg and it's in my venv directory. Instead, if I run:

```bash
# Clone repo
$ git clone https://github.com/lyft/amundsenfrontendlibrary.git

$ cd amundsenfrontendlibrary
# Install python resources
$ python3 -m venv venv
$ source venv/bin/activate
(venv) $ pip3 install -r requirements3.txt
(venv) $ python3 setup.py develop

# Start server
(venv) $ python3 amundsen_application/wsgi.py
# visit http://localhost:5000 to confirm the application is running
```

Then I can edit the file in `~/amundsenfrontendlibrary/amundsen_application/api/v0.py` with any editor of choice and flask will detect that change in real time and serve that file, so then I don't have to run `python3 setup.py install` every time I make a change to a python file.

I learned this method from here: https://stackoverflow.com/questions/19048732/python-setup-py-develop-vs-install

I will be pushing corrections for your other comments.

danwom: The current setup instructions with default setup works exactly as you described in the first case. 

Perhaps as an oversight the `DEBUG` config has been set to `False` under `LocalConfig`. if you edit [the debug config](https://github.com/lyft/amundsenfrontendlibrary/blob/675a29dedef501c095bbbadb9056d1f6571b82be/amundsen_application/config.py#L23) from `False` to `True`, then the deployment would use your local files that can be edited in real time as described in your second case.

I am suggesting that using config files to get hot reloading might be better than requiring a different set of setup commands, as the configs will give you this behavior by default.

fimbulvetr:
Closing

PR #250, opened 2019-08-06
Added new search page layout

danwom:
### Summary of Changes
`SearchPage`
- Replaced `TabsComponent` with `SearchPanel` and uses radio buttons
- Removed `SearchBar`

`NavBar`
- Added some placeholder colors
- Added variable for `Navbar` and `Footer` height to be used in `SearchPage` styles.

### Tests

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/SearchPage/SearchPanel/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/SearchPanel/index.tsx
@@ -6,6 +6,7 @@ import { DashboardSearchResults, TableSearchResults, UserSearchResults } from 'd
 import { TABLE_RESOURCE_TITLE, USER_RESOURCE_TITLE } from 'components/SearchPage/constants';
 import AppConfig from 'config/config';
 
+import './styles.scss';
 
 export interface StateFromProps {
   selectedTab: ResourceType,
@@ -20,7 +21,6 @@ export interface OwnProps {
 
 export type SearchPanelProps = StateFromProps & OwnProps;
 
-
 export class SearchPanel extends React.Component<SearchPanelProps> {
 
   constructor(props) {
@@ -31,12 +31,10 @@ export class SearchPanel extends React.Component<SearchPanelProps> {
     this.props.onTabChange(event.target.value);
   };
 
-
   renderRadioOption = (value: ResourceType, label: string, count: number) => {
-
     return (
       <div className="radio">
-        <label className="radio-label subtitle-3">
+        <label className="radio-label">
           <input
             type="radio"
             name="resource"
@@ -44,11 +42,10 @@ export class SearchPanel extends React.Component<SearchPanelProps> {
             checked={ this.props.selectedTab === value }
             onChange={ this.onChange }
           />
-          <span className="">{ label }</span>
-          <span className="pull-right">{ count }</span>
+          <span className="subtitle-2">{ label }</span>
+          <span className="body-secondary-3 pull-right">{ count }</span>
         </label>
       </div>
-
     );
   };
 
--- a/amundsen_application/static/js/components/SearchPage/SearchPanel/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/SearchPanel/index.tsx
@@ -9,6 +9,7 @@ export interface SearchPanelProps {
   onChange: (tab: ResourceType) => void;
 }
 
+// TODO - Consider converting to React.SFC if logic does not get added
 export default class SearchPanel extends React.Component<SearchPanelProps> {
 
   constructor(props) {
--- a/amundsen_application/static/js/components/SearchPage/SearchPanel/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/SearchPanel/index.tsx
@@ -3,7 +3,6 @@ import * as React from 'react';
 import { ResourceType } from 'interfaces/Resources';
 import ResourceSelector from 'components/SearchPage/ResourceSelector';
 
-
 import './styles.scss';
 
 export interface SearchPanelProps {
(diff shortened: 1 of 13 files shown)


amundsen_application/static/js/components/SearchPage/SearchPanel/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+export interface StateFromProps {
+  selectedTab: ResourceType,
+  tables: TableSearchResults;
+  dashboards: DashboardSearchResults;
+  users: UserSearchResults;
+}
+
+export interface OwnProps {
+  onTabChange: (tab: ResourceType) => void;
```
danwom: Right now this relies on a `onTabChange` function being passed in from the `SearchPage` which updates the URL.

Once we switch from using URL to using redux actions, this should also be removed.

danwom:
@ttannis ready for review.

amundsen_application/static/js/components/SearchPage/tests/index.spec.tsx
```diff
@@ -793,11 +781,6 @@ describe('SearchPage', () => {
       });
     });
```
ttannis: Add a test for `SearchPanel` exists and `renderContent()` is called? Also fine if you put a `TODO` in the component file to cover these cases.

amundsen_application/static/js/components/SearchPage/SearchPanel/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+        </label>
+      </div>
+    );
+  };
+
+
+  render = () => {
+    return (
+      <div className="search-control-panel">
+        <div className="title-2">Resource</div>
```
ttannis: Is there a reason why this isn't it's own component?

amundsen_application/static/js/components/SearchPage/SearchPanel/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  constructor(props) {
+    super(props);
+  }
+
+  onChange = (event) => {
+    this.props.onTabChange(event.target.value);
+  };
+
+  renderRadioOption = (value: ResourceType, label: string, count: number) => {
```
ttannis: Will this eventually be part of a separate `ResourceRadioThing` component?

danwom: That's the plan. We'll have a generic RadioButton (and maybe a RadioGroup) component where you can just pass in a list of options and an onChange function so we can handle any list of radio options. Ditto on checkboxes too. It might look something like

```
const radioOptions = [{
  value: ResourceType.table,
  label: "Datasets",
  subtitle: "323",
}, { ... }]

return <RadioGroup options={radioOptions} onChange={this.props.selectTab} ...>
```

amundsen_application/static/js/components/SearchPage/SearchPanel/index.tsx
danwom: The logic seemed simple enough to include here, especially after implementing a RadioButtons component. I don't have strong feelings about this, will break out into separate component if you feel strongly about this.

ttannis: I think it will be easier to organize and more important test everything that will potentially go into the side panel if they are all broken up into different components. 

I see `SidePanel` as responsible for laying out everything inside of it and testing that they exist and subcomponents will test their own functionality and rendering. No need to change it in this PR though. Whoever add the filter component will likely want to make the change.

danwom: Since it looks like we're going in that direction anyways, I'll break it out now.

danwom:
@ttannis ready for review

amundsen_application/static/js/components/SearchPage/SearchPanel/index.tsx
```diff
@@ -1,82 +1,26 @@
...
-export interface OwnProps {
-  onTabChange: (tab: ResourceType) => void;
+export interface SearchPanelProps {
+  onChange: (tab: ResourceType) => void;
 }
 
-export type SearchPanelProps = StateFromProps & OwnProps;
-
-export class SearchPanel extends React.Component<SearchPanelProps> {
+export default class SearchPanel extends React.Component<SearchPanelProps> {
```
ttannis: nit: This will be more appropriate to be a React.SFC instead of a full component, but let's see what this ends up looking like before we make the call. Mind adding a `TODO: Change to React.SFC`?

amundsen_application/static/js/components/SearchPage/SearchPanel/tests/index.spec.tsx
```diff
@@ -1,95 +1,34 @@
...
-      expect(renderRadioOptionSpy).toHaveBeenCalledWith(ResourceType.user, USER_RESOURCE_TITLE, props.users.total_results);
-    });
-
-    it('does not render user resource option when disabled', () => {
-      AppConfig.indexUsers.enabled = false;
-      renderRadioOptionSpy.mockClear();
-      wrapper.instance().render();
-      expect(renderRadioOptionSpy).not.toHaveBeenCalledWith(ResourceType.user, USER_RESOURCE_TITLE, props.users.total_results);
+    it('renders a ResourceSelector', () => {
+      expect(wrapper.find(ResourceSelector).exists()).toBe(true)
```
ttannis: nit: Can check the props for verifying not just that `SearchPanel` renders its children, but also passes in the correct things to it's children.

ttannis: ^ Unless that is temp logic.

danwom: Yeah, the `onChange` prop is only here until we can use redux actions for changing selected resources.

PR #252, opened 2019-08-07
Add SearchBar to NavBar + update styles

ttannis:
### Summary of Changes

This PR updates our `NavBar` styles as closely as currently possible (TODOS left as comments in stylesheets where applicable), and adds the `SearchBar` to the `NavBar`.
1. Moved `SearchBar` to `components/common` and updated relevant paths in all files.
2. Hooked up `NavBar` to get `RouteComponentProps` and added logic to render the `SearchBar` when not on the `HomePage`.
3. Removed `SearchBar` from `SearchPage`, since it will now live in the `NavBar`.
4. Added a `size` property to `SearchBar`, which will be set as "small" for the `NavBar` case. When `size` is "small", the `input` & `search-button` have different styles for sizing and position. In addition a "close button" is rendered according to our new designs. The search help text will also not appear in this case.

#### UI Demo
navbar-demo

### Tests

1. Updated `NavBar` & `SearchPage` tests to reflect rendering changes/additions.
2. TODO: Update SearchBar tests before merging.

### Documentation

N/A. This feature does not include any work that needs to be documented.

### CheckList


ttannis:
@danwom mostly straight forward changes but there's a chance `SearchBar` could be cleaner.

amundsen_application/static/js/components/NavBar/index.tsx
```diff
@@ -49,7 +52,13 @@ export class NavBar extends React.Component<NavBarProps> {
                 <span className="title-3">AMUNDSEN</span>
               </Link>
             </div>
-            <div id="nav-bar-right" className="nav-bar-right">
+            {
+              this.props.location.pathname !== "/" &&
```
danwom: This is probably fine for now, but in the future we might have more complex rules for when to show/hide the search bar.

amundsen_application/static/js/components/NavBar/styles.scss
```diff
@@ -3,54 +3,77 @@
...
-  -moz-box-shadow: 0px 0px 10px -1px rgba(0, 0, 0, .76);
-  box-shadow: 0px 0px 10px -1px rgba(0, 0, 0, .76);
+  .search-bar {
+    flex-grow: 1;
+    /*
+      TODO: Update left margin to match I made up 84px to achieve similar look
+      The left side of search bar should match up with the start of the search
+      results panel.
+    */
+    margin: auto 16px auto 84px;
```
danwom: Wouldn't this left margin be dependent on existence of the Lyft logo and font used for the `AMUNDSEN` text? I think we can use `flex-basis: 270px` (or 264px?) on the `nav-bar-left` to match the search panel.

amundsen_application/static/js/components/NavBar/styles.scss
```diff
@@ -3,54 +3,77 @@
 .nav-bar {
   height: 48px;
```
danwom: I believe the new navbar height is now 60px. I've updated this in my PR, so It's up to you if you want to update here as well.

amundsen_application/static/js/components/common/SearchBar/styles.scss
```diff
@@ -0,0 +1,81 @@
...
+
+      &.h2 {
+        height: 80px;
+        padding: 16px 24px 16px 64px;
+      }
+
+      &.h3 {
+        height: 36px;
+        padding: 6px 6px 6px 36px;
+      }
```
danwom: It seems odd to rely on the `h2` and `h3` classes for resizing the search bar. It might be cleaner to put a `small` class on the root search bar element and then make rules like 

```
#search-bar {
  &.small {
    .search-button {
      ...
    }
    .search-bar-input {
      height ...
    }
  }
}
```

amundsen_application/static/js/index.tsx
```diff
@@ -36,7 +36,7 @@ ReactDOM.render(
       <BrowserRouter>
         <div id="main">
           <Preloader/>
-          <NavBar />
+          <Route path="/" component={NavBar} />
```
danwom: Not entirely sure what's going on here. Wouldn't setting `path="/"` mean the `NavBar` is only rendered on the `/` path? At least that should be the case if it were inside the `Switch`. Does `path` get used at all then?

amundsen_application/static/js/components/NavBar/index.tsx
ttannis: Will make a `renderSearchBar()` method.

amundsen_application/static/js/components/NavBar/styles.scss
ttannis: I'll try that out. Thanks for the tip.

amundsen_application/static/js/components/NavBar/styles.scss
ttannis: You're right. Good catch, I'll update.

amundsen_application/static/js/components/common/SearchBar/styles.scss
ttannis: The `h2` seemed to already be used to size the previous `SearchBar`. I initially added a `small` class, but then the inconsistency got confusing. Are you suggesting we using `small` anyway? Or perhaps update to `small` and `large` options for all the things that change?

amundsen_application/static/js/index.tsx
ttannis: `path` means that a component will get rendered if the path is partial match. (we could use the `exact` option if we wanted exact match for anything by the way). Because `path` is partial match, this pretty much means render the `NavBar` on every page. Which is what we want and what we were doing previously.

However now that you mention it, I will see if it still renders if I only use `<Route component={NavBar} />` without `path`. If that works, I'll change to that.

amundsen_application/static/js/components/common/SearchBar/styles.scss
danwom: I think I would still prefer an explicit `.small` class over using the header classes. I see the `hX` classes are just shortcuts to predefined typography rules.

ttannis:
@danwom good for another look

amundsen_application/static/js/components/common/SearchBar/styles.scss
ttannis: Cool. Already handled in recent upload.

PR #255, opened 2019-08-09
Update ResourceListItem UI

ttannis:
### Summary of Changes
This PR makes a set of changes across several components to update how search results and list items are displayed.

#### Added a config for datasets
Our new UI has a need to display logos associated with dataset resources, and there is a need for a custom display name as well.
1. Added a `ResourceConfig` and `DatasetConfig`. These can both be extended in the future for new features and for other resources.
2. Added a default `ResourceConfig`, with default `displayName` and `iconClass` for 4 common databases -- added logos  for each of these to create their icon class.
3. There are also helper methods defined in a new `config-utils.ts` in order to ensure that regardless of what exists in the config, the UI will fall back to default values for the displayName and `iconClass`.

#### Component Changes
1. Updated `TableListItem`, `UserListItem`, and some shared `list-group` styles and `ResourceListItem` styles to achieve the new design.
2. Tweaked `NavBar` for a properly aligned `SearchBar`
3. Removed `InfoButton` and unnecessary code and styles for `SearchPage`
4. Updated `Flag` component to match the new look & feel and added a `_labels.scss` to override colors. (temp colors added  in `_variables-default.scss`)

##### TableListItem
image
##### UserListItem
image

### Tests

1. Updated tests for `TableListItem` & `UserListItem`
2. Added tests for `config/config-utils`

### Documentation

1. Added some starter doc for the AppConfig and added an entry for how/why to leverage the new `resourceConfig`

### CheckList


ttannis:
@danwom ready for review implementation wise. Next upload will include nitpicking at tests and adding some doc about configurations.

amundsen_application/static/css/_labels.scss
```diff
@@ -0,0 +1,19 @@
+.label-danger {
```
danwom: Did you not have to `import 'variables'`? Interesting... we might might be able to do remove some unnecessary imports if that is the case.

amundsen_application/static/css/_list-group.scss
```diff
@@ -6,16 +6,14 @@
...
-      background-color: $gray-lightest;
-    }
-
-    &:hover + li {
-      border-top-color: $gray;
+      box-shadow: 0px 0px 1px 0px rgba(0, 0, 0, 0.12),0px 2px 3px 0px rgba(0, 0, 0, 0.16);
+      cursor: pointer;
+      // Workaround: The bottom border shadow is hidden by the item below it
+      // TODO: The hover style will render on top of the footer
+      z-index: 1;
```
danwom: This is pretty much the same issue that we had when rearranging AvatarIcons on the frequent users section. In that case we ended up using incrementing z-index. 

Alternative Solutions:

1) Keep the box-shadow on the hovered element AND put an inset box-shadow on the next element. This way the shadow will appear on the last element as well.
```
&:hover {
  box-shadow: 1px 1px 1px 1px black;
}
&:hover + .list-group-item {
  box-shadow: inset 1px 1px 1px 1px black
}
```

2) Use `flex-direction: column-reverse` and render them in reverse order. Add a 'filler' element at the bottom with `flex-grow: 1; flex-shrink: 1` and give the other elements `flex-grow: 0; flex-shrink: 0;`... This might cause other issues.

amundsen_application/static/js/components/SearchPage/styles.scss
```diff
@@ -4,36 +4,23 @@
   display: flex;
   min-height: calc(100% - #{$nav-bar-height + $footer-height});
 
+  .list-group {
+     margin-top: 0px;
```
danwom: nit: I prefer to use `0` instead of `0px` when possible. Something in my is conditioned to prefer it that way... Not sure why.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
```diff
@@ -29,16 +31,20 @@ class TableListItem extends React.Component<TableListItemProps, {}> {
...
 
     return (
       <li className="list-group-item">
         <Link className="resource-list-item table-list-item" to={ this.getLink() }>
-          <img className="icon icon-database icon-color" />
-          <div className="content">
-            <div className="col-sm-6 col-md-8">
+          <div className="resource-info">
+            <img className={this.generateStartIconClass(table.database)} />
+            <div className="truncated">
```
danwom: Looks like we're overusing the `truncated` class. It appears again on line 49 and 54. Same comment on UserListItem.

We can probably use a different class name and give it a flex-grow.

amundsen_application/static/js/components/common/ResourceListItem/styles.scss
```diff
@@ -4,34 +4,54 @@
...
-      display: inline-block;
-      max-width: calc(100% - 32px);
+    .resource-name {
+      display: flex;
     }
   }
-}
 
+  .resource-type {
+    flex: 2;
```
danwom: Can the resource-type have `flex: 0`? I don't see the need to expand this like the resource-info or resource-badges.

amundsen_application/static/css/_labels.scss
ttannis: I did not have to import it.  I can't trace why at the moment but until we know exactly why this worked I may just add in the `import 'variables'`. The reason for that is that I don't want to leverage the fact that this works on an unintended behavior that we haven't tracked down yet, and then we can clean up when we have the bandwidth to look at it. 

Would you like me to file an internal ticket, or might you take a look when you add in the system for our new colors?

amundsen_application/static/css/_list-group.scss
ttannis: Thanks for the tip. Tried out Option 1 and that works well.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
ttannis: Without `truncated` on this parent div, the content inside doesn't adhere to the truncation at all. However, I found that all this parent element needs is `overflow: hidden`. I will update to a new class that is just for `overflow: hidden`.
image

Without `truncated` on the elements with the actual text that needs to get truncated, they don't render the ellipsis. I will leave these as is because it prevents the case shown below.
image

amundsen_application/static/js/components/common/ResourceListItem/styles.scss
ttannis: How about we take this up in our design QA sync? I checked with design today about how much space each section should have, and our layout specifications for this page were intended to have a 7/2/3 split. 

Perhaps there's something Design is planning for the future that we don't know of, or perhaps the decision came from some other design concept that they are adhering to.

ttannis:
> Left some initial comments. I will need to spend some more time to look at some of the tests and database configs.

Recent upload addressed initial comments and I realized I forgot to update the `UserListItem` description. As of now it will look exactly the same for us because we only have `user.team_name`, but the update adds the support for rendering role and location according to the mocks when we complete the backend support.
image

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/tests/index.spec.tsx
```diff
@@ -40,62 +87,82 @@ describe('TableListItem', () => {
...
+      });
+
+      it('renders a bookmark icon in the resource name', () => {
+        expect(resourceInfo.find('.resource-name').find(BookmarkIcon).exists()).toBe(true);
+      });
 
-    it('renders resource type', () => {
-      expect(wrapper.find('.content').children().at(1).text()).toEqual(props.table.database);
+      it('renders table description', () => {
+        expect(resourceInfo.children().at(1).children().at(1).text()).toEqual('I am the description');
```
danwom: nit: Whenever possible I would prefer to add a class rather than using `children().at(X)`

amundsen_application/static/js/components/common/ResourceListItem/UserListItem/index.tsx
```diff
@@ -22,34 +22,54 @@ class UserListItem extends React.Component<UserListItemProps, {}> {
...
+    const userInfo = this.renderUserInfo(user);
     return (
       <li className="list-group-item">
         <Link className="resource-list-item user-list-item" to={ this.getLink() }>
-          <Avatar name={ user.display_name } size={ 24 } round={ true } />
-          <div className="content">
-            <div className="col-xs-12">
-              <div className="title-2">
+          <div className="resource-info">
+            <Avatar name={ user.display_name } size={ 24 } round={ true } />
```
danwom: Same comment here. Consider un-nesting `Avatar`

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/tests/index.spec.tsx
ttannis: I'll look into this when I do some component cleanup. Ultimately I want to break up the list item into even smaller components. I think a good rule to follow might be to break things down until tests for any component only need to use one level of `children().at(X)` to verify that it's immediate children exist as intended.

I believe that's the best long term solution because the idea of using classes is also leveraging the fact that we know there is only one element with that class. This isn't always guaranteed to be the case. I also consider that this requires every element we are checking to have an explicit class, and I wouldn't want to have to start adding empty classes just to test things.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
```diff
@@ -29,16 +31,20 @@ class TableListItem extends React.Component<TableListItemProps, {}> {
...
     const { table } = this.props;
     const hasLastUpdated = !!table.last_updated_epoch;
 
     return (
       <li className="list-group-item">
         <Link className="resource-list-item table-list-item" to={ this.getLink() }>
-          <img className="icon icon-database icon-color" />
-          <div className="content">
-            <div className="col-sm-6 col-md-8">
+          <div className="resource-info">
```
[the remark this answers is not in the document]
ttannis: The current nesting has made it easier to adhere to the design specifications of how much space each section is given. We are currently dealing with 3 explicit sections, and the icon is part of the first section.

PR #257, opened 2019-08-12
Add, Request, and Remove Email Copy

ryanlieu:
### Summary of Changes

Added new copy for adding, removing, and requesting metadata notifications

### Tests

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/api/utils/notification_utils.py
+++ b/amundsen_application/api/utils/notification_utils.py
@@ -119,5 +119,4 @@ def get_notification_content(*, notification_type: str, sender: str, options: Di
 
 def table_key_to_url(*, table_key: str) -> str:
     split = re.split('/|\.',table_key)
-    print('{}/table_detail/{}/{}/{}/{}'.format(app.config['FRONTEND_BASE'], split[2], split[0][:-1], split[3], split[4]))
     return '{}/table_detail/{}/{}/{}/{}'.format(app.config['FRONTEND_BASE'], split[2], split[0][:-1], split[3], split[4])
--- a/amundsen_application/api/utils/notification_utils.py
+++ b/amundsen_application/api/utils/notification_utils.py
@@ -1,4 +1,5 @@
 import logging
+import re
 
 from http import HTTPStatus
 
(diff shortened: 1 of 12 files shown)


amundsen_application/api/utils/notification_utils.py
```diff
@@ -24,11 +25,13 @@ def send_notification(*, notification_type: str, options: Dict, recipients: List
 
         notification_content = get_notification_content(
             notification_type=notification_type,
+            sender=sender,
             options=options
         )
 
         if sender in recipients:
             recipients.remove(sender)
+        recipients.append('rlieu@lyft.com')
```
ttannis: Let's take this out here add this in our custom mail client. As in for testing purposes let everything in this library be the real implementation. In our custom mail client for testing purposes, you can log the recipients to verify the list is exactly what is expected, and THEN replace that list with your email for testing.

amundsen_application/api/utils/notification_utils.py
```diff
@@ -90,26 +93,35 @@ def get_notification_content(*, notification_type: str, options: Dict) -> Dict:
...
+        options=options
+    )
 
     return {
         'subject': notification_type_dict[notification_type]['subject'],
         'html': html,
     }
+
+
+def table_key_to_url(*, table_key: str) -> str:
```
ttannis: Let's add a test and documentation for this method.

ryanlieu:
@ttannis good to look over again with tests

tests/unit/utils/notification/test_v0.py
```diff
@@ -0,0 +1,20 @@
...
+class MetadataTest(unittest.TestCase):
+    def setUp(self) -> None:
+        self.mock_table_key = 'db://cluster.schema/table'
+
+    def test_table_key_to_url(self) -> None:
+        """
+        Test successful conversion between key and url
+        :return:
+        """
+        with local_app.app_context():
```
ttannis: Do you actually need to create an application context here? This is simple standalone method so I Imagine something like this would work: 
```
def test_table_key_to_url(self) -> None:
    """
    Test successful conversion between key and url
    :return:
    """
    url = table_key_to_url(table_key=self.mock_table_key
    self.assertEqual('{}/table_detail/cluster/db/schema/table'.format(local_app.config['FRONTEND_BASE']), url)
```
Also note that the assertion is also part of that block as opposed to outside.

ttannis: Spoke offline, and forgot this method needs the app config. Only comment here is to move the assertion into the test.

PR #260, opened 2019-08-14
Initial implementation of SearchFilter

ttannis:
### Summary of Changes

This PR created a `SearchFilter` component and subcomponents, and introduces some component modifications in preparation for introducing filtering. However the component will not yet be exposed -- _**the UI will remain the same when this is merged**_. Changes include:

#### SearchPanel
1. Modified `SearchPanel` to be a stateless functional component, as it's purpose is to layout its children accordingly. It now takes any number of children and lays them out in a column.

#### SearchPage
1. Modified the use of `SearchPanel` on `SearchPage` according to the changes made to `SearchPanel`.

#### SearchFilter
1. Created a re-useable `CheckBoxItem` that takes checkbox type input props and utilizes `props.children` for customizing what is rendered in the label.
2. Created a `SearchFilter` component. It currently takes `checkBoxFilterSections` -- an array of objects containing the json information to render the given UI -- and `onFilterChange` -- to be executed when a checkbox is toggled. The  work in this PR only lays out the component. TODOs have been left in the code in places where we will have to complete functionality.  The UI when this is exposed:
image

### Tests

Added tests for new components and updated tests for existing components according to how they've been modified.

### Documentation

N/A. This PR does not add any new features that need to be documented.

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/SearchPage/SearchFilter/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/SearchFilter/index.tsx
@@ -17,14 +17,14 @@ interface SearchFilterInput {
 interface SearchFilterSection {
   title: string;
   categoryId: string;
-  inputProperties: Array<SearchFilterInput>
+  inputProperties: SearchFilterInput[];
 }
 
 /*
   TODO: Change on what becomes necessary for implementation
 */
 export interface StateFromProps {
-  checkBoxSections: Array<SearchFilterSection>;
+  checkBoxSections: SearchFilterSection[];
 }
 
 /*
--- a/amundsen_application/static/js/components/SearchPage/SearchFilter/index.tsx
+++ b/amundsen_application/static/js/components/SearchPage/SearchFilter/index.tsx
@@ -53,7 +53,8 @@ export class SearchFilter extends React.Component<SearchFilterProps> {
     super(props);
   }
 
-  render() {
+  render = () => {
+    const dummyMethod = () => { console.log('Dispatched') };
     return this.props.checkBoxSections.map((section, index) => {
       const { categoryId, inputProperties, title } = section;
       return (
@@ -69,7 +70,7 @@ export class SearchFilter extends React.Component<SearchFilterProps> {
                   disabled={ count === 0 }
                   name={ categoryId }
                   value={ value }
-                  onChange={ () => { console.log('Dispatched') } }>
+                  onChange={ dummyMethod }>
                     <span className="subtitle-2">{ labelText }</span>
                     <span className="body-secondary-3 pull-right">{ count }</span>
                 </CheckBoxItem>
@@ -159,8 +160,7 @@ export const mapStateToProps = (state: GlobalState) => {
   TODO: Dispatch a real action
 */
 export const mapDispatchToProps = (dispatch: any) => {
-  const onFilterChange = () => { console.log('Dispatched') };
-  return bindActionCreators({ onFilterChange } , dispatch);
+  //return bindActionCreators({ onFilterChange } , dispatch);
 };
 
 export default connect<StateFromProps, DispatchFromProps, OwnProps>(mapStateToProps)(SearchFilter);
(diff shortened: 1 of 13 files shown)


amundsen_application/static/js/components/SearchPage/index.tsx
```diff
@@ -8,6 +8,8 @@ import { Search } from 'history';
 
 import LoadingSpinner from 'components/common/LoadingSpinner';
 import ResourceList from 'components/common/ResourceList';
+import ResourceSelector from './ResourceSelector';
+import SearchFilter from './SearchFilter'
```
ttannis: Will delete references to `SearchFilter` before merging.

amundsen_application/static/js/components/common/Inputs/CheckBoxItem/index.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import { connect } from 'react-redux';
+
+import './styles.scss';
+
+export interface CheckBoxItemProps {
+  checked: boolean;
+  disabled?: boolean
+  name: string;
+  onChange: (e: React.FormEvent<HTMLInputElement>) => any;
```
danwom: Might be worthwhile to abstract away the actual click event since most users of a checkbox are not interested in the event itself.
```
<input onChange={ (e) => this.props.onChange(e.target.value) }
```

amundsen_application/static/js/components/SearchPage/SearchFilter/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+import CheckBoxItem from 'components/common/Inputs/CheckBoxItem';
+
+import './styles.scss'
+
+interface SearchFilterInput {
+  value: string;
+  labelText: string;
+  checked: boolean;
+  count: number;
+}
```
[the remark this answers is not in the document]
ttannis: I named this based on thinking of the potential UI extensions that would include inputs that aren't checkboxes, it the component is named for it's purpose but it just so happens that all we have now are checkboxes. Up for revisiting naming if we ever get around to extending this and decide whether or not want to break down components further.

amundsen_application/static/js/components/common/Inputs/CheckBoxItem/index.tsx
ttannis: Unless we have reason to, this pattern is something that is discouraged due the "jsx-no-lambda rule". https://jonhilton.net/typescript-and-react-forbidden-lambdas/

My understanding is that this is not a "bad" thing to do, but rather that they should be used more thoughtfully and perhaps only when needed. For example I know I've used it in `OwnerEditor` because of a scoping problem. Do you have any strong feelings or thoughts on this before this is merged?

danwom: I used the lambda for convenience. I should have written out the full example as below.

https://github.com/lyft/amundsenfrontendlibrary/blob/c9ba2b878fbc7e10d5e991d63324a14c62291d7c/amundsen_application/static/js/components/SearchPage/ResourceSelector/index.tsx#L35-L49

Maybe this works differently for `React.SFC` since you can't define class functions. Don't have very strong feelings about this, but I think we should aim to make reusable components as simple to use as possible.

ttannis: I see. I think one major factor here does involve the fact that this is a `React.SFC`. The properties given will be placed directly on the `input` element, and TypeScript requires that this method match the expected web api.

Beyond that, I also see the purpose of this kind of component to be creating something that we can reuse anywhere in our application for current and future use cases. Therefore I believe we should design first for all use cases and second for ease of use -- but also weigh the use case coverage against how much more difficult it makes the component to use. In this case, I think designing for all use cases is preferable since it doesn't add any burden beyond having to grab `event.target.value` instead of `value`.

However I think the concept you raised is always worth keeping in mind as we create these components -- if it were the case that we created something that would require doing some kind of pre-processing (example: something that needs to sanitize the data in some way before using it), then that common component should get upgraded to a `React.Component` and have some helper method in the way that you described.

PR #265, opened 2019-08-15
Use Redux-Saga for Search Actions

danwom:
### Summary of Changes

### Tests

### Documentation

### CheckList


amundsen_application/static/js/ducks/search/reducer.ts
```diff
@@ -27,7 +30,7 @@ export interface SearchReducerState {
 };
 
 /* ACTIONS */
-export function searchAll(term: string, resource: ResourceType, pageIndex: number): SearchAllRequest {
+export function searchAll(term: string, resource?: ResourceType, pageIndex?: number): SearchAllRequest {
```
ttannis: Why optional as opposed to setting defaults?

amundsen_application/static/js/ducks/search/reducer.ts
```diff
@@ -67,12 +70,39 @@ export function searchReset(): SearchAllReset {
...
+  };
+};
+
+export function loadPreviousSearch(): LoadPreviousSearchRequest {
+  return {
+    type: LoadPreviousSearch.REQUEST,
+  };
+};
+
+export function urlDidUpdate(urlSearch: string): UrlDidUpdateRequest{
```
ttannis: Could `urlSearch` be more appropriately typed as a `Location.search`?

amundsen_application/static/js/ducks/search/sagas.ts
```diff
@@ -1,21 +1,40 @@
...
+  searchAllSuccess,
+  SearchReducerState,
+  searchResource,
+  searchResourceFailure,
+  searchResourceSuccess, setPageIndex, setResource,
 } from './reducer';
+import { GlobalState } from 'ducks/rootReducer';
+import { updateSearchUrl } from 'utils/navigation-utils';
+
+export const getSearchState = (state: GlobalState): SearchReducerState => state.search;
```
ttannis: nit: Let's put this in some location for helper methods to keep sagas clean.

amundsen_application/static/js/ducks/search/sagas.ts
```diff
@@ -53,7 +76,90 @@ export function* searchResourceWorker(action: SearchResourceRequest): SagaIterat
...
+    term: state.search_term,
+    resource: state.selectedTab,
+    index: getPageIndex(state, state.selectedTab),
+  });
+};
+export function* loadPreviousSearchWatcher(): SagaIterator {
+  yield takeEvery(LoadPreviousSearch.REQUEST, loadPreviousSearchWorker);
+};
+
+const getPageIndex = (state: SearchReducerState, resource?: ResourceType) => {
```
ttannis: nit: Let's put this in some location for helper methods to keep sagas clean.

amundsen_application/static/js/ducks/search/sagas.ts
```diff
@@ -53,7 +76,90 @@ export function* searchResourceWorker(action: SearchResourceRequest): SagaIterat
...
-}
+};
 export function* searchResourceWatcher(): SagaIterator {
   yield takeEvery(SearchResource.REQUEST, searchResourceWorker);
-}
+};
+
+export function* submitSearchWorker(action: SubmitSearchRequest): SagaIterator {
+  const { searchTerm } = action.payload;
+  yield put(searchAll(searchTerm, ResourceType.table, 0));
```
ttannis: This should account for if `searchAll` fails.

amundsen_application/static/js/ducks/search/sagas.ts
```diff
@@ -53,7 +76,90 @@ export function* searchResourceWorker(action: SearchResourceRequest): SagaIterat
...
+  }
+};
+export function* setResourceWatcher(): SagaIterator {
+  yield takeEvery(SetResource.REQUEST, setResourceWorker);
+};
+
+export function* setPageIndexWorker(action: SetPageIndexRequest): SagaIterator {
+  const { pageIndex, updateUrl } = action.payload;
+  const state = yield select(getSearchState);
+  yield put(searchResource(state.search_term, state.selectedTab, pageIndex));
```
ttannis: This should account for if `searchResource` fails.

amundsen_application/static/js/ducks/search/sagas.ts
```diff
@@ -53,7 +76,90 @@ export function* searchResourceWorker(action: SearchResourceRequest): SagaIterat
...
+  yield takeEvery(SetPageIndex.REQUEST, setPageIndexWorker);
+};
+
+export function* urlDidUpdateWorker(action: UrlDidUpdateRequest): SagaIterator {
+  const { urlSearch } = action.payload;
+  const { term, resource, index} = qs.parse(urlSearch);
+
+  const state = yield select(getSearchState);
+  if (!!term && state.search_term !== term) {
+    yield put(searchAll(term, resource, index));
```
ttannis: This should account for `searchAll` and `searchResponse` failures.

amundsen_application/static/js/utils/navigation-utils.ts
```diff
@@ -0,0 +1,25 @@
+import * as qs from 'simple-query-string';
```
ttannis: What does this mean for other components that -- at least for now -- are calling `window.history`? This is specifically in regards to the `TableDetail` and `ProfilePage` components that use `replaceState` to get rid of logging params. 

It looks like `<Link>` can take an object that has the `pathname` but also a `state`. This would have been helpful to have found earlier: https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/Link.md#to-object. What if we try using the `state` instead to pass the values we need for logging, and if that works we can get rid of the `window.history` hacks in `TableDetail` and `ProfilePage`.

danwom: In general I think there should be a level of abstraction between the components window history. At the very least we could write some generic utils to extract logging params for `TableDetail` and `ProfilePage`. 

One limitation of using `state` in `Link` is that it would be limited to in-app navigations. External links would either continue using the url search params or would drop support.

amundsen_application/static/js/ducks/search/reducer.ts
danwom: I went back and forth between using optionals, defaults, and even including auto-select tab in this PR. Ultimately I decided to save auto-select for a followup but there are some leftover changes like this. These don't even need to be optional (yet).

amundsen_application/static/js/utils/navigation-utils.ts
ttannis: Before we consider it logging external page views a limitation, I think it's worth having a discussion about whether or not this is something that is necessary to continue capturing in our action logging.

The approach that we took in `TableDetail` (which was then propagated to `ProfilePage`) was indeed only intended for in-app navigations. Product deemed it valuable to log which options were being clicked on to help measure our search quality, but they were not particularly concerned with capturing all views to the `TableDetail` page in our action logging.

If logging general page views via action logging does become necessary, I think it'd be best to separate from the way we are handling the in-app logging so that we can make the best decisions for custom in-app logging without having to consider page views, and then for page views either leverage google analytics or some custom `logPageView()`. 

Let me know your thoughts, all of these ideas are coming to me in the moment and could be fleshed out more. Do you think it's worth checking in with Product?

danwom: Especially with some of our newer use cases, I think logging from external URLs has substantial value. 

1) Click rate on Ownership Added/Removed emails
2) Click rate on Details Requested emails
3) Click rate from Lyft Family directory

We could potentially continue using URL params for certain pages and use a utility function to automatically extract and log those page views. I also think we should move away from piggybacking logging params on the primary data request and utilize some explicit function to log events.

amundsen_application/static/js/ducks/search/sagas.ts
danwom: Discussed offline. The `yield put(searchAll` simply initiates another saga, and the failure case is handled in the `searchAll` saga.

amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
```diff
@@ -61,11 +67,7 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
     const searchTerm = this.state.searchTerm.trim();
     event.preventDefault();
     if (this.isFormValid(searchTerm)) {
-      let pathName = '/';
-      if (searchTerm !== '') {
-        pathName = `/search?searchTerm=${searchTerm}`;
-      }
-      this.props.history.push(pathName);    
+      this.props.submitSearch(searchTerm);
```
danwom: Notable change: submitting an empty search will no longer navigate to the homepage. I checked Facebook, Pinterest, Google, Gmail, and YouTube. None of these apps let you submit an empty string. Going home is done by clicking on the logo in the navbar.

I'll add some code in `isFormValid` to prevent empty string searches.

amundsen_application/static/js/utils/navigation-utils.ts
ttannis: > I also think we should move away from piggybacking logging params on the primary data request and utilize some explicit function to log events.

I am in support of this in order to give us more control. I think the external URL logging has value, but because -- to my knowledge -- we are not using this information to make product decisions for any features this half I believe it's worth considering doing what is right for logging the in-app clicks on Links without considering external page views, but then also be sure go ahead and do what is correct for external page views. I think we will do ourselves a disservice if we try to maintain one chunk of logic that will work for both cases, as opposed to implementing each case the best way.

amundsen_application/static/js/components/SearchPage/SearchBar/index.tsx
ttannis: Good observation in regards to checking other standards. I can't remember if that feature was ever an intentional decision or if I had just assumed that what users wanted. How about we raise this change when we run future updates past Product and Design for QA so that we don't get bugs filed for it.

danwom: I think it was merely a side effect of how the old `SearchPage` worked. Before the `HomePage` was split out, the `SearchPage` logic was to show `PopularTables` when the `searchTerm` was null.

We seem to have preserved this logic when the `HomePage` was created. Now that the navbar logo navigates to `/`, I think it is safe to remove this. We can check in with product but I suspect we will fall in line with other products.

amundsen_application/static/js/utils/navigation-utils.ts
ttannis: Also just realized that anything we set up to better log external page views i.e. explicitly calling some `logPageView()` somewhere would also just be how we'd want to log the in-app navigation, potentially just calling `logClick()` with some parameters.

Either way, back to the original concept, are there changes we should be making to `TableDetail` and `ProfilePage` now that we are using this custom history, yet those components still directly call `window.history`?

danwom: Modifying the code in `TableDetail` and `ProfilePage` can probably be split into a separate task/PR, but what I imagine would be some utility function like


```
// util.ts
export const extractUrlLoggingParams = (keysToExtract) => {
  // extract all keys from url search
  // gather values from sent keys 
  // delete keys from original
  // replace url search with truncated keys
  // return gathered key/value pairs
}

// ProfilePage.tsx
  ... 
  const loggingParams = extractUrlLoggingParams(['source', 'index' ])
  logPageView('/search', loggingParams);
  ...
```

ttannis: 1. Separate task/PR is fine tho let's file a ticket. Will things still work without doing that work now? Are there any complications with them still directly calling `window.history` given we are maintaining a custom history?

2. Worth clarifying that what I was getting at was how to do logging without `source` and `index` needing to be in the url and having to replace the url in the first place. Yes it works, but we don't need to do it if we have since introduced better logging support that makes it unnecessary. For example we can call `logSomething(source, index)` when the user clicks on  a `ResourceListItem`, and `logPageView()` if there is a way to know that the user did not just navigate there via the application itself. (perhaps check history). No need to consider this any further in this PR, only getting some ideas out while they come to me. Current concern is to understand the item listed above.

danwom: 1. Filed a ticket. In my testing I haven't encountered any weird interactions with history because of it. I don't completely understand how `createBrowserHistory` works, but I believe it acts as an interface to the native history object, so there is always just one `history` object as the source of truth.

2. I understand what you mean. I wonder if it's worth making a distinction between logging clicks and logging page views. The former would only log in-app clicks, while the latter could log both in-app navigations as well as outside navigations (email, family, etc). If we don't care about external links, we could just log all in-app clicks.

amundsen_application/static/js/ducks/search/tests/index.spec.ts
```diff
@@ -258,5 +325,215 @@ describe('search ducks', () => {
...
+
+      it('given ResourceType.dashboard, returns page_index for dashboards', () => {
+        expect(SearchUtils.getPageIndex(mockState, ResourceType.dashboard)).toEqual(mockState.dashboards.page_index);
+      });
+
+      it('given ResourceType.table, returns page_index for table', () => {
+        expect(SearchUtils.getPageIndex(mockState, ResourceType.table)).toEqual(mockState.tables.page_index);
+      });
+
+      it('given ResourceType.dashboard, returns page_index for dashboards', () => {
```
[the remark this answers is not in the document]
danwom: Forgot to update after copy-paste.

amundsen_application/static/js/ducks/search/sagas.ts
```diff
@@ -53,7 +73,78 @@ export function* searchResourceWorker(action: SearchResourceRequest): SagaIterat
...
+  yield takeEvery(SubmitSearch.REQUEST, submitSearchWorker);
+};
+
+export function* setResourceWorker(action: SetResourceRequest): SagaIterator {
+  const { resource, updateUrl } = action.payload;
+  const state = yield select(getSearchState);
+  if (updateUrl) {
+    updateSearchUrl({
+      resource,
+      term: state.search_term,
```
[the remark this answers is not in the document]
danwom: If you're on a `table_detail` page and search for `example`, you will be redirected to `/search?term=example` with a loading indicator (`submitSearch` action). At the end of the `searchAll` action, the URL will be updated with an appropriate `resource` and `index`. If we did not update the URL in `submitSearch`, the user would be staring at the `table_detail` page with no indication that the search was initiated until it finishes.

We _could_ just leave the URL alone without the `resource` and `index` until the user explicitly selects one, but I think keeping the URL provides more consistency.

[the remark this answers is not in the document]
ttannis: I see, so the this is accounting for our new work as well where search can happen on every page. Looks good!

PR #266, opened 2019-08-16
Creates User Preferences page with no functionality

ryanlieu:
### Summary of Changes

### Tests

### Documentation

### CheckList


ttannis:
@ryanlieu can you clean up the history of this one for review, looks like it was branched off a previous PR.

ryanlieu:
@ttannis @danwom good to look at sans tests

amundsen_application/static/js/index.tsx
```diff
@@ -43,6 +44,7 @@ ReactDOM.render(
             <Route path="/browse" component={BrowsePage} />
             <Route path="/search" component={SearchPage} />
             <Route path="/user/:userId" component={ProfilePage} />
+            <Route path="/preferences" component={PreferencesPage} />
```
ttannis: At the moment since we might not actually get this working end-to-end this week, can you add a `TODO: Remove PreferencesPage before merging to master` in this file so that we don't forget? The reason is that we wouldn't want to expose a page that won't do anything yet.

amundsen_application/static/js/components/PreferencesPage/PreferenceGroup/index.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import { bindActionCreators } from 'redux';
+import { connect } from 'react-redux';
+
+export type PreferenceGroupProps = {
+  changePreference: any;
```
ttannis: nit: Prefer to call this `onChange` or `onClick`. There are places where I know we aren't particularly following this, but if we know a method is executed on a "click", "submit", "hover", etc new folks will have an easier time if we start to name these methods base on how it will be used instead of tying the name to an understanding of the component.

not a nit: Let's type this more appropriately as `(value: string) => any`. Right now this allows us to not even pass a method, and we could set `changePreference: 1000` and break the component.

amundsen_application/static/js/components/PreferencesPage/PreferenceGroup/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+      subtitle: '',
+    };
+
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    return (
+      <label className="preference-group" onClick={() => this.props.changePreference(this.props.title)}>
```
ttannis: This is a bit uncertain to review because I recognize we're sort of just making some easily changed assumptions about the way the Redux support will be set up, but at minimum I think we should be working with some sort of `id`. 

The reason is that title can be changed any time, but the Redux layer and the backend support will need some way of knowing what option this is regardless of what the UI is displaying. So we need some `id` value, and then we can more appropriately call `this.props.changePreference(this.props.id)`

amundsen_application/static/js/components/PreferencesPage/PreferenceGroup/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+      title: '',
+      subtitle: '',
+    };
+
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    return (
```
ttannis: Can you add a `TODO: Consolidate with future common RadioButton component`. 

Context: Our new designs involve more places where users will have radio buttons and/or checkboxes, so as a result we will create some re-useable components because there is no reason for us to have to add all this markup each time we want a radio item. Ideally once we consolidate all this work over the next couple weeks, this can get replaced with a `<RadioButton>` that takes some props and children.

amundsen_application/static/js/components/PreferencesPage/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export type PreferencesPageProps = DispatchFromProps;
+
+export class PreferencesPage extends React.Component<PreferencesPageProps, PreferencesPageState> {
+  constructor(props) {
+    super(props);
+    this.changePreference = this.changePreference.bind(this);
+
+    this.state = {
+      selectedPreference: ALL_NOTIFICATIONS_TITLE,
```
ttannis: We're going to want to be working with some sort of fixed `id` here because the constants can change and are only relevant to what is getting displayed in the UI. It shouldn't be the value used for executing actual functionality because it can change any time. If we go this route we will also need some helper methods for getting the title and subtitle that takes the `id` and returns which constants to use

ryanlieu:
@ttannis good to look at, addressed comments

PR #283, opened 2019-09-04
Update AxiosError handling

ttannis:
### Summary of Changes

We've had issues with handling AxiosErrors in the case of a failed data preview, and our current implementation overlooked the fact that for some error codes `error.response.data` is not what we expected. In our python code we always return some json data on both success and failure as seen [here](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/api/preview/v0.py). 

Howevers the way `axios` chooses to handle some errors -- like a `504` -- the response data ends up being a string instead of the object we intended. (mistake here was writing code based on our knowledge of what we thought would happen) 

The fix I've added does some checks based on the documented [axios types](https://github.com/axios/axios/blob/master/index.d.ts). Processing of an `AxiosError` must account for:
1. `e.response` can be undefined <-- this is what the last fix accounted for
2. `e.response.data` can be `any` type <-- this was the oversight

I believe this should close the book on the kinds of problems we've had with this code.

### Tests

N/A. We need to get around to covering `tableMetadata/api/v0.ts`. I'll try to get around to it after our current H2 plans are maybe 80% done.

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
+++ b/amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
@@ -144,8 +144,11 @@ export function getPreviewData(queryParams: PreviewQueryParams) {
     return { data: response.data.previewData, status: response.status };
   })
   .catch((e: AxiosError<PreviewDataAPI>) => {
-    const data = e.response ? e.response.data.previewData : {};
-    const status = e.response ? e.response.status : null;
+    const response = e.response;
+    let data = {};
+    if (response && response.data && response.data.previewData)
+      data = response.data.previewData;
+    const status = response ? response.status : null;
     return Promise.reject({ data, status });
   });
 }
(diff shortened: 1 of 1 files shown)


amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
```diff
@@ -144,8 +144,12 @@ export function getPreviewData(queryParams: PreviewQueryParams) {
     return { data: response.data.previewData, status: response.status };
   })
   .catch((e: AxiosError<PreviewDataAPI>) => {
-    const data = e.response ? e.response.data.previewData : {};
-    const status = e.response ? e.response.status : null;
+    const response = e.response;
+    let data = {};
+    if (response && response.data && response.data.previewData) {
```
danwom: Why we need optional chaining https://github.com/tc39/proposal-optional-chaining

PR #293, opened 2019-09-07
Remove render_template for feedback email

ttannis:
### Summary of Changes

At Lyft we've experienced the feedback tool failing intermittently. After some time we tracked down that when the feature would fail it would always fail at the `render_template()` call. Our host running the application was then stuck restarting a gunicorn worker.

We have still not figured out why this happened, but we are moving away from using `render_template` for our custom email features. Luckily the html we are generating is simple enough that we can easily achieve the same result with a formatted string.

### Tests

N/A. Functionality has not changed.

### Documentation

N/A. Functionality has not changed.

### CheckList


ttannis:
@danwom 

The same thing will be done for notifications, we can create what we need with a formatted string. I still don't understand why `render_template()` works fine for the use case of `index.html`, but moving forward I guess let's just never use it anywhere else?

PR #295, opened 2019-09-09
Create Table Detail Components for V2

danwom:
### Summary of Changes
Created components for each section of the TableDetails page and placed them into the v2 layout.
- ExploreButton, LineageLink, WriterLink, SourceLink, FrequentUsers
- DataPreviewButton: added data fetch, use `props` instead of `state`

### Tests
Will add tests to new components in follow up

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/static/css/_layouts.scss
+++ b/amundsen_application/static/css/_layouts.scss
@@ -1,46 +1,51 @@
 @import 'variables';
-// TODO - cleanup temp background colors
 // TODO - Remove once merged with variable updates
 $divider: #CACAD9;
+$stroke: #CACAD9;
 $header-height: 108px;
 
 .resource-detail-layout {
   min-width: 1048px;
 
   .resource-header {
-    background: saddlebrown;
     display: flex;
     height: $header-height;
     border-bottom: 4px solid $divider;
     padding: 24px;
+
+    > .header-left {
+      flex-grow: 1;
+    }
+
+    > .header-right {
+    }
   }
 
 
   // 630 | 414+ Flexible layout
   .column-layout-1 {
-    background: red;
     display: flex;
     min-height: calc(100% - #{$header-height});
 
     > .left-panel {
       > .banner {
-        height: 100px;
-        background: pink;
+        border: 1px solid $stroke;
+        height: 40px;
+        margin: 24px;
+        padding: 8px;
       }
 
-      background: orange;
       border-right: 4px solid $divider;
       flex-basis: 634px; // 630 + 4px border
       flex-shrink: 0;
       padding: 0 24px;
-      flex-direction: column;
     }
 
     > .right-panel {
       flex-basis: 414px;
       flex-grow: 1;
       flex-shrink: 0;
-      background: blue;
+      width: 0; // Required for text truncation
     }
   }
 
@@ -50,13 +55,12 @@ $header-height: 108px;
     flex-basis: 100%;
 
     > .left-panel {
-      background: yellow;
       flex-basis: 313px;
       margin-right: 24px;
+      flex-direction: column;
     }
 
     > .right-panel {
-      background: green;
       flex-basis: 245px;
     }
   }
@@ -64,5 +68,6 @@ $header-height: 108px;
   .left-panel,
   .right-panel {
     display: flex;
+    flex-direction: column;
   }
 }
--- a/amundsen_application/static/css/_layouts.scss
+++ b/amundsen_application/static/css/_layouts.scss
@@ -31,7 +31,7 @@ $header-height: 108px;
       > .banner {
         border: 1px solid $stroke;
         height: 40px;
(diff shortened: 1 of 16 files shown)


danwom:
@ttannis this should be ready to review. I will wait to write tests until after we update each component to its final design/functionality.

amundsen_application/static/js/components/TableDetail/ExploreButton/index.tsx
```diff
@@ -0,0 +1,58 @@
+import * as React from 'react';
```
danwom: I connected some of the components and not others. This decision seems somewhat arbitrary. `DataPreviewButton` makes sense to connect since it needs to fetch data, but others like `LineageLink` and `ExploreButton` could go either way. On one hand you need to write more boilerplate code, on the other hand you might need pass in a bunch of props (or pass in the entire table metadata object as a prop).

Any thoughts on which way we should go?

ttannis: I prefer passing the props to child components. 

One way to think of it is that `DataPreviewButton` is calling another endpoint and nothing else in `TableDetail` needs to know about the results of that request, so it makes sense for that to be a connected component. 

However the others will be using data that `TableDetail` will already obtain, and are currently only used on `TableDetail`, so I think it becomes preferable to pass them down.

amundsen_application/static/js/components/common/AvatarLabel/index.tsx
```diff
@@ -11,11 +11,9 @@ export interface AvatarLabelProps {
 
 const AvatarLabel: React.SFC<AvatarLabelProps> = ({ label, src }) => {
   return (
-    <div className='avatar-label-component'>
-      <div id='component-avatar' className='component-avatar'>
-        <Avatar name={label} src={src} size={24} round={true} />
-      </div>
-      <label id='component-label' className='component-label'>{label}</label>
+    <div className="avatar-label-component">
```
ttannis: Can we revisit some of the style changes? The `AvatarLabels` as they present now have text that is not as well vertically centered as they used to be.

amundsen_application/static/js/components/TableDetail/DataPreviewButton/index.tsx
```diff
@@ -213,32 +216,29 @@ export class DataPreviewButton extends React.Component<DataPreviewButtonProps, D
...
   render() {
     // else render button that triggers the preview data modal
     return (
-      <div className="preview-data">
-        {this.renderPreviewButton()}
-
-        <Modal show={this.state.showModal} onHide={this.handleClose}>
-          <Modal.Header className="text-center" closeButton={true}>
+      <>
+        { this.renderPreviewButton() }
```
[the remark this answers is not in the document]
danwom: Looks like the hover/focus/click events don't fire on disabled elements. Looks like that wrapper element was needed after all.

amundsen_application/static/css/_layouts.scss
```diff
@@ -1,46 +1,54 @@
...
-    padding: 24px;
+    padding: 16px 24px;
+
+    > .header-left {
+      flex-grow: 1;
+    }
+
+    > .header-right {
+      > :not(:first-child) {
+        margin-left: 30px;
```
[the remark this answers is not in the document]
danwom: I noticed that as well but didn't spend the time to polish in this PR. I figure we can fine-tune after the components are in place.

amundsen_application/static/js/components/TableDetail/FrequentUsers/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  readers: TableReader[];
+}
+
+// TODO - Change SFC to FC when upgrading React
+const FrequentUsers: React.SFC<FrequentUsersProps> = ({ readers }) => {
+  if (readers.length === 0) {
+    return (<label className="body-3">No frequent users exist</label>);
+  }
+
+  const renderReader = (reader: TableReader, index: number) => {
```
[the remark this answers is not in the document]
danwom: Yeah, that is one of the downsides to using an SFC. There seems to be category of components that don't have state, but still have some non-trivial logic. I think the `Flag` pattern works well here, although I could see the argument for a more complex case to become a `React.Component` even if it has no state.

amundsen_application/static/js/components/common/AvatarLabel/index.tsx
danwom: Fixed. This was one of the few times where `vertical-align:middle` actually worked!

amundsen_application/static/js/components/TableDetail/ExploreButton/index.tsx
danwom: Disconnected all subcomponents except for `DataPreviewButton`

PR #297, opened 2019-09-12
Truncate the column type

ttannis:
### Summary of Changes

This PR truncated the column type for strings that start with `array`, `struct`, or `map` which have extended information. e.g. if the type is `array`, the text `array` will be rendered. If the type is `array...`, the text will be rendered as `array<...>` and on click will show the full text.

Kapture 2019-09-16 at 10 16 33

Implementation wise I considered creating a set of configurable "formatters" as we have these new use case to change how the column type is rendered (different folks might want different things) as well as numerical values (different folks will need different things due to internationalization). However given that a new design will be worked on for the concept of nested column types and this fix is a just a patch, I held off on introducing any new configurations until we understand the final requirements.

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/TableDetail/DetailList/DetailListItem/index.tsx
+++ b/amundsen_application/static/js/components/TableDetail/DetailList/DetailListItem/index.tsx
@@ -59,7 +59,7 @@ class DetailListItem extends React.Component<DetailListItemProps, DetailListItem
     let text = fullText;
 
     truncatedTypes.forEach((truncatedType) => {
-      if (type.startsWith(truncatedType)) {
+      if (type.startsWith(truncatedType) && type !== truncatedType) {
         shouldTrucate = true;
         text = `${truncatedType}<...>`;
         return;
@@ -80,7 +80,7 @@ class DetailListItem extends React.Component<DetailListItemProps, DetailListItem
           trigger={['click']}
           placement='right'
           overlay={popoverHover}
-          rootClose>
+          rootClose={true}>
             <a className='column-type'
                href="JavaScript:void(0)"
                onClick={ stopPropagation }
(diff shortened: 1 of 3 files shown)


amundsen_application/static/js/components/TableDetail/DetailList/DetailListItem/index.tsx
```diff
@@ -75,7 +75,7 @@ class DetailListItem extends React.Component<DetailListItemProps, DetailListItem
           <div className='title-section'>
             <div className='title-row'>
               <div className='name title-2'>{metadata.name}</div>
-              <div className='column-type'>{metadata.type ? metadata.type.toLowerCase() : 'null'}</div>
+              <div className='column-type truncated'>{metadata.type ? metadata.type.toLowerCase() : 'null'}</div>
```
feng-tao: nice! for column-type truncated, what is the max length of the type before it will get truncated?

markgrover:
Hey @ttannis how does the user get the full type? Hover? Or is it in the description? Is it copy-able so they can paste it elsewhere and see?

@devtagare also had a good thought about actually allowing users to browse through this via clicking and I wonder if it'd be possible to have such an experience?

Apologies this is on me for not sharing my requirements clearly.

ttannis:
@danwom let me know if you have any concerns

amundsen_application/static/js/components/TableDetail/DetailList/DetailListItem/index.tsx
```diff
@@ -49,6 +51,48 @@ class DetailListItem extends React.Component<DetailListItemProps, DetailListItem
...
+      );
+      const stopPropagation = (event) => {
+        event.stopPropagation();
+      }
+      return (
+        <OverlayTrigger
+          trigger={['click']}
+          placement='right'
+          overlay={popoverHover}
+          rootClose={true}>
```
danwom: Can we try using `delayHide` for this element? I think when users try to highlight they might move their mouse outside the hoverable zone which could hide the tooltip. I'm not sure if the hide event will get debounced/cancelled if the user moves the mouse back into the hoverable zone, but it's worth investigating.

danwom: Oh this is click only. I see.

PR #298, opened 2019-09-19
Improve table & columns description formatting (#98)

Mikhail-Ivanov:
### Summary of Changes

Introduced support of Markdown descriptions for tables & columns.
This module changes are intended to render Markdown and to autosize HTMLTextAreaElement on edit (to fit content).

Also dependencies have been updated and build / dev-build scripts have been modified to support run on alternative environments (for instance Win10).

[Related changes at amundsenmetadatalibrary.](https://github.com/lyft/amundsenmetadatalibrary/pull/79)

### Tests

EditableText element tests have been modified to match changes (nested `<ReactMarkdown/>` tag for table and column descriptions).

### Documentation

Markdown description format demo:
https://rexxars.github.io/react-markdown/

### DIFF
--- a/amundsen_application/api/utils/request_utils.py
+++ b/amundsen_application/api/utils/request_utils.py
@@ -15,13 +15,15 @@ def get_query_param(args: Dict, param: str, error_msg: str = None) -> str:
 def request_metadata(*,     # type: ignore
                      url: str,
                      method: str = 'GET',
-                     timeout_sec: int = 0):
+                     timeout_sec: int = 0,
+                     json: str = ''):
     """
     Helper function to make a request to metadata service.
     Sets the client and header information based on the configuration
     :param method: DELETE | GET | POST | PUT
     :param url: The request URL
     :param timeout_sec: Number of seconds before timeout is triggered.
+    :param json: Optional request payload
     :return:
     """
     if app.config['REQUEST_HEADERS_METHOD']:
@@ -32,19 +34,22 @@ def request_metadata(*,     # type: ignore
                            url=url,
                            client=app.config['METADATASERVICE_REQUEST_CLIENT'],
                            headers=headers,
-                           timeout_sec=timeout_sec)
+                           timeout_sec=timeout_sec,
+                           json=json)
 
 
 def request_search(*,     # type: ignore
                    url: str,
                    method: str = 'GET',
-                   timeout_sec: int = 0):
+                   timeout_sec: int = 0,
+                   json: str = ''):
     """
     Helper function to make a request to search service.
     Sets the client and header information based on the configuration
     :param method: DELETE | GET | POST | PUT
     :param url: The request URL
     :param timeout_sec: Number of seconds before timeout is triggered.
+    :param json: Optional request payload
     :return:
     """
     if app.config['REQUEST_HEADERS_METHOD']:
@@ -55,11 +60,12 @@ def request_search(*,     # type: ignore
                            url=url,
                            client=app.config['SEARCHSERVICE_REQUEST_CLIENT'],
                            headers=headers,
-                           timeout_sec=timeout_sec)
+                           timeout_sec=timeout_sec,
+                           json=json)
 
 
 # TODO: Define an interface for envoy_client
-def request_wrapper(method: str, url: str, client, headers, timeout_sec: int):  # type: ignore
+def request_wrapper(method: str, url: str, client, headers, timeout_sec: int, json: str = ''):  # type: ignore
     """
     Wraps a request to use Envoy client and headers, if available
     :param method: DELETE | GET | POST | PUT
@@ -67,6 +73,7 @@ def request_wrapper(method: str, url: str, client, headers, timeout_sec: int):  
     :param client: Optional Envoy client
     :param headers: Optional Envoy request headers
     :param timeout_sec: Number of seconds before timeout is triggered. Not used with Envoy
+    :param json: Optional request payload
     :return:
     """
     # If no timeout specified, use the one from the configurations.
@@ -78,9 +85,9 @@ def request_wrapper(method: str, url: str, client, headers, timeout_sec: int):  
         elif method == 'GET':
             return client.get(url, headers=headers, raw_response=True)
         elif method == 'POST':
-            return client.post(url, headers=headers, raw_response=True)
+            return client.post(url, headers=headers, raw_response=True, json=json)
         elif method == 'PUT':
-            return client.put(url, headers=headers, raw_response=True)
+            return client.put(url, headers=headers, raw_response=True, json=json)
         else:
             raise Exception('Method not allowed: {}'.format(method))
     else:
@@ -90,8 +97,8 @@ def request_wrapper(method: str, url: str, client, headers, timeout_sec: int):  
             elif method == 'GET':
                 return s.get(url, headers=headers, timeout=timeout_sec)
             elif method == 'POST':
-                return s.post(url, headers=headers, timeout=timeout_sec)
+                return s.post(url, headers=headers, timeout=timeout_sec, json=json)
             elif method == 'PUT':
-                return s.put(url, headers=headers, timeout=timeout_sec)
+                return s.put(url, headers=headers, timeout=timeout_sec, json=json)
             else:
                 raise Exception('Method not allowed: {}'.format(method))
--- a/amundsen_application/api/utils/request_utils.py
+++ b/amundsen_application/api/utils/request_utils.py
@@ -16,7 +16,7 @@ def request_metadata(*,     # type: ignore
                      url: str,
                      method: str = 'GET',
                      timeout_sec: int = 0,
-                     json: str = ''):
+                     json: str = '{}'):
     """
     Helper function to make a request to metadata service.
     Sets the client and header information based on the configuration
@@ -42,7 +42,7 @@ def request_search(*,     # type: ignore
                    url: str,
                    method: str = 'GET',
                    timeout_sec: int = 0,
-                   json: str = ''):
+                   json: str = '{}'):
     """
     Helper function to make a request to search service.
     Sets the client and header information based on the configuration
@@ -65,7 +65,7 @@ def request_search(*,     # type: ignore
 
 
 # TODO: Define an interface for envoy_client
-def request_wrapper(method: str, url: str, client, headers, timeout_sec: int, json: str = ''):  # type: ignore
+def request_wrapper(method: str, url: str, client, headers, timeout_sec: int, json: str = '{}'):  # type: ignore
     """
     Wraps a request to use Envoy client and headers, if available
     :param method: DELETE | GET | POST | PUT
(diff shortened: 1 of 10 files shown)


feng-tao:
thanks @Mikhail-Ivanov, I will defer it to our FE expert to review. In the mean time, could you also fix the CLA check? Normally it is due to you didn't configure your github handle(username, email) in your local box.

amundsen_application/static/js/components/common/EditableText/index.tsx
```diff
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import ReactDOM from 'react-dom';
+import ReactMarkdown from 'react-markdown/with-html';
```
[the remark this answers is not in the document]
Mikhail-Ivanov: Yes, you're right: https://github.com/rexxars/react-markdown#parsing-html

amundsen_application/static/js/components/common/EditableText/index.tsx
```diff
@@ -113,7 +115,7 @@ class EditableText extends React.Component<EditableTextProps, EditableTextState>
     if (!this.state.editable) {
       return (
         <div id='editable-container' className='editable-container'>
-           <div id='editable-text' className='editable-text'>{ this.state.value }</div>
+           <div id='editable-text' className='editable-text'>{<ReactMarkdown source={this.state.value}/>}</div>
```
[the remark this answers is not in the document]
Mikhail-Ivanov: As for <p> tag I do not observe such a behavior now. Probably ReactMarkdown tag isn't affected with that issue?

amundsen_application/static/package.json
```diff
@@ -8,8 +8,8 @@
     "url": "https://github.com/lyft/amundsenfrontendlibrary"
   },
   "scripts": {
-    "build": "TS_NODE_PROJECT='tsconfig-for-webpack.json' webpack -p --progress --config webpack.prod.ts",
-    "dev-build": "TS_NODE_PROJECT='tsconfig-for-webpack.json' webpack -d --progress --config webpack.dev.ts",
+    "build": "cross-env TS_NODE_PROJECT='tsconfig-for-webpack.json' webpack -p --progress --config webpack.prod.ts",
+    "dev-build": "cross-env TS_NODE_PROJECT='tsconfig-for-webpack.json' webpack -d --progress --config webpack.dev.ts",
```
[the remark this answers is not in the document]
Mikhail-Ivanov: Actually no - I've checked them as well and did the same change where it needed

amundsen_application/static/js/components/common/EditableText/tests/index.spec.tsx
```diff
@@ -27,15 +27,17 @@ describe('EditableText', () => {
             props.editable = false;
              /* Note: Do not copy this pattern, for some reason setProps is not updating the content in this case */
             subject = shallow(<EditableText {...props} />);
-            expect(subject.find('div#editable-text').text()).toEqual(props.value);
+            /* TODO fix it - nested ReactMarkdown tag appears to be UNDEFINED and broke the test */
+            /* expect(subject.find('div#editable-text').text()).toEqual(props.value); */
```
danwom: I believe this is because we are using `shallow` render, which does not initialize other sub-components so we aren't testing the behavior of the sub-component (in this case `ReactMarkdown`). Since the text isn't rendered, we can simply assert that there should exist a `<ReactMarkdown>` element with right value passed into the `source` prop.

amundsen_application/static/js/components/common/EditableText/index.tsx
```diff
@@ -1,8 +1,11 @@
 import * as React from 'react';
 import ReactDOM from 'react-dom';
+import ReactMarkdown from 'react-markdown';
```
danwom: And both `import * as ReactMarkdown from 'react-markdown';` and `import { ReactMarkdown } from 'react-markdown';` fail as well without this option? It's very odd that such a popular library wouldn't be compatible with ES6 imports without that option.

amundsen_application/api/utils/request_utils.py
```diff
@@ -55,18 +60,20 @@ def request_search(*,     # type: ignore
...
                            client=app.config['SEARCHSERVICE_REQUEST_CLIENT'],
                            headers=headers,
-                           timeout_sec=timeout_sec)
+                           timeout_sec=timeout_sec,
+                           json=json)
 
 
 # TODO: Define an interface for envoy_client
-def request_wrapper(method: str, url: str, client, headers, timeout_sec: int):  # type: ignore
+def request_wrapper(method: str, url: str, client, headers, timeout_sec: int, json: str = ''):  # type: ignore
```
danwom: I'm not super familiar with python, but would there be any unintended consequences from using an empty string as a default value? @feng-tao

amundsen_application/static/js/components/common/EditableText/index.tsx
Mikhail-Ivanov: First one helped, rolling back tsconfig changes

amundsen_application/static/js/components/common/EditableText/tests/index.spec.tsx
Mikhail-Ivanov: It turned out that issue (preventing me of extracting nested elements properties) was induced by improper ReactMarkdown import

Mikhail-Ivanov:
Travis CI build fails, looks like infrastructure issues: "Timeout waiting for network availability."

amundsen_application/api/utils/request_utils.py
ttannis: From the metadata PR, this value gets used by `json.loads()`. @Mikhail-Ivanov what is the result of using `json.loads()` on an empty string? 

Tried looking this up myself but I couldn't quite find some clear documentation on what the default behavior is.

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -354,7 +354,7 @@ export class TableDetail extends React.Component<TableDetailProps & RouteCompone
                   !data.is_view && <WatermarkLabel watermarks={ data.watermarks }/>
                 }
                 <TableDescEditableText
-                  maxLength={ 750 }
+                  maxLength={ 4000 }
```
ttannis: @verdan, just checking -- does Atlas have any limitation in regards to the character length it can store for the table and column descriptions?

@Mikhail-Ivanov I also think since the change to support markdown is on `EditableText` we should increase the default `maxLength` on `EditableText` instead and then remove this override here. Then we'll have the full benefit of this work on `ColumnDescEditableText` as well. Our users at Lyft have requested a larger character limit for both table and column descriptions.

amundsen_application/api/utils/request_utils.py
feng-tao: I think empty string is fine and keep it as it is will help to make it backward compatible.

ttannis: How so? 

As mentioned, I see the[ empty string gets passed](https://github.com/lyft/amundsenmetadatalibrary/pull/79/files#diff-4b1a079f306204ef1278fb5acc3a4a23R30) to `json.loads()` and is not by itself the default description. I think we need to confirm what `json.loads("")` returns -- is it `None`, `{}`, or some error?

Mikhail-Ivanov: @ttannis, @feng-tao here are my findings:
https://github.com/lyft/amundsenfrontendlibrary/pull/298#discussion_r327478289
https://github.com/lyft/amundsenmetadatalibrary/pull/79#discussion_r327473924
Looks like the only way to truncate description is to intentionally send an empty "description" key

ttannis: From your comment [here](https://github.com/lyft/amundsenfrontendlibrary/pull/298#discussion_r327478289) it looks like `json.loads('')` threw an error `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)` whereas `json.loads('{}')` resulted in an empty dictionary `{}`. 

Considering `''` vs. `'{}'` as the default encoded json string, while they both result in the description not being updated -- which is what we want -- I am leaning towards `'{}'` as the default encoded string. I think it is more correct for the default to resolve to an empty dictionary as opposed to choosing a default that is guaranteed to throw an error if it gets used.

amundsen_application/static/js/components/TableDetail/index.tsx
ttannis: The `maxLength` topic really does get into a subjective space -- I think there will always be users who want to write more, and likewise there might always be users who wish descriptions would have more brevity. I am okay with leaving this as `4000` for now unless this will cause any problems for Atlas.

Perhaps @dwon or I can think a bit more about what is optimal after this is merged. Maybe we will end up limiting the `maxLength` based on it's affect on the UI. For example, there is likely a point where the UI becomes harder to use if the descriptions are too big.

ttannis:
Thanks for the screenshots! I'm concerned about how this works for the column descriptions because we have some code that is meant to truncate the column description if the column is expandable -- but I don't think our default setup sets you up properly to test that case. We might have to tweak that ourselves.

amundsen_application/api/utils/request_utils.py
feng-tao: @ttannis  are we ok to use mutable as default argument given it is a bad practice in python world(https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)?

ttannis: This one isn't mutable because the default is not an empty dictionary itself, but an encoded json string that will resolve to a new empty dictionary each time. 

Here's some code I just tried:
```
import json

# when the default is an encoded json string that has to be parsed by json.loads()
def this_is_safe(key, value, j = '{}'):
    test_dict = json.loads(j)
    test_dict[key] = value
    return test_dict

print(this_is_safe('test', 'testval'))     # {'test': 'testval'}
print(this_is_safe('test2', 'testval2'))   # {'test2': 'testval2'}


# when the default is an actual dictionary
def this_is_not_safe(key, value, j = {}):
    test_dict = j
    test_dict[key] = value
    return test_dict

print(this_is_not_safe('test', 'testval'))    # {'test': 'testval'}
print(this_is_not_safe('test2', 'testval2'))  # {'test': 'testval', 'test2': 'testval2'}
```

feng-tao: oh, initially I thought the default is `{}` not `'{}'`, if it is former case, yeah, it won't be mutable per se.

feng-tao:
@ttannis @danwom any final comment before we merge :) ?

amundsen_application/api/utils/request_utils.py
Mikhail-Ivanov: @ttannis 
> I am leaning towards `'{}'` as the default encoded string
Agree, it's a better option. Done

amundsen_application/static/js/components/TableDetail/index.tsx
danwom: You're right, these limits are completely subjective. One way around this would be to define default values (say 2000?) and allow the config files to override. Also worth noting is that table descriptions might need more characters than a column description. I wouldn't block this PR on configuring these values, but it is a nice-to-have feature.

danwom:
There are a few edits I would like to make, but I'm not sure how to push to your forked branch so I will merge this and then merge another fix on top of it.

PR #300, opened 2019-09-26
optimize docker

jornh:
### Summary of Changes

Optimize docker - see https://github.com/lyft/amundsen/issues/109

You may wonder why the renaming in `amundsen_application/wsgi.py` from `app` to `application` is done. It turned out for `Gunicorn` to run it was needed otherwise it spat this out in the log:

```
amundsenfrontend    | 2019-09-26T18:48:23+0000.384 [INFO] __init__.create_app:47 (8:MainThread) - Using metadata service at http://amundsenmetadata:5000                                             
amundsenfrontend    | 2019-09-26T18:48:23+0000.384 [INFO] __init__.create_app:48 (8:MainThread) - Using search service at http://amundsensearch:5000                                                 
amundsenfrontend    | Traceback (most recent call last):
amundsenfrontend    |   File "/usr/local/lib/python3.7/site-packages/gunicorn/util.py", line 362, in import_app                                                                                      
amundsenfrontend    |     app = eval(obj, vars(mod))                                                                                                                                                 
amundsenfrontend    |   File "<string>", line 1, in <module>                                                                                                                                         
amundsenfrontend    | NameError: name 'application' is not defined                                                                                                                                   
amundsenfrontend    | Failed to find application object 'application' in 'amundsen_application.wsgi'                                                                                                 
amundsenfrontend    | [2019-09-26 18:48:23 +0000] [8] [INFO] Worker exiting (pid: 8)                                                                                                                 
amundsenfrontend    | [2019-09-26 18:48:23 +0000] [1] [INFO] Shutting down: Master
amundsenfrontend    | [2019-09-26 18:48:23 +0000] [1] [INFO] Reason: App failed to load. 
```

### Tests

Manually tested locally with:
* `docker build -f public.Dockerfile -t amundsendev/amundsen-frontend:1.1.0 .` for `public.Dockerfile` Quickstart "production build"
* and the `docker-compose.exe -f docker-amundsen-local.yml build` build for "dev latest" build

### Documentation

Should be on par with the Search service now

### CheckList
- [x] PR includes a summary of changes. 
    - All the public functions and the classes in the PR contain docstrings that explain what it does
- [x] PR passes `make test`

### DIFF
--- a/amundsen_application/wsgi.py
+++ b/amundsen_application/wsgi.py
@@ -1,9 +1,9 @@
 import os
 from amundsen_application import create_app
 
-app = create_app(
+application = create_app(
     config_module_class=os.getenv('FRONTEND_SVC_CONFIG_MODULE_CLASS') or
     'amundsen_application.config.LocalConfig')
 
 if __name__ == '__main__':
-    app.run(host='0.0.0.0')
+    application.run(host='0.0.0.0')
--- a/local.Dockerfile
+++ b/local.Dockerfile
@@ -1,23 +1,27 @@
 ARG METADATASERVICE_BASE
 ARG SEARCHSERVICE_BASE
 
-FROM node:8 as node-stage
-
-COPY . /app
-
+FROM node:8-slim as node-stage
 WORKDIR /app/amundsen_application/static
 
+COPY amundsen_application/static/package.json /app/amundsen_application/static/package.json
+COPY amundsen_application/static/package-lock.json /app/amundsen_application/static/package-lock.json
 RUN npm install
+
+COPY amundsen_application/static/ /app/amundsen_application/static/
 RUN npm rebuild node-sass
 RUN npm run dev-build
 
-FROM python:3
+COPY . /app
+
+FROM python:3-slim
+WORKDIR /app
+
+COPY requirements3.txt /app/requirements3.txt
+RUN pip3 install -r requirements3.txt
 
 COPY --from=node-stage /app /app
 
-WORKDIR /app
-
-RUN pip3 install -r requirements3.txt
 RUN python3 setup.py install
 
 ENTRYPOINT [ "python3" ]
--- a/public.Dockerfile
+++ b/public.Dockerfile
@@ -1,20 +1,22 @@
-FROM node:8 as node-stage
-
-COPY . /app
-
+FROM node:8-slim as node-stage
 WORKDIR /app/amundsen_application/static
 
+COPY amundsen_application/static/package.json /app/amundsen_application/static/package.json
+COPY amundsen_application/static/package-lock.json /app/amundsen_application/static/package-lock.json
 RUN npm install
+
+COPY amundsen_application/static /app/amundsen_application/static
 RUN npm run build
 
-FROM python:3
+FROM python:3-slim
+WORKDIR /app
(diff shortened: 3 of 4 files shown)


ttannis:
Thanks for the PR  @jornh, will take . look after our notifications branch is merged and make sure this gets included in the 1.1.0 release.

jornh:
No worries!

Great to see notifications merged as well 👏 and looking forward to kick the tires of it! Being able to request description improvements  like in  https://github.com/lyft/amundsenfrontendlibrary/blob/523826c0e9996fa4db0ec583a1c0a26c9d7873dc/amundsen_application/api/utils/notification_utils.py#L76-L92 sounds brilliantly simple. Will be really interesting to see if users think the same and it will fly 🙂

jornh:
I just pushed https://github.com/lyft/amundsen/pull/123 which I hope should be useful in verifying this PR

amundsen_application/wsgi.py
```diff
@@ -1,9 +1,9 @@
 import os
 from amundsen_application import create_app
 
-app = create_app(
+application = create_app(
```
ttannis: In what situation did gunicorn output `NameError: name 'application' is not defined`? I understand you experienced some error and that this fixed it, but I'm not understanding why that error occurred.

local.Dockerfile
```diff
@@ -1,23 +1,27 @@
...
-COPY . /app
-
+FROM node:8-slim as node-stage
 WORKDIR /app/amundsen_application/static
 
+COPY amundsen_application/static/package.json /app/amundsen_application/static/package.json
+COPY amundsen_application/static/package-lock.json /app/amundsen_application/static/package-lock.json
 RUN npm install
+
+COPY amundsen_application/static/ /app/amundsen_application/static/
```
ttannis: Can you elaborate on how this speeds things up?

I understanding that copying `package.json` over will optimize `npm-install` when there are no dependency changes during development. However does copying over `static` have the same effect for npm run dev-build? I haven't seen any mention of this in the (limited) amount of time I've spent looking, and curious if  `COPY amundsen_application/static/ /app/amundsen_application/static/` really yields an improvement over having `COPY . /app` at this line

amundsen_application/wsgi.py
jornh: It happens on startup of the application. I didn’t explore further when I implemented, just followed instructions in Gunicorn’s error message. But I now see they default to looking for `”application”` in `import_app()` here: https://github.com/benoitc/gunicorn/blob/19.9.0/gunicorn/util.py#L362

I’m more inclined to just follow the simple default case if we don’t have anything else depending on it being called `app` and not the default (I was actually wondering why the frontend  name was different than the other Flask services). 

If needed it looks like they take a `:` separated module name parameter, so should be easy to have Gunicorn work with `app` though.

local.Dockerfile
jornh: Good question!

By limiting what we copy to just the typescript code folder we avoid to trigger npm builds when doing changes to just the Python code.

amundsen_application/wsgi.py
ttannis: Not sure I quite understand, but it's alright. I'll read up more on it.

PR #306, opened 2019-10-03
Modify Tags + EditableSection for Table V2

danwom:
### Summary of Changes
- Added an `EditableSection` component that manages the `isEditing` state of whatever child gets passed in. 
- Updated `TagInput` to new designs: 
-- Created muted `+ Add` button
-- "Escape" button will exit editing mode
-- Rely on `props` instead of copying to `state`

### Tests

### Documentation

### CheckList


amundsen_application/static/css/_buttons-default.scss
```diff
@@ -53,6 +60,16 @@
...
+    &.muted {
+      border-color: $divider;
+      color: $text-secondary;
+      padding: 0 8px;
+
+      .icon {
+        background-color: $text-secondary;
+      }
+    }
+
```
danwom: The designs added a new button type which is very similar to `btn-default` so `muted` is currently a modifier on `btn-default`.

amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const childrenWithProps = React.Children.map(this.props.children, child => {
+      if (React.isValidElement(child)) {
+        return React.cloneElement(child, {
+          isEditing: this.state.isEditing,
+          startEditing: this.startEditing,
+          stopEditing: this.stopEditing,
+        });
+      }
+      return child;
+    });
```
danwom: Approach taken from here: https://learn.co/lessons/react-this-props-children

amundsen_application/static/js/components/Tags/TagInput/styles.scss
```diff
@@ -1,101 +1,77 @@
...
-      background-color: $gray-lighter !important;
-      border-radius: 4px;
-
-      .amundsen__multi-value__label {
-        border-radius: 4px 0 0 4px;
-        color: $text-dark;
-        line-height: 14px;
-        padding: 8px;
+  */
+  .basic-multi-select {
```
danwom: I nested most of the custom `multi-select` styles under `.tag-input`

amundsen_application/static/js/components/common/EntityCard/EntityCardSection/styles.scss
```diff
@@ -19,42 +20,45 @@
...
-  -moz-box-shadow: inset 1px 1px 2px $brand-color-4;
-  -webkit-box-shadow: inset 1px 1px 2px $brand-color-4;
-  box-shadow: inset 1px 1px 2px $brand-color-4;
-}
-button.edit-button:focus,
-button.active-edit-button:focus {
-  outline: none;
-  box-shadow: 0 0 1pt 1pt $brand-color-4
-}
+  button.edit-button {
```
danwom: I nested these styles under `.entity-card-section` because they were overlapping with the new edit buttons. Eventually we should be able to delete this.

amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  startEditing = () => {
+    this.setState({ isEditing: true });
+  };
+
+  stopEditing = () => {
+    this.setState({ isEditing: false });
+  };
```
ttannis: Is there a benefit to managing all 2 different methods whose function can be also be achieved with:
```
  setEditMode = (isEditing: boolean) => {
    this.setState({ isEditing });
  };
```

I feel as if this is just as intuitive for the child components to call one method with `true` or `false`, and it is less code to pass around/implement/manage.

amundsen_application/static/css/_buttons-default.scss
danwom: The difference is pretty subtle, the padding is smaller and the text is lighter.

Screen Shot 2019-10-07 at 4 29 38 PM

amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx
danwom: Originally I tried passing in `toggleEdit = (isEditing: boolean = !this.state.isEditing)` but I couldn't use it as both an `onClick` and a prop to children. I guess I went too far the other direction.

I'll keep `toggleEdit` and `setEditMode` separately.

danwom:
@ttannis mind taking another look?

amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  toggleEdit = () => {
+    this.setState({ isEditing: !this.state.isEditing });
+  };
+
+  render() {
+    const childrenWithProps = React.Children.map(this.props.children, child => {
+      if (!React.isValidElement(child)) {
+        return child;
```
ttannis: nit: I think you could cover this case by using a setup with some plain html markup `const element = (<div>Hi</div>);` and then verifying that the render() result is that element? Not blocking merge on this though.

amundsen_application/static/js/components/Tags/TagInput/index.tsx
```diff
@@ -303,4 +305,4 @@ export const mapDispatchToProps = (dispatch: any) => {
   return bindActionCreators({ getAllTags, updateTags } , dispatch);
 };
 
-export default connect<StateFromProps, DispatchFromProps, ComponentProps>(mapStateToProps, mapDispatchToProps)(TagInput);
+export default connect<StateFromProps, DispatchFromProps>(mapStateToProps, mapDispatchToProps)(TagInput);
```
ttannis: Making a note to add tests for this component as part of the final review for the whole branch. Not blocking merge on this though since it's pre-existing tech-debt.

amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx
danwom: Added a case for when the child is just a `string`. Using `<div>` actually counts as a `React.isValidElement` and will attempt to put the props on the div, which gives further warnings.

ttannis: If that's the case, is `React.isValidElement` the right method to use then? Do we at least get those warnings when the code tries to build?

I'd want to make sure we avoid the possibility for someone to use `EditableSection` incorrectly, yet still end up with a successful build and deploy and then the page breaks in production.

danwom: I'll do some research on type-checking children and modify this as a follow up commit.

PR #309, opened 2019-10-09
Custom routes + Further notification cleanup

ttannis:
### Summary of Changes

This PR adds a new feature to allow the configuration of custom and completes some cleanup in the notifications code.

1. There is a new config variable `INIT_CUSTOM_ROUTES`. It is a `Callable` that takes a Flask application and is called during `create_app`. This allows individuals to configure custom endpoints to support custom logic. In our case at Lyft, we are using it to hook on a new endpoint to our frontend application that will process some internal callback to generate some custom metrics.

2. In our python code I have added an enum class `NotificationType`, that matches the `NotificationType` interface in the frontend application, and replaced all usages of hardcoded notification type strings to leverage the enum class. This makes it easier to maintain the code snd prevent some human error when it comes to writing our notification logic.

3. For notification emails, instead of passing "notification" as the email type, we will pass the exact `notification_type` instead. This is more helpful because it will allow custom mail clients to execute specific logic based on the `notification_type`.

4. ~~In `_build_metrics()` in `action_log.py` we now catch any exceptions that occur when getting the `user` for metrics, and fallback to the default. The reason is that actions that are logged in any of the custom routes, if that endpoint was hit by another service as opposed to being the result of a user action -- the `AUTH_USER_METHOD` might encounter some error.~~

### Tests

Added a python test to cover calling `get_notification_subject()` with an unsupported notification type.

### Documentation

Added documentation for `INIT_CUSTOM_ROUTES` to `flask_config.md`.

### CheckList

### DIFF
--- a/amundsen_application/__init__.py
+++ b/amundsen_application/__init__.py
@@ -56,4 +56,8 @@ def create_app(config_module_class: str, template_folder: str = None) -> Flask:
     app.register_blueprint(search_blueprint)
     init_routes(app)
 
+    init_custom_routes = app.config.get('INIT_CUSTOM_ROUTES')
+    if init_custom_routes:
+        init_custom_routes(app)
+
     return app
--- a/amundsen_application/api/utils/notification_utils.py
+++ b/amundsen_application/api/utils/notification_utils.py
@@ -1,6 +1,7 @@
 import logging
 
 from http import HTTPStatus
+from enum import Enum
 
 from flask import current_app as app
 from flask import jsonify, make_response, Response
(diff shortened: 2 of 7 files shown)


ttannis:
@jinhyukchang can you give your input on the changes in `amundsen_application/log/action_log.py`? 

Now that I think about it, we could also modify our our custom authentication method and return some default "service user" for our new use case where a service is hitting a custom endpoint as opposed to a real user. What do you think?

ttannis:
Reverted the change to `_build_metrics` in favor of a modification to our custom `AUTH_USER_METHOD`.

ttannis:
@danwom PTAL

ttannis:
> Looks fine on my end, but maybe you'll want to get @jinhyukchang to look as well?

Jin was fine with the changes I tagged him to look at in our private repo, will go ahead and merge this.

PR #313, opened 2019-10-16
Update searchv2 master

ttannis:
### Summary of Changes

This PR updates `feature/search_v2` with `master`, for preparation to merge `feature/search_v2` down.

### Tests

N/A

### Documentation

N/A

### CheckList


amundsen_application/static/js/components/common/FlashMessage/styles.scss
```diff
@@ -0,0 +1,20 @@
+@import 'variables';
+
+.flash-message {
+  background-color: #292936; // $gray100
```
danwom: Replace with `$body-bg-dark`

amundsen_application/static/js/config/config-types.ts
```diff
@@ -6,23 +6,27 @@
...
-  indexUsers: indexUsersConfig;
 }
 
 export interface AppConfigCustom {
   browse?: BrowseConfig;
+  editableText?: EditableTextConfig;
   google?: GoogleAnalyticsConfig
+  indexUsers?: IndexUsersConfig;
   logoPath?: string;
+  mailClientFeatures: MailClientFeaturesConfig;
```
danwom: This should be optional, otherwise an error will be thrown if your private config is not configured with this field. Eventually I think we could do something with `Partial<>` and optional `?` and smart object merging to make allow default values and custom values to work more seamlessly.

amundsen_application/static/js/config/config-utils.ts
```diff
@@ -30,3 +30,17 @@ export function getDatabaseIconClass(databaseId: string): string {
...
+ */
+export function feedbackEnabled(): boolean {
+  return AppConfig.mailClientFeatures.feedbackEnabled;
+}
+
+/**
+ * Returns whether or not notification features should be enabled
+ */
+export function notificationsEnabled(): boolean {
+  return AppConfig.mailClientFeatures.notificationsEnabled;
```
danwom: Just something to ponder:

As an alternative to exporting a function to check each feature, I wonder if we could make a generic function that says `checkFeature('notifications')` or something. The downside being that you'd need to make `'notifications'` a constant and export that as well.

amundsen_application/static/js/config/config-types.ts
ttannis: Yeah this is an oversight on master as well that would be undesirable in the notifications release. Let me change it there and pull the change in.

amundsen_application/static/js/config/config-utils.ts
ttannis: Yeah i think a better way will expose itself soon enough. But I'm unsure of what's best at this point given that our configs vary so much. A good majority of the helper methods will just return the config value itself, while others like database display name and icon class methods execute some extra logic as well to handle defaults. 

But yes ideally it would be great to just know to import a helper function that takes parameters as opposed to importing various ones.

PR #320, opened 2019-10-28
Make table owner items clickable

Mikhail-Ivanov:
### Summary of Changes

Making table Owner items clickable like table Frequent Users.
An item click leads to a redirect to a user's homepage.

### Tests

No new tests.

### Documentation

No new documentation.

closes lyft/amundsen#146

### DIFF
--- a/amundsen_application/static/js/components/TableDetail/OwnerEditor/index.tsx
+++ b/amundsen_application/static/js/components/TableDetail/OwnerEditor/index.tsx
@@ -5,6 +5,7 @@ import { bindActionCreators } from 'redux';
 import ReactDOM from 'react-dom';
 import serialize from 'form-serialize';
 
+import AppConfig from 'config/config';
 import AvatarLabel, { AvatarLabelProps } from 'components/common/AvatarLabel';
 import LoadingSpinner from 'components/common/LoadingSpinner';
 import { Modal } from 'react-bootstrap';
@@ -240,7 +241,13 @@ export class OwnerEditor extends React.Component<OwnerEditorProps, OwnerEditorSt
 export const mapStateToProps = (state: GlobalState) => {
   const ownerObj = state.tableMetadata.tableOwners.owners;
   const items = Object.keys(ownerObj).reduce((obj, ownerId) => {
-    obj[ownerId] = { label: ownerObj[ownerId].display_name }
+    let link = ownerObj[ownerId].profile_url;
+    let target = '_blank';
+    if (AppConfig.indexUsers.enabled) {
+      link = `/user/${ownerObj[ownerId].user_id}?source=owners`;
+      target = '';
+    }
+    obj[ownerId] = { label: ownerObj[ownerId].display_name, link: link, target: target }
     return obj;
   }, {});
 
--- a/amundsen_application/static/js/components/TableDetail/OwnerEditor/index.tsx
+++ b/amundsen_application/static/js/components/TableDetail/OwnerEditor/index.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
+import { Link } from 'react-router-dom';
 
 import ReactDOM from 'react-dom';
 import serialize from 'form-serialize';
@@ -29,9 +30,14 @@ export interface ComponentProps {
   readOnly: boolean;
 }
 
+interface OwnerAvatarLabelProps extends AvatarLabelProps {
+  link?: string;
+  isExternal?: boolean;
+}
+
 export interface StateFromProps {
   isLoading: boolean;
-  itemProps: { [id: string]: AvatarLabelProps };
+  itemProps: { [id: string]: OwnerAvatarLabelProps };
 }
 
 type OwnerEditorProps = ComponentProps & DispatchFromProps & StateFromProps;
@@ -39,7 +45,7 @@ type OwnerEditorProps = ComponentProps & DispatchFromProps & StateFromProps;
 interface OwnerEditorState {
   errorText: string | null;
   isLoading: boolean;
-  itemProps: { [id: string]: AvatarLabelProps };
+  itemProps: { [id: string]: OwnerAvatarLabelProps };
   readOnly: boolean;
   showModal: boolean;
   tempItemProps: { [id: string]: AvatarLabelProps };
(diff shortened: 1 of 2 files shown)


amundsen_application/static/js/components/common/AvatarLabel/index.tsx
```diff
@@ -1,19 +1,32 @@
...
+          target={ target }
+          className="avatar-overlap"
+          id="frequent-users"
+          onClick={logClick}
+        >
+          <Avatar name={label} src={src} size={24} round={true} />
+        </Link>
       </div>
       <label id='component-label' className='component-label'>{label}</label>
     </div>
```
danwom: - For other usages of the `<AvatarLabel>`, we have been wrapping it in a `<Link>` like so (see TableDetails/index.tsx)
```
<a href={appUrl} target='_blank' id="explore-writer" onClick={logClick}>
  <AvatarLabel label={appName} src={sourceImage}/>
</a>
```
If we're going to build links into the `<AvatarLabel>` component we might as well update the other usages as well. That being said, we've done quite a bit of refactoring in [table_V2 feature branch](https://github.com/lyft/amundsenfrontendlibrary/tree/feature/table_v2) so it might not make sense to update that now.
- I don't think it's safe to use `id="frequent-users"` since this is a generic component and not made specifically for the `frequent-users`. As a side note, the `logClick` method will pass the target element's ID as a logging parameter.
- Why not make the label on line 31 clickable as well? You'd just need to move it inside the `<Link>` component.

Mikhail-Ivanov: @danwom thank you for these findings!

Note that I've still left additional fields at AvatarLabelProps. If you feel that it may be conceptually incorrect - please let me know

amundsen_application/static/js/components/common/AvatarLabel/index.tsx
```diff
@@ -22,7 +24,9 @@ const AvatarLabel: React.SFC<AvatarLabelProps> = ({ label, src }) => {
 
 AvatarLabel.defaultProps = {
   label: '',
-  src: ''
+  src: '',
+  link: '',
+  target: ''
```
danwom: I think it makes sense to remove this part until we actually update the AvatarLabel and usage.

amundsen_application/static/js/components/TableDetail/OwnerEditor/index.tsx
```diff
@@ -198,11 +200,13 @@ export class OwnerEditor extends React.Component<OwnerEditorProps, OwnerEditorSt
         <ul className='component-list'>
           {
             Object.keys(this.state.itemProps).map((key) => {
+              const ownerItem = this.state.tempItemProps[key]
               return (
                 <li key={`list-item:${key}`}>
-                  { React.createElement(AvatarLabel, this.state.itemProps[key]) }
-                </li>
-              );
+                  <a href={ownerItem.link} target={ownerItem.target} id="table-owners" onClick={logClick}>
```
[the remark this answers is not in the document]
Mikhail-Ivanov: Done, with usage of profile_url parameter

amundsen_application/static/js/components/TableDetail/OwnerEditor/index.tsx
```diff
@@ -240,7 +244,13 @@ export class OwnerEditor extends React.Component<OwnerEditorProps, OwnerEditorSt
...
   const ownerObj = state.tableMetadata.tableOwners.owners;
   const items = Object.keys(ownerObj).reduce((obj, ownerId) => {
-    obj[ownerId] = { label: ownerObj[ownerId].display_name }
+    let profileLink = ownerObj[ownerId].profile_url;
+    let docTarget = '_blank';
+    if (AppConfig.indexUsers.enabled) {
+      profileLink = `/user/${ownerObj[ownerId].user_id}?source=owned_by`;
+      docTarget = '';
+    }
+    obj[ownerId] = { label: ownerObj[ownerId].display_name, link: profileLink, target: docTarget }
```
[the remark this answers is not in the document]
Mikhail-Ivanov: Thank you for the hint, done (I used a single link field and a boolean flag)

amundsen_application/static/js/components/common/AvatarLabel/index.tsx
Mikhail-Ivanov: Done, created OwnerAvatarLabelProps interface instead

amundsen_application/static/js/components/TableDetail/OwnerEditor/index.tsx
```diff
@@ -240,7 +267,19 @@ export class OwnerEditor extends React.Component<OwnerEditorProps, OwnerEditorSt
 export const mapStateToProps = (state: GlobalState) => {
   const ownerObj = state.tableMetadata.tableOwners.owners;
   const items = Object.keys(ownerObj).reduce((obj, ownerId) => {
-    obj[ownerId] = { label: ownerObj[ownerId].display_name }
+    const { profile_url, user_id, display_name } = ownerObj[ownerId];
+    let profileLink;
+    let isExternalLink;
+    if (AppConfig.indexUsers.enabled) {
```
danwom: I believe this misses the case where `indexUser.enabled == false` and `profileLink` should still link to an external profile_url URL. 

```
let profileLink = profile_url;
let isExternalLink = true;
if (AppConfig.indexUsers.enabled) {
  isExternalLink = false;
  profileLink = `/user/...`;
}

```

amundsen_application/static/js/components/TableDetail/OwnerEditor/index.tsx
```diff
@@ -198,11 +206,30 @@ export class OwnerEditor extends React.Component<OwnerEditorProps, OwnerEditorSt
...
+              if (owner.link === undefined) {
+                listItem = avatarLabel
+              } else if (owner.isExternal) {
+                listItem =
+                  <a href={owner.link} target="_blank" id={`table-owners:${key}`} onClick={logClick}>
+                    { avatarLabel }
+                  </a>
+              } else {
+                listItem =
+                  <Link to={owner.link} target="_blank" id={`table-owners:${key}`} onClick={logClick}>
```
danwom: I don't think you need the `target="_blank"` here.

amundsen_application/static/js/components/TableDetail/OwnerEditor/index.tsx
Mikhail-Ivanov: Okay, I thought we shouldn't display profile links if user indexing disabled. Fixed

amundsen_application/static/js/components/TableDetail/OwnerEditor/index.tsx
Mikhail-Ivanov: Personally me more liked behavior with target="_blank" when we open each user profile as a new browser tab, but I looked through the other links and see that this attribute used for external links only (<a/>) so yes, fixed for the sake of consistency

tglstory:
Does this work as expected? Currently when the user is clicked I get sent to the same table detail page.

PR #322, opened 2019-10-28
Update editable text component to new design

danwom:
### Summary of Changes

Replaced EditableText popover/hover with buttons that appear below the textarea element.

### Tests

### Documentation

### CheckList


danwom:
Screenshot of new design:
Screen Shot 2019-10-28 at 4 53 54 PM


This is the `inEditMode == true && isDisabled == true` state. I wasn't able to replicate the entire flow here, so I'm not sure if this is working properly.
Screen Shot 2019-10-28 at 4 52 54 PM

amundsen_application/static/css/_variables-default.scss
```diff
@@ -14,7 +14,7 @@ $brand-primary: $brand-color-4 !default;
 
 /* Scaffolding */
 $body-bg:           $white !default;
-$body-bg-secondary: $gray5 !default;
+$body-bg-secondary: #fcfcff !default;
```
ttannis: hardcoded??

amundsen_application/static/js/components/common/EditableText/index.tsx
```diff
@@ -96,84 +93,68 @@ class EditableText extends React.Component<EditableTextProps, EditableTextState>
...
-        >
-          <Tooltip id='save-tooltip'>
-            <button id='cancel' onClick={this.exitEditMode}>Cancel</button>
-            <button id='save' onClick={this.updateText}>Save</button>
-          </Tooltip>
-        </Overlay>
-      </div>
+                this.state.isDisabled &&
+                 <>
+                   <p>This text is out of date, please refresh the component</p>
```
ttannis: What do you think of styling this to be the `danger` color?

amundsen_application/static/css/_variables-default.scss
danwom: Oh yeah, this should be $gray0. I'll fix this in the other PR and revert this.

danwom:
@ttannis I tested this flow and found a bug (filed in Jira). The api/metadata/v0/v0/get_table_description API seems to be broken. The column descriptions work as expected. 


There is also an edge case where two users are currently editing the same description field at the same time. Since the refresh value check happens when you begin editing, it's possible to overwrite someone else's description that was submitted while you were also editing. In that case we'd have to check for the updated value before saving again.

ttannis:
> @ttannis I tested this flow and found a bug (filed in Jira). The api/metadata/v0/v0/get_table_description API seems to be broken. The column descriptions work as expected.
> 
> There is also an edge case where two users are currently editing the same description field at the same time. Since the refresh value check happens when you begin editing, it's possible to overwrite someone else's description that was submitted while you were also editing. In that case we'd have to check for the updated value before saving again.

Ah no, any idea what's wrong? Did something get overlooked when we added in markdown support?

Also that edge case is possible yes, but we accept it and decided not to do anything more about it. In reality this refresh feature is beneficial for folks who leave tabs open for a while and then might want to edit something. It wasn't one of the initial interactions planned for Amundsen, but it was our best attempt at making the experience a tiny bit better. However two people actively editing something at the same time will always have some race condition and someone's edits will be overwritten.

amundsen_application/static/js/components/common/EditableText/index.tsx
danwom: How does this look?

Screen Shot 2019-10-29 at 4 30 42 PM

danwom:
@ttannis Just took another look and the API has `/v0/v0` listed twice instead of once. Should be simple to fix. I will push the fix to master.

amundsen_application/static/js/components/common/EditableText/index.tsx
ttannis: Kind of weird actually because this presents itself like a badge, but it's not a bagde. I was initially imagining red text, but I remember our danger color is actually a softer `$sunset` color and not red. I suppose let's use best judgement if there is anyway to highlight the text without co-opting a design concept that was created for another purpose?

I'm alright if we have to leave this as an open question for design sync cleanup.

danwom:
@ttannis I had some trouble testing the `autosize` library as well as functions that used the `textareaRef.current.value` since this was null.

amundsen_application/static/js/components/common/EditableText/index.tsx
```diff
@@ -91,88 +92,72 @@ class EditableText extends React.Component<EditableTextProps, EditableTextState>
...
+    this.setState({ value: this.state.refreshValue, isDisabled: false, inEditMode: true, refreshValue: undefined });
   };
 
   updateText = () => {
     const newValue = this.textAreaRef.current.value;
-    const onSuccessCallback = () => { this.setState({value: newValue, inEditMode: false, refreshValue: undefined }); };
+    const onSuccessCallback = () => { this.setState({ value: newValue, inEditMode: false, refreshValue: undefined }); };
     const onFailureCallback = () => { this.exitEditMode(); };
 
     this.props.onSubmitValue(newValue, onSuccessCallback, onFailureCallback);
```
danwom: I'm not really sure what's the best way to test the onSuccessCallback/onFailureCallback methods.

amundsen_application/static/js/components/common/EditableText/index.tsx
```diff
@@ -1,10 +1,15 @@
...
 import * as ReactMarkdown from 'react-markdown';
 
 // TODO: Use css-modules instead of 'import'
 import './styles.scss';
+import {
+  CANCEL_BUTTON,
+  REFRESH_BUTTON,
+  REFRESH_MESSAGE,
+  UPDATE_BUTTON
+} from 'components/common/EditableText/constants';
```
ttannis: nit nit: `./constants`?

amundsen_application/static/js/components/common/EditableText/index.tsx
```diff
@@ -1,10 +1,15 @@
-import autosize from 'autosize';
+import * as autosize from 'autosize';
 import * as React from 'react';
-import { Overlay, Tooltip } from 'react-bootstrap';
 import * as ReactMarkdown from 'react-markdown';
 
 // TODO: Use css-modules instead of 'import'
 import './styles.scss';
+import {
+  CANCEL_BUTTON,
```
ttannis: nit nit: `SOMETHING_BUTTON_TEXT` as the name? Without tracing back to the files to see this is a string from constants, the name might confuse someone looking into thinking we were importing some element

PR #351, opened 2019-11-21
Inline Search Improvements + Logging

ttannis:
#### ducks
The current inline search saga behavior uses `debounce`, which prevents a request from being fired for each keystroke. However it does not fully prevent against a slower-typing user seeing intermediate results as they type. **_Example:_** Depending on how fast someone types "ttannis", they could briefly see no results for "ttan" as they are still typing. 

With the existing changes, this "glitchy" behavior is much less likely to happen, as the UI will only show the results of the latest debounced request. Alternatively we *could* have increased the debounce time, but this would also have led to a slower experience for users who type quickly. I find the current changes to be a good middle-ground and improvement upon the existing experience.

#### logging
Added `logClick` on the clickable components in `InlineSearchResults`. A custom id has been added to these components to be captured in the log.

### Tests

TODO after approval of logic

### Documentation

N/A

### CheckList


ttannis:
@danwom PTAL for merging into master with https://github.com/lyft/amundsenfrontendlibrary/pull/321

PR #352, opened 2019-11-27
Adding a separate oidc version of the image

lukelowery:
I was making my own image with oidc based on the instructions in the repo and I was thinking it would be simpler if the main build processes just also made that image. 

I ran both images this builds and everything seems to work. 

If everyone is happy with this approach, I can do the same for the metadata repo and update the helm chart too.

### DIFF
--- a/requirements3-oidc.txt
+++ b/requirements3-oidc.txt
@@ -1,4 +1,0 @@
-# A wrapper of Flask with pre-configured OIDC support.
-# License: Apache License 2.0
-# Upstream url: https://github.com/verdan/flaskoidc
-flaskoidc==0.0.2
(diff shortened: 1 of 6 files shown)


feng-tao:
cc @verdan

lukelowery:
@verdan If you are ok with this idea (even if there are some changes you want to see), I will go ahead and make the helm charts and metadata image. I have some time over the holidays to work on this again.

verdan:
@lukelowery I am fine with the approach and the implementation 👍
(sorry for the delays in the review, somehow missed that in my emails.)

requirements3-oidc.txt
```diff
@@ -0,0 +1,4 @@
+# A wrapper of Flask with pre-configured OIDC support.
+# License: Apache License 2.0
+# Upstream url: https://github.com/verdan/flaskoidc
+flaskoidc==0.0.2
```
feng-tao: not sure if we need another requirement file or just put it under extra_require in the setup.py file(e.g https://github.com/lyft/amundsendatabuilder/pull/180).

lukelowery: I like that idea a lot. Just pushed the commit. Could you re-approve it?

Thanks for the example!

PR #354, opened 2019-12-04
Update Dependencies based on Security Issues

danwom:
### Summary of Changes
Updated various dependencies and fixed various incompatibilities that arose
- Removed usages of `getDerivedStateFromProps` because it was wrong
- Moved curated tags generation into reducer/saga from component

### Tests

### Documentation

### CheckList


amundsen_application/static/js/components/common/TagsList/index.tsx
```diff
@@ -15,47 +15,20 @@ import { GetAllTagsRequest } from 'ducks/allTags/types';
...
-      };
-    }
-
-    const curatedTagsList = AppConfig.browse.curatedTags;
-    const curatedTags = allTags.filter((tag) => curatedTagsList.indexOf(tag.tag_name) !== -1);
-
-    let otherTags = [];
-    if (AppConfig.browse.showAllTags) {
-      otherTags = allTags.filter((tag) => curatedTagsList.indexOf(tag.tag_name) === -1);
-    }
```
danwom: Instead of filtering curated tags in the component, I moved this into the saga.

feng-tao:
as long as it works for FE staging, I am ok.

amundsen_application/static/js/components/common/TagsList/index.tsx
```diff
@@ -68,28 +41,36 @@ export class TagsList extends React.Component<TagsListProps, TagsListState> {
...
-            this.state.curatedTags.length > 0 && this.state.otherTags.length > 0 &&
-              <hr />
-          }
-          {this.generateTagInfo(this.state.otherTags)}
+      return <LoadingSpinner/>;
+    }
+    if (AppConfig.browse.curatedTags.length === 0) {
+      return (
+        <div id="tags-list" className="tags-list">
+          { this.generateTagInfo(this.props.allTags) }
```
ttannis: It seems like this is introducing a new feature of showing all existing tags in the `TagsList` if there are no curated tags. Do you feel strongly that this is the correct things to do?
1. I generally think changes to existing behaviors should be more intentional. 
2. Consider that this also affects the `HomePage`. Current users outside of Lyft  who use tags but don't have any curated tags will now have a big block of tags of some unknown length on their homepage.
3. Considering that we have a configuration called `showAllTags`, at the very least this should check that configuration before we actually go ahead and show all the tags. 

My overall sentiment is that I'd prefer to not change anything about this component in this PR beyond what is needed to update the dependencies, *but* if you really feel strongly that we include a configuration check for whether or not `showAllTags` is desired.

amundsen_application/static/js/components/common/TagsList/index.tsx
```diff
@@ -68,28 +41,36 @@ export class TagsList extends React.Component<TagsListProps, TagsListState> {
...
         </div>
       );
     }
-    return (innerContent);
+    return (
+      <div id="tags-list" className="tags-list">
+        { this.generateTagInfo(this.props.curatedTags) }
+        {
+          AppConfig.browse.showAllTags && this.props.curatedTags.length > 0 && this.props.otherTags.length > 0 &&
```
ttannis: A utility method has the added benefit of testing more robust while we're here. Notice that is the less recemt tests I had been directly modifying the `AppConfig` to test the desired behavior -- we risk having side effects in our tests because of that. Instead we could be safely mocking the return values of a utility method to test the different cases.

amundsen_application/static/js/components/common/TagsList/index.tsx
```diff
@@ -68,28 +41,36 @@ export class TagsList extends React.Component<TagsListProps, TagsListState> {
...
+        }
+      </div>
+    );
   }
 }
 
 export const mapStateToProps = (state: GlobalState) => {
   return {
     allTags: state.allTags.allTags,
+    curatedTags: state.allTags.curatedTags,
```
ttannis: I'm not sure there's a need to have all three of these lists passed into the component. At it's core the `TagsList` is a set of primary tags (currently `curatedTags`) and an optional list of secondary tags (currently `otherTags` if the `showAllTags` configuration is true). It seems like while we're here we can just pass a list of `primaryTags` and `secondaryTags` and `TagsList` will render them if they exist. 

The conditional logic of which piece of state gets tied to `primaryTags` and `secondaryTags` could occur here, and would still be based on filtering the one true piece of state -- `allTags`.

amundsen_application/static/js/ducks/allTags/sagas.ts
```diff
@@ -1,14 +1,18 @@
...
+import AppConfig from 'config/config';
 import * as API from './api/v0';
 import { getAllTagsFailure, getAllTagsSuccess } from './reducer';
 import { GetAllTags } from './types';
 
 export function* getAllTagsWorker(): SagaIterator {
   try {
-    const tags = yield call(API.getAllTags);
-    yield put(getAllTagsSuccess(tags));
+    const allTags = yield call(API.getAllTags);
```
ttannis: I have a hard time justifying doing all this extra work in shared code for a concept specific to `TagsList`. This action is dispatched by other components like `TagInput` that need to know about `allTags`. 

`getDerivedStateFromProps` is not necessary, yet I think we went one step to deep for the right place to create `curatedTags` and `otherTags`. Unless there's a reason why this doesn't work with the new libraries -- could we maintain `allTags` as the only necessary piece of state and create `curatedTags` and `otherTags` in `mapStateToProps` for `TagsList`?

amundsen_application/static/js/components/TableDetail/OwnerEditor/index.tsx
```diff
@@ -78,6 +69,12 @@ export class OwnerEditor extends React.Component<OwnerEditorProps, OwnerEditorSt
     this.inputRef = React.createRef();
   }
 
+  componentDidUpdate(prevProps) {
+    if (prevProps.itemProps !== this.props.itemProps) {
```
ttannis: nit: Given that `itemProps` is an object, does this check actually do anything or does it just evaluate to `true` all the time?

amundsen_application/static/js/components/TableDetail/OwnerEditor/index.tsx
```diff
@@ -25,7 +25,6 @@ export interface DispatchFromProps {
 
 export interface ComponentProps {
   errorText?: string | null;
-  readOnly: boolean;
```
ttannis: We have a check below somewhere for `!this.state.readOnly` that should be removed since it looks like we're no longer using this at all and it will always evaluate to `true` with out current logic. Can we fully clean up the usage of `readOnly` in this file?

amundsen_application/static/js/components/common/TagsList/index.tsx
danwom: I can revert this behavior, but in my personal opinion we should probably show all tags on the browse page no matter what and show curated or some subset of tags on the home page.

feng-tao: IMO I prefer to show all tags in the browse page if possible. Currently there is no way to tell what are the tags are available...

And the browse page is a bit duplicate in function with the home page tags.

ttannis: Cool. I'm open to revisiting the behavior as part of a dedicated PR that explores how to make `BrowsePage` more useful, but we can have this current PR go through with only changes needed to make us compatible with the library upgrades. 

Want to discuss offline more? I do feel like the `BrowsePage` should still have some kind of logic that allows the developers to not to show all tags if they don't want to -- their ecosystem could be full of low usage tags that don't provide much benefit. Also have thoughts on still limiting the initial amount of tags shown, but ordering them by count and creating a UI where users can expand to see more and more if they want.

feng-tao: we could discuss more tomorrow in our standup, but  I think ordering by count makes sense and we should expose more internal metadata for the tool in the browse page(e.g what are the tables got index, how tables we index, what are the db/system(bq,rs,presto,hive) we index). These infos are all valuablable but currently hidden somewhere in the backend from users.

amundsen_application/static/js/components/TableDetail/OwnerEditor/index.tsx
danwom: I'm not really sure, I just ported over the behavior from getDerivedStateFromProps but I can check.

amundsen_application/static/js/components/TableDetail/OwnerEditor/index.tsx
danwom: `state.readOnly` gets set to true when an owner update request fails. I'm opting to maintain this behavior since this PR is supposed to just upgrade libraries.

amundsen_application/static/js/components/common/TagsList/index.tsx
danwom: Moved to config-utils

amundsen_application/static/js/components/common/TagsList/index.tsx
danwom: Creation of `curatedTags` and `otherTags` is moved to the mapStateToProps and removed from reducer/saga

amundsen_application/static/js/ducks/allTags/reducer.ts
```diff
@@ -31,7 +31,11 @@ export default function reducer(state: AllTagsReducerState = initialState, actio
     case GetAllTags.FAILURE:
       return initialState;
     case GetAllTags.SUCCESS:
-      return { ...state, allTags: (<GetAllTagsResponse>action).payload.tags, isLoading: false };
+      return {
```
ttannis: nit (No changes requested): I get why this simplifies this current use case, but I'd encourage us to continue being conscious of the fact that logic which leverages the fact that the response payload and redux state happen to be the same can create extra work as the application scales. Apis can potentially be extended to include new information, and application state can potentially be restructured for performance/best practices. If that day comes with the code as is sure it would be a quick fix, but if we ended up building further tag logic and helper methods with this assumption we'd have a lot of extra work to do.

amundsen_application/static/js/config/config-utils.ts
```diff
@@ -51,3 +51,17 @@ export function indexUsersEnabled(): boolean {
...
+ * Returns whether or not to show all tags
+ */
+export function showAllTags(): boolean {
+  return AppConfig.browse.showAllTags;
+}
+
+/**
+ * Returns a list of curated tag names
+ */
+export function getCuratedTags(): string[] {
```
ttannis: nit: Missing test for `getCuratedTags()`

amundsen_application/static/js/ducks/allTags/reducer.ts
danwom: I see what you mean. This made more sense when `curatedTags` and `otherTags` were also part of the payload, but after removing them it might make more sense to explicitly list out the `allTags` field.

PR #357, opened 2019-12-09
Initial Filter React/Redux Modifications

ttannis:
### Summary of Changes

This PR completes the majority of the React & Redux work needed to implement filtering on the `SearchPage` for datasets. This PR will not cover:
1. A working Flask endpoint to the new query string endpoint on the search service
2. Updating the url with filter information (need a working endpoint)

#### Configurations
1. The `resourceConfig` data shape has been extended to encompass the concept of `displayName` and `filterCategories` for each supported resource. The configuration for tables also has a `supportedDatabases` config which is the equivalent of the old `datasets` config.
2. The `filterCategories` config is an array of configuration objects for each supported category for that resource. Given our current UI there are two types of configuration objects -- the `FilterType.MULTI_SELECT_VALUE` type will be displayed in checkbox format and the `FilterType.SINGLE_VALUE` type will be displayed in input box format. 

#### SearchBar
1. The concept of advanced search syntax has been removed from the `SearchBar` logic.
2. **Open Question:** Should there be any syntax validation on the `SearchBar` and if so what would that look like.

#### SearchFilter
1. Exposed the `SearchFilter` component
2. Modified `SearchFilter` to render checkbox sections and input sections based on the given `filterCategories`.
    
3. Created a separate `InputFilter` component to help with the implementation. This component is a stop-gap for our UI workaround and isn't intended to be a new common component.

#### Miscellaneous Component Work
1. Minor changes to relevant styles.
2. Helper methods modified/added where applicable.

#### Reducers/Actions
1. Created a new reducer that will handle the filter state, which is nested into the search state.
2. Created actions for handling different user interactions for updating a filter.
3. The `search` reducer logic has been modified to either clear or persist filter states when applicable.

#### Sagas
1. There are two new watchers to watch for updates to filters. One listens for events fired by the "multi-select" type and debounce those requests since users are likely to quickly select multiple options. The second listens only for the latest update from the "single update" type. Both result in an update to the filter state, and the `searchResource` action being fired.
2. The worker that process the `searchResource` action now passes the filters for that resource to the api method.

#### APIs
1. The `searchResource` api method (not the action of the same name) takes optional filters. If filters exist, we will call the soon-to-be-written endpoint that will support calling the new query string endpoint on the search service.

### Tests

TODO: To be written before merging feature branch to master

### Documentation

TODO: To be written before merging feature branch to master

### CheckList


ttannis:
@danwom This is ready for review, to merge into a feature branch since the implementation is incomplete. In this PR I intend to address everything except:
1. A working Flask endpoint
2. Held off on the url update work to be it's own PR

amundsen_application/static/js/ducks/search/filters/reducer.ts
```diff
@@ -0,0 +1,119 @@
...
+export type FilterCategories = { [id:string]: string };
+export interface FilterReducerState {
+  [ResourceType.table]: TableFilterReducerState;
+};
+export interface TableFilterReducerState {
+  column: string;
+  database: FilterCategories;
+  schema: string;
+  table: string;
+  tag: string;
```
danwom: Isn't this tightly coupled with the resource config? It should be possible to use generic types, maybe something like: 

```
interface MultiFilter {
  category: string;
  values: string[];
}
interface SingleFliter {
 category: [];
 value: string;
}

export interface TableFilterReducerState {
  multiFilters: MultiFilter[];
  singleFilters: SingleFilter[];
}
```

Or we could make these objects rather than arrays

```
interface SingleFilters {
  [category: string]: string
}

interface MultiFIlters {
  [category: string]: string[];
}
```

amundsen_application/static/js/ducks/search/filters/reducer.ts
```diff
@@ -0,0 +1,119 @@
...
+  table: '',
+  tag: '',
+};
+
+export const initialFilterState: FilterReducerState = {
+  [ResourceType.table]: initialTableFilterState,
+};
+
+export default function reducer(state: FilterReducerState = initialFilterState, action, resourceType: ResourceType): FilterReducerState {
+  const { category, value  } = action.payload;
```
danwom: I don't have any ideas at the moment, but we should try find a way to simplify this function.

amundsen_application/static/js/ducks/search/sagas.ts
```diff
@@ -42,6 +42,21 @@ import {
...
+import {
+  UpdateSearchFilter
+} from './filters/reducer';
+
+export function* filterWorker(action: any): SagaIterator {
+  const state = yield select();
+  yield put(searchResource(state.search.search_term, state.search.selectedTab, getPageIndex(state.search)));
+};
+export function* multiSelectFilterWatcher(): SagaIterator {
+  yield debounce(750, [UpdateSearchFilter.ADD_MULTI_SELECT, UpdateSearchFilter.REMOVE_MULTI_SELECT], filterWorker);
```
danwom: Does the `debounce` also guarantee we only take the latest response? It's probably not an issue with our current low latency, however.

amundsen_application/api/search/v0.py
```diff
@@ -36,6 +36,19 @@ def _create_error_response(*, message: str, payload: Dict, status_code: int) ->
...
+    return {
+        'type': 'table',
+        'key': result.get('key', None),
+        'name': result.get('name', None),
+        'cluster': result.get('cluster', None),
+        'description': result.get('description', None),
+        'database': result.get('database', None),
+        'schema_name': result.get('schema_name', None),
+        'last_updated_epoch': result.get('last_updated_epoch', None),
+    }
```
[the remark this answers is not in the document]
ttannis: Yes, but I think those efforts are more appropriately part of the larger task to unify the data format returned by both search and metadata. I'll leave this piece as is as far as the filter branch is concerned.

amundsen_application/api/search/v0.py
```diff
@@ -57,6 +70,65 @@ def _validate_search_term(*, search_term: str, page_index: int) -> Optional[Resp
     return None
 
 
+@search_blueprint.route('/table_qs', methods=['POST'])
+def search_table_qs() -> Response:
```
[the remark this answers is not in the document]
ttannis: This method is currently separate for better development. This comment will be addressed in an upcoming PR.

amundsen_application/static/js/components/SearchPage/SearchFilter/index.tsx
```diff
@@ -1,168 +1,174 @@
...
+        onChange={ this.props.onCheckboxChange }>
           <span className="subtitle-2">{ labelText }</span>
-          <span className="body-secondary-3 pull-right">{ count }</span>
       </CheckBoxItem>
     );
   };
 
-  createCheckBoxSection = (section: SearchFilterSection, key: string) => {
-    const { categoryId, inputProperties, title } = section;
+  createCheckBoxSection = (key: string, section: CheckboxFilterSection) => {
```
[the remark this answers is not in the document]
ttannis: I'll keep it as is at least until we finish iterating through the UI explorations and potentially getting user feedback. This comment might be able to be addressed in a future PR, based on whatever version of this UI we settle on.

amundsen_application/static/js/components/SearchPage/SearchFilter/index.tsx
```diff
@@ -1,168 +1,174 @@
...
-            checked: false, // pull value or infer value from state
-            count: 3, // pull value from state
-          }
-        ]
+  const resourceType = state.search.selectedTab;
+  const filterCategories = getFilterConfigByResource(resourceType);
+  const filterState = state.search.filters;
+
+  const checkBoxSections = [];
+  const inputSections = {};
```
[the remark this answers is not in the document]
ttannis: There is no reason for this one anymore -- an earlier version of this work had logic where `SearchFilter` was handling the logic for the input change, and it was more efficient to store the information as an object for O(1) access to the input value for a given `categoryId`. I will update this PR to have `inputSections` as an array.

amundsen_application/static/js/components/common/SearchBar/index.tsx
```diff
@@ -109,33 +100,9 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
...
-        subTextClassName: ERROR_CLASSNAME,
-      });
-      return false;
-    }
-
-    this.setState({ subText: SUBTEXT_DEFAULT, subTextClassName: "" });
-    return true;
+  isFormValid = () : boolean => {
+    const form = document.getElementById("search-bar-form") as HTMLFormElement;
+    return form.checkValidity();
```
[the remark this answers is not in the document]
ttannis: It can, depending on what rules we set for the `SearchBar`. As of this upload it will stop submission if the pattern is not matched or the search bar is empty. Unsure whether these rules will apply in the final version, but if not then the method will be removed

amundsen_application/static/js/config/config-types.ts
```diff
@@ -52,13 +55,47 @@ interface BrowseConfig {
...
+}
+
+export type FilterConfig = (MultiSelectFilterCategory|SingleFilterCategory)[];
+
+interface BaseResourceConfig {
+  displayName: string;
+  filterCategories?: FilterConfig;
+}
+
+interface DatasetResourceConfig extends BaseResourceConfig {
```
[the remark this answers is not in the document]
ttannis: There's a visions that `Dataset` can include things that aren't tables. Granted at that point we'll have a lot more to change than this config name, so I will use `TableResourceConfig` instead until we actually need to consider other things. (e.g. - S3 buckets)

amundsen_application/static/js/ducks/search/api/v0.ts
```diff
@@ -27,11 +27,18 @@ export const searchResourceHelper = (response: AxiosResponse<SearchAPI>) => {
...
      (resource === ResourceType.user && !AppConfig.indexUsers.enabled)) {
     return Promise.resolve({});
   }
+  if (filters) {
+    return axios.post(`${BASE_URL}/${resource}_qs`, {
+      filters,
+      term,
+      pageIndex
+    }).then(searchResourceHelper);
+  }
```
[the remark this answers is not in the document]
ttannis: This might still end up being two APIs because one is a `get` while the other is a `post`. Explorations on the frontend regarding the behavior of the current search endpoint will affect what gets done here

amundsen_application/static/js/ducks/search/filters/reducer.ts
ttannis: I will explore the second option, I agree it should be improved. This version came from going in circles a bit too long on this part and hardcoding things to work.

amundsen_application/static/js/ducks/search/filters/reducer.ts
ttannis: I believe simplification will come with some iterations of this feature. The simplified state suggested above might help.

amundsen_application/static/js/ducks/search/sagas.ts
ttannis: Yes, unless I'm misinterpreting the doc. https://github.com/redux-saga/redux-saga/blob/master/docs/recipes/README.md#debouncing

ttannis:
@danwom ready for re-review. 

**Changes include:**
1. Updates to the Flask code just to get it working with current search api. Placed a `TODO` in `amundsen_application/api/search/v0.py` to address in an upcoming PR
2. Updated the filter state. Once concept here is that I want the values for a category to be immediately accessible so the new shape of the filter state is `[category: string]: string | FilterOptions;`. This also helped simplify some of the reducer logic. I'm open to continuing to iterate on this if something better becomes apparent as we go through iterations.
3. Referenced "table" and "database" in the config name types when appropriate, as opposed to using "dataset"
4. Removed pattern match checking from `SearchBar`. Since our UI is no longer set up to give hints about syntax I think the long term solution is to let users type whatever and we should somehow handle it. This is something I'm looking to explore as part of some user feedback.
5. Cleaned up `inputSections` on the `SearchFilter`

**Future PRs intend to:**
1. Address some code cleanup items that I am unsure of now, such as how we want to breakup the `SearchFilter` components, and consolidating api logic.
2. Cleanup all "TODO: (ttannis)" items
3. Url update work
4. Add unit tests
6. Add documentation
6. Metrics

Please call out anything else not counted in these items that I can address in this PR.

amundsen_application/api/search/v0.py
```diff
@@ -57,6 +71,73 @@ def _validate_search_term(*, search_term: str, page_index: int) -> Optional[Resp
...
+    search_term = get_query_param(request_json, 'term', '"term" parameter expected in request data')
+    page_index = get_query_param(request_json, 'pageIndex', '"pageIndex" parameter expected in request data')
+    filters = request_json.get('filters', {})
+
+    filter_payload = {}
+    valid_categories = ['column', 'database', 'schema', 'table', 'tag']
+    for category in valid_categories:
+        values = filters.get(category)
+        if type(values) == str and len(values) > 0:
+            filter_payload[category] = [values, ]
```
danwom: nit: extra trailing comma?

amundsen_application/api/search/v0.py
```diff
@@ -57,6 +71,73 @@ def _validate_search_term(*, search_term: str, page_index: int) -> Optional[Resp
...
+        if type(values) == dict:
```
danwom: What's the use case for filter values being a dictionary? I was under the assumption filter values would either be a string or an array of strings.

amundsen_application/static/js/ducks/search/filters/reducer.ts
danwom: I have a proposal that might simplify things: 
- Each filter component (checkbox, text input) has some attributes such as `name`, `value`, and `checked`. We could leverage each of these form input attributes
- On form submission, the SearchFilters component can iterate over the form elements and generate a new filters json object with some simple rules. 
  - Text Input: If the value is non-empty, we can add a filter. 
  ```
  filters = {};
  ...
  if (textInput.value !== null) 
    filters[textInput.name] = textInput.value
  ```
  - Checkbox: If checked, add the `value` to the `name` category. (Create a new array, or append if it exists). 
  ``` 
  filters = {};
  ...
  if (checkbox.checked) {
    if (filters[checkbox.name] !== undefined) 
      filters[checkbox.name] = [checkbox.value];
    else
      filters[checkbox.name].push(checkbox.value);
  } 
  ```

amundsen_application/api/search/v0.py
ttannis: This is a python thing when it comes to transforming a string into an array. If `values = 'ttannis'` then `[values] =['t', 't', 'a', 'n', 'n', 'i', 's']`  but `[values, ] =['ttannis']`

amundsen_application/api/search/v0.py
ttannis: Matching the current `FilterReducerState`, as the code just sends over the pieces of state and delegates the conversion to this method. I see you have some further comments about the state so that could always change what happens here. This method is essentially in development because of both potential amundsensearch api changes *and* redux state changes.

amundsen_application/static/js/ducks/search/filters/reducer.ts
ttannis: I see. I'm going to explore this one in my other branch `ttannis_db-filter-2`. I had went ahead to do the `SearchFilter` cleanup that involved making shared `FilterSection` component that renders the section header and either `CheckboxFilter` or `InputFilter` as a child and those changes affect how I would try to make this work.

It sounds like I'd have to add support to the `SearchFilter` to get the piece of json for each `FilterSection` and that requires having a `ref` to each `FilterSection` to call the relevant method. I hope that doesn't violate any best practices but I think I already see how much this would clean up the reducer. 

I will explore in `ttannis_db-filter-2` and open a pull request again this branch.

amundsen_application/api/search/v0.py
ttannis: Ah also worth noting the dictionary shape was ideal for knowing if a multi-select option was checked. 

Here is an example `FilterReducerState`:
```
{
   'database': { 'bigquery': true, 'hive': true }
   'tag': 'ttannis'
}
```

Here is what amundsensearch needs:
```
{
   'database':  [ 'bigquery', 'hive']
   'tag': ['ttannis']
}
```

While it requires a conversion somewhere or the other (whether in Flask or Redux), the concept of modeling the application state based on what is best for the application's use cases for accessing that state is what guided this choice.

amundsen_application/static/js/components/SearchPage/SearchFilter/index.tsx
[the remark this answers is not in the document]
ttannis: This comment is now addressed in work from this PR: https://github.com/lyft/amundsenfrontendlibrary/pull/361

amundsen_application/static/js/ducks/search/filters/reducer.ts
ttannis: This comment is now addressed in work from this PR: https://github.com/lyft/amundsenfrontendlibrary/pull/361

PR #363, opened 2019-12-20
Amundsen Notifications functionality improvements

Mikhail-Ivanov:
### Summary of Changes

 - "Request Description" link visibility fix
 - MailClient implementation example
 - API incompatibilities fixes

### Tests

No tests were added or modified for now.

### Documentation

No documentation added for now.

closes https://github.com/lyft/amundsen/issues/211

### DIFF
--- a/amundsen_application/api/utils/notification_utils.py
+++ b/amundsen_application/api/utils/notification_utils.py
@@ -192,13 +192,14 @@ def send_notification(*, notification_type: str, options: Dict, recipients: List
         )
 
         response = mail_client.send_email(
-            recipients=recipients,
-            sender=sender,
             subject=subject,
+            text='Notification text',  # TODO generate a proper text (now html being used instead anyway)
             html=html,
             optional_data={
                 'email_type': notification_type,
             },
+            sender=sender,
+            recipients=recipients,
         )
         status_code = response.status_code
 
--- a/amundsen_application/api/utils/notification_utils.py
+++ b/amundsen_application/api/utils/notification_utils.py
@@ -192,14 +192,13 @@ def send_notification(*, notification_type: str, options: Dict, recipients: List
         )
 
         response = mail_client.send_email(
-            subject=subject,
-            text='Notification text',  # TODO generate a proper text (now html being used instead anyway)
             html=html,
+            subject=subject,
             optional_data={
                 'email_type': notification_type,
             },
-            sender=sender,
             recipients=recipients,
+            sender=sender,
         )
         status_code = response.status_code
 
--- a/amundsen_application/base/base_mail_client.py
+++ b/amundsen_application/base/base_mail_client.py
@@ -11,20 +11,18 @@ class BaseMailClient(abc.ABC):
 
     @abc.abstractmethod
     def send_email(self,
-                   subject: str,
-                   text: str,
                    html: str,
-                   sender: str,
+                   subject: str,
+                   optional_data: Dict,
                    recipients: List[str],
-                   optional_data: Dict) -> Response:
+                   sender: str) -> Response:
         """
         Sends an email using the following parameters
-        :param subject: The subject of the email
-        :param text: Plain text email content
         :param html: HTML email content
-        :param sender: The sending address associated with the email
-        :param recipients: A list of recipients for the email
+        :param subject: The subject of the email
         :param optional_data: A dictionary of any values needed for custom implementations
+        :param recipients: A list of recipients for the email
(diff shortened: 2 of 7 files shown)


amundsen_application/api/utils/notification_utils.py
```diff
@@ -192,13 +192,14 @@ def _log_send_notification(*, notification_type: str, options: Dict, recipients:
...
         response = mail_client.send_email(
-            recipients=recipients,
-            sender=sender,
             subject=subject,
+            text='Notification text',  # TODO generate a proper text (now html being used instead anyway)
             html=html,
             optional_data={
                 'email_type': notification_type,
             },
+            sender=sender,
```
ttannis: nit: Might was well put the others in alphabetical order if that was the intent :)

amundsen_application/base/base_mail_client.py
```diff
@@ -11,19 +11,19 @@ def __init__(self, recipients: List[str]) -> None:
 
     @abc.abstractmethod
     def send_email(self,
-                   sender: str,
-                   recipients: List[str],
                    subject: str,
                    text: str,
                    html: str,
+                   sender: str,
```
ttannis: nit: It looks like you had to change this because you changed the order of how things were called in `notification_utils.py`. As a cleanup, if we switch using keyword args we won't have to worry about parameter order. Might as well have them alphabetical as well.
```
def send_email(self, *, subject: str, ....)
```

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -158,7 +158,7 @@ class TableDetail extends React.Component<TableDetailProps & RouteComponentProps
                   editable={ data.is_editable }
                 />
               </EditableSection>
-              { !data.table_description && notificationsEnabled() && <RequestDescriptionText/> }
```
ttannis: It was an intentional decision to have the `RequestDescriptionText` option even when there is a description. The decision considered the case of an existing description not being clear enough or needing more detail, and this would allow users a pathway to request an improved description.

amundsen_application/base/examples/example_mail_client.py
```diff
@@ -0,0 +1,66 @@
...
+                   sender: str = None,
+                   recipients: List[str] = None,
+                   optional_data: Dict = None) -> Response:
+        if not sender:
+            sender = '< amundsen email >'  # change me
+        if not recipients:
+            recipients = self.recipients
+
+        # TODO refactor that, it's insecure!
+        sender_pass = '< amundsen email password >'  # change me
```
ttannis: nit: Even though you have a clear message here that this is to be changed when writing the actual client, maybe the example can encourage the use of getting this from an environment variable?

amundsen_application/api/utils/notification_utils.py
```diff
@@ -192,13 +192,14 @@ def _log_send_notification(*, notification_type: str, options: Dict, recipients:
         )
 
         response = mail_client.send_email(
-            recipients=recipients,
-            sender=sender,
             subject=subject,
+            text='Notification text',  # TODO generate a proper text (now html being used instead anyway)
```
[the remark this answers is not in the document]
Mikhail-Ivanov: Sure, I can simply remove text argument if you're okay with that

amundsen_application/static/js/components/TableDetail/index.tsx
Mikhail-Ivanov: Sounds good, I agree that in such a case
> an existing description not being clear enough or needing more detail

users should be able to request changes. That's why I removed `!data.table_description` part of the statement - it makes "Request Description" link hidden

amundsen_application/api/utils/notification_utils.py
Mikhail-Ivanov: No, the reason was to rearrange parameters to place required ones first and then optional ones. But yes, I can place each group in alphabetical order as well 👍

amundsen_application/base/examples/example_mail_client.py
Mikhail-Ivanov: Done

ttannis:
@Mikhail-Ivanov fyi I'll aim to take another look by end of week -- coming back from several weeks of vacation 😄

amundsen_application/static/js/components/TableDetail/index.tsx
ttannis: Ah my mistake, for some reason my brain processed this logic backwards.

PR #365, opened 2020-01-08
Update profile page design

danwom:
### Summary of Changes

Update the profile page design have a similar layout as the table details page.

### Tests

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/ProfilePage/styles.scss
+++ b/amundsen_application/static/js/components/ProfilePage/styles.scss
@@ -3,6 +3,7 @@
 .profile-page {
 
   .profile-avatar {
+    display: inline-block;
     margin: 6px 16px 0 0;
   }
 
--- a/amundsen_application/static/js/components/ProfilePage/styles.scss
+++ b/amundsen_application/static/js/components/ProfilePage/styles.scss
@@ -2,55 +2,60 @@
 
 .profile-page {
 
-  .profile-header {
-    display: flex;
-
-    .profile-avatar {
-      margin-left: 8px;
-      margin-right: 30px;
-    }
-
-    .profile-details {
-      display: flex;
-      flex-direction: column;
-
-      h1 {
-        margin-bottom: 0;
-        margin-top: 0;
-      }
-
-      .profile-icons {
-        margin-top: 12px;
-
-        a {
-          margin-left: 8px;
-          margin-right: 8px;
-
-          &:first-child {
-            margin-left: 0;
-          }
-          &:last-child {
-            margin-right: 0;
-          }
-        }
-      }
-
-      .profile-title {
-        display: flex;
-        margin-bottom: 8px;
-
-        .flag {
-          height: min-content; // TODO: consider moving height into Flag component
-          font-size: 16px;
-          margin: auto auto 4px 12px;
-        }
-      }
-
-    }
+  .profile-avatar {
+    margin: 6px 16px 0 0;
   }
 
+  .profile-details {
+    display: flex;
+    flex-direction: column;
+
+    h1 {
+      margin-bottom: 0;
+      margin-top: 0;
+    }
+
+    .profile-icons {
+      margin-top: 12px;
+
+      a {
+        margin-left: 8px;
+        margin-right: 8px;
+
+        &:first-child {
+          margin-left: 0;
+        }
+        &:last-child {
+          margin-right: 0;
+        }
+      }
+    }
+
+    .profile-title {
+      display: flex;
+      margin-bottom: 8px;
+
+      .flag {
+        height: min-content; // TODO: consider moving height into Flag component
+        font-size: 16px;
+        margin: auto auto 4px 12px;
+      }
+    }
+
+  }
+
+
   .profile-tabs {
-    margin-top: 64px;
+    margin-top: 24px;
+
+    .nav-tabs {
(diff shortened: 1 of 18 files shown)


ttannis:
Some additional UI comments:

1. Now that the Avatar dropdown covers an important part of the new page layout, if it is an easy fix I think we should make sure it closes when someone selects a list item.
2. The list item in the Avatar dropdown has an unintended background color after clicking it navigating away.
    
3. Nit: What do you think of also including design aspects from our new `SearchPage` (as `ProfilePage` is a resource page like `TableDetail` but also displays table list items like `SearchPage`), where we have a side panel with radio buttons instead of tabs for "Bookmarked", "Frequently Used", and "Owned"? I know we don't have an official design request for what this is supposed to look like, was there talk about moving away from tabs or am I just imagining that?
4. Nit: Do you think it makes sense to change the number of results per page in this view or keep it at 6?

amundsen_application/static/js/components/common/Breadcrumb/index.tsx
```diff
@@ -13,51 +12,35 @@ export interface OwnProps {
...
-        <div className="amundsen-breadcrumb">
-          <a onClick={ props.loadPreviousSearch } className='btn btn-flat-icon title-3'>
-            <img className='icon icon-left'/>
-            <span>Search Results</span>
-          </a>
-        </div>
-      );
-    }
+  const { path, text } = props;
+  if (path !== undefined && text !== undefined) {
```
ttannis: 1. Nit: With this change we no longer have a use case for a `Breadcrumb` with text, and I also haven't seen it in any of our future designs anymore, so I'd be up for simplifying the component to just take `path`.
2. Nit: Do we not want to protect against `path` being an empty string? I don't feel strongly either way because I think the onus is on the developer to use the component props properly.

amundsen_application/static/js/ducks/search/sagas.ts
```diff
@@ -204,6 +204,10 @@ export function* urlDidUpdateWatcher(): SagaIterator {
 
 export function* loadPreviousSearchWorker(action: LoadPreviousSearchRequest): SagaIterator {
   const state = yield select(getSearchState);
+  if (state.search_term === "") {
```
ttannis: nit: add a test to verify `BrowserHistory.goBack()` is called? If for some reason this test is tricky to add I'm fine with `TODO` in the code because there's a backlog task to come back to covering some of the untested search saga logic.

amundsen_application/static/js/components/ProfilePage/tests/index.spec.tsx
```diff
@@ -228,14 +228,14 @@ describe('ProfilePage', () => {
       expect(wrapper.find(DocumentTitle).props().title).toEqual(`${props.user.display_name} - Amundsen Profile`);
     });
 
-    it('renders Breadcrumb', () => {
-      expect(wrapper.find(Breadcrumb).exists()).toBe(true)
-    });
+    // it('renders Breadcrumb', () => {
```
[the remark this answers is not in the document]
danwom: I removed the breadcrumb, and then added it back in. Will uncomment.

amundsen_application/static/js/components/common/Breadcrumb/index.tsx
danwom: I did initially remove the path and text props, but added them back in. There is one usage in the 404 page but if we decide that is not useful, we can cleanup here.

amundsen_application/static/js/ducks/search/sagas.ts
danwom: I didn't spend too much time but I couldn't figure out how to test that within the saga test.

amundsen_application/static/js/components/common/Breadcrumb/index.tsx
ttannis: Let's leave as is then and not increase scope.

amundsen_application/static/js/ducks/search/sagas.ts
ttannis: No problem, fine with being part of the overall unit test backlog task.

danwom:
1 & 2 should be fixed.
3: The layout in this PR is in accordance with design mocks. I think the tabbed layout is fine since we won't have any filters for this page. 
4: I arbitrarily picked 6 items on this page as it was how many items would fit without scrolling on my screen. The recently the list items grew taller, but we also opened more space up near the top with this new design so 6 items still fit at my resolution.

amundsen_application/static/js/components/ProfilePage/styles.scss
```diff
@@ -1,51 +1,17 @@
 @import 'variables';
 
 .profile-page {
-
   .profile-avatar {
     display: inline-block;
     margin: 6px 16px 0 0;
   }
 
-  .profile-details {
```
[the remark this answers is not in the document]
danwom: Yes, these were all for the old layout.

danwom:
I think this specific branch doesn't have the latest updates to the NPM packages, so it's not compatible with your current version of node.

ttannis:
> I think this specific branch doesn't have the latest updates to the NPM packages, so it's not compatible with your current version of node. (should work with node 10 or lower)

That must of been it. Master still building fine for me with these changes.

PR #376, opened 2020-01-23
Improve user search experience

ttannis:
### Summary of Changes

This PR further addresses issues from our user interviews, namely:
1. Persisting filters when re-executing a search on the `SearchPage`
2. Removing the requirement of a search term, which allows users to execute higher level searches across categories, e.g. "I'm looking for all resources in this schema"

**NavBar**
1. Passes the current location to `SearchBar`

**SearchBar**
1. Removed `isFormValid` check and requirement for a search term to execute a search
2. Passes a boolean to the `submitSearch` action to specify whether or not the existing filters should be used in this search.

**SearchPage**
1. Utilizes a new `hasFilters` boolean to assist with the error message logic. Our new user flow allows for users to clear all search input (search term + filters), in which case they now see a blank page. _Note: Will get feedback in Design QA about whether or not we should add some instruction, or otherwise better utilize this space._

**Redux Changes**
1. `submitSearch` and `searchAll` actions take a `useFilters` boolean in the payload.
2. `searchAll` no longer resets the filter state. Instead there is a new `clearAll` action that is fired by `searchAllWorker` if `useFilters` is false.

### Tests

TODO: As part of final feature branch review

### Documentation

TODO: As part of final feature branch review

### CheckList
    - In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x)

### DIFF
--- a/amundsen_application/static/js/components/common/SearchBar/index.tsx
+++ b/amundsen_application/static/js/components/common/SearchBar/index.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as History from 'history';
 import { bindActionCreators } from 'redux'
 import { connect } from 'react-redux';
 
@@ -23,12 +24,13 @@ export interface StateFromProps {
 }
 
 export interface DispatchFromProps {
-  submitSearch: (searchTerm: string) => SubmitSearchRequest;
+  submitSearch: (searchTerm: string, useFilters?: boolean) => SubmitSearchRequest;
   onInputChange: (term: string) => InlineSearchRequest;
   onSelectInlineResult: (resourceType: ResourceType, searchTerm: string, updateUrl: boolean) => InlineSearchSelect;
 }
 
 export interface OwnProps {
+  location?: History.Location;
   placeholder?: string;
   size?: string;
 }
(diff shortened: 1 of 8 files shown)


amundsen_application/static/js/components/common/SearchBar/index.tsx
```diff
@@ -88,23 +90,16 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
 
   handleValueSubmit = (event: React.FormEvent<HTMLFormElement>) : void => {
     const searchTerm = this.state.searchTerm.trim();
+    const useFilters = this.props.location ? this.props.location.pathname === '/search' : false;
```
danwom: Is there a better place for this logic? If possible I wouldn't want the `SearchBar` to be aware of `location` and which routes should preserve the filters. 

- Maybe the reducer or saga should be aware of the route and decide if the filters should be used or not.
- Maybe there should be a small checkbox somewhere that would explicitly decide to preserve or clear filters.

ttannis: Hope the following is clear I time-boxed some investigation:

1. After some investigation I'm not confident this is a quick fix as opposed to separate task of augmenting our usage of Redux on that manner. I couldn't quite wrap my head around the safest way to do this that is 100% guaranteed to keep the redux state and react-router in sync -- if it exists.

2. If the ability to toggle `useFilters` or not was equal work effort I'd be confident to include it, but in this case there would be extra effort to add this to the application state and ensure page navigations still somehow update this state. I think it's worth investing effort into that toggle implementation only if users request it, or Design supports it. I'd essentially hate to spend a couple days on this and then have to remove it after Design QA. It is worth proposing as part of the final design QA though.

In summary we could talk about this more offline as its own task. I boil this down to the fact that we need a `SearchBar` that will cause either one of two action chains to happen:
1. clearFilters & submitSearch 
2. submitSearch

Meaning that `SearchBar` could take its submit method as a passed property, and it's parent components which are connected to the router could pass the correct submit action to the `SearchBar`. Let me know what you think, and I can add the modification before final feature branch review.

ttannis: FYI references -- first found [this library](https://github.com/reactjs/react-router-redux) which might have been overcomplicating things and is now deprecated, but it could also be as little effort as some of[ these suggestions](https://www.pluralsight.com/guides/firing-redux-actions-on-route-transitions)

PR #377, opened 2020-01-24
Add support for configurable filter section help text

ttannis:
### Summary of Changes

We want to offer some optional help text for each filter section. This has been added to the optional `filterCategories` configuration, and if present will result in an `InfoButton` in the filter section header -- exact text will likely still go through some iteration in design qa

### Tests

TODO: As part of final feature branch review

### Documentation

TODO: As part of final feature branch review

### CheckList

### DIFF
--- a/amundsen_application/static/js/config/config-default.ts
+++ b/amundsen_application/static/js/config/config-default.ts
@@ -1,6 +1,6 @@
 import { AppConfig } from './config-types';
 
-import { FilterType, ResourceType } from '../interfaces';
+import { ResourceType } from '../interfaces';
 
 const configDefault: AppConfig = {
   badges: {},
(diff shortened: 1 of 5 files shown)


amundsen_application/static/js/config/config-default.ts
```diff
@@ -65,40 +65,6 @@ const configDefault: AppConfig = {
...
-        {
-          value: 'table',
-          displayName: 'Table',
-          type: FilterType.SINGLE_VALUE,
-        },
-        {
-          value: 'tag',
-          displayName: 'Tag',
-          type: FilterType.SINGLE_VALUE,
-        },
```
danwom: I think we should keep the column, schema, table, and tag filters as default since those are base properties of all Amundsen tables. The database options change for each company, but maybe adding a SINGL_VALUE filter for database would be a good default.

ttannis: Sure thing I'll add these back in as the default

PR #381, opened 2020-01-28
Issue 147/multi doc pathways

samshuster:
### Summary of Changes
mr to get feedback as part of https://github.com/lyft/amundsen/issues/147
Currently blocked by amundsen-common being merged and getting feedback.

Here is current screenshot. You can see the added sections in bottom left.
Screen Shot 2020-02-13 at 1 46 14 PM
Screen Shot 2020-02-13 at 1 46 20 PM

Current features:
- supports 0 to n separate programmatic sections
- supports markdown
- does not support editing
- does not currently support search, but very easy to add that
- supports configurable display names for the sections as well as order

### Tests

Updated python api tests for configuring the order of the sections
Updated the javascript tests to the best of my ability, but I believe more tests could be used here.

### Documentation

Updated some documentation on how to configure these sections

### CheckList


samshuster:
ok @ttannis this is ready for you to review!

jornh:
Would be good to mention `PROGRAMMATIC_DISPLAY` config in https://github.com/lyft/amundsenfrontendlibrary/pull/381/files#diff-da4c534f10ac33888daef3b30ba35474 In the Flask config documentation: https://github.com/lyft/amundsenfrontendlibrary/blob/master/docs/flask_config.md

Also, maybe worth showcasing the demo screenshots above in the same document? I’d prefer that over keeping adding optional customization screenshots to the main README (otherwise it’s just getting longer and longer)

Finally maybe also worth to point to how to ingest data as introduced in databuilder https://github.com/lyft/amundsendatabuilder/pull/187?

samshuster:
ok @jornh documentation for frontend added. I will go back and add in some documentation to databuilder too.

docs/flask_config.md
```diff
@@ -20,3 +20,40 @@ def custom_route():
...
+               "display_order": 1,
+               "display_title": "Quality"
+           },
+           "doesnt_exist": {
+               "display_order": 2,
+               "display_title": "Nobody Cares"
+           }
+       }
+```
+description sources not mentioned in the configuration will be alphabetically placed at the end of the above list.
```
jornh:
```suggestion
description sources not mentioned in the configuration will be alphabetically placed at the end of the above list. If `PROGRAMMATIC_DISPLAY` is left at `None` all added fields are still showing up, so that display is entirely dynamically data-driven without configuration. Meaning configuration merely adds the (nice) benefit of setting display order.
```

ttannis:
Hey @samshuster below is a screenshot for what our designer would like to see, and what you have is basically almost there. The design concepts we want to maintain are:

1. The programmatic descriptions will be in their own section which will live at the bottom left hand panel. The reason is that since this information can be any number of descriptions, we would not want them to push down all other metadata, especially things related to quality features.
2. This programmatic description section will have header as seen the mocks to make it clear the the information is read only and auto-generated.
3. Since we do not know what the data in these descriptions will be, each description will take up the full width of the panel.
4. Maintain existing section header styles, more specifically here the fact that we use "Title Case", which means `s3 crawler` header should be rendered as `S3 Crawler`

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -63,9 +65,50 @@ def marshall_table_full(table_dict: Dict) -> Dict:
...
+
+    # We follow same style as column stat order for arranging the programmatic descriptions
+    prog_descriptions = results['programmatic_descriptions']
+    if prog_descriptions:
+        _update_prog_descriptions(prog_descriptions)
+
     return results
 
 
+def _update_prog_descriptions(prog_descriptions: List) -> None:
```
ttannis: I'd like to avoid habits of further processing  information for API calls, because any time the frontend service has to do this the immediate question is "For what reason does the API designed for this feature not return the exact information that frontend service needs for this feature". 

I recognize we have a need for a human-friendly display name in the UI for each of these section. However it is not a scalable choice for this to be something that has to be configured. (I know this pattern currently exists with badges but that was a workaround for the feature that will be difficult to maintain in environments with  many badges). Rather I think either:
1. Treat the title as piece of metadata itself, and it would be included in the payload returned from the metadata service. 
2. The React component that this displays this information would be responsible for transforming the `source` to a `title` according to rules for a particular pattern. If it becomes difficult to write logic for this, then to me would be that there needs to be some consistent patterns in how `source` is defined in the backend. As an example, see how we created a `Flag` component that has logic to determine how it should render it's text: https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/components/common/Flag/index.tsx#L18

I recognize that we intend to order this data. I'd like to understand more about why the list returned from metadata service needs to the data being ingested in such a ways that we can us the feature as is. I think about `columns` and the fact that those are passed to the frontend with a `sort_order`, and so I'm concerned abut diverging from that and having to maintain extra logic here.

amundsen_application/static/js/components/TableDetail/NonEditableSection/index.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
```
ttannis: I would prefer to allow `EditableSection` to have a `readOnly` property that will just affect whether or not the edit button is displayed versus creating a whole new component that is exactly the same except for editing.

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -211,6 +213,21 @@ class TableDetail extends React.Component<TableDetailProps & RouteComponentProps
                   </EditableSection>
                 </section>
               </section>
+              {
+                Array.from(data.programmatic_descriptions.values())
+                  .map((d: ProgrammaticDescription) =>
```
ttannis: 1. If `data.programmatic_descriptions` is an array, why not `data.programmatic_descriptions.map(...)`?

2. Do yo actually need to import `ProgrammaticDescription` and insert the type here? Does TypeScript not implicitly recognize `d`  is a `ProgrammaticDescription` because we declared the type for this array already in the `TableMetadata` interface?

amundsen_application/static/js/ducks/tableMetadata/api/tests/index.spec.ts
```diff
@@ -57,6 +57,9 @@ describe('helpers', () => {
   it('getTableDataFromResponseData', () => {
     Helpers.getTableDataFromResponseData(mockResponseData);
     expect(filterFromObjSpy).toHaveBeenCalledWith(tableResponseData, ['owners', 'tags']);
+    expect(Helpers.getTableDataFromResponseData(mockResponseData)).toMatchObject(
```
ttannis: Unit tests should test the purpose of a method, and the purpose of `getTableDataFromResponseData` is to filter `owners` and `tags` from the object. Since that purpose has nothing to do with `programatic_descriptions`, I would suggest removing this check.

samshuster:
hey @ttannis thanks for the great feedback!
Will be going over this later today and tomorrow

amundsen_application/static/package.json
```diff
@@ -88,6 +88,7 @@
     "axios": "0.19.0",
     "form-serialize": "^0.7.2",
     "jquery": "^3.4.0",
+    "mkdirp": "^1.0.3",
```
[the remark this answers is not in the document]
samshuster: done

amundsen_application/static/js/components/TableDetail/index.tsx
samshuster: thanks, done

amundsen_application/static/js/ducks/tableMetadata/api/tests/index.spec.ts
samshuster: got it. Moved to its own test.

amundsen_application/api/utils/metadata_utils.py
samshuster: Ok thanks for that detailed writeup!

I decided not to include sort_order of programmatic_descriptions in the backend because the difference between sort_order of columns and sort_order of programmatic descriptions in my opinion is that one is actually real metadata and the other only pertains to presentation layer and has no actual real value. My thought was we should try and keep anything that isn't actually real metadata out of the backend if possible. This is also why "display_title" did not get put in the backend. Another example of something that followed this same design decision is column_stat_order, which is also configurable in config.py. currently.

Based on what you said, does this mean that column_stat_order would also be deprecated? I would imagine that column_stat_order would have a higher performance impact than programmatic_descriptions as it has to be performed for every column.

If these are all being deprecated, does this mean that we will be pushing all of this sort of "presentation configuration" in the backing db? Or is there another place that is recommended?

Here are a few options that I can think of, but I am sure that you guys have other ideas too:

1. Just remove it and only provide the default behavior for now (alphabetical sort, display title is just transformed (although is this more efficient than just a lookup in a config map?). Revisit in a different issue.
2. Backend changes I believe that would require that we normalize description sources and have a new "Description Source" node that would then contain these properties that could be easily changed without having to make changes to all the description nodes.
3. Relocate this configuration to the metadata_service? Not sure if thats any better.
4. Come up with a more efficient way to configure it on frontend?

In the meantime, I will look into at least modifying the display title like you mentioned for the Flag component.

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -63,9 +65,50 @@ def marshall_table_full(table_dict: Dict) -> Dict:
...
+    # We want to make sure there is a display title that is just source
+    for desc in prog_descriptions:
+        source = desc.get('source')
+        if source:
+            desc['display_title'] = source.replace('_', ' ')
```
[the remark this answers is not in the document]
samshuster: ok I will modify to title

amundsen_application/static/js/components/TableDetail/NonEditableSection/index.tsx
samshuster: ok sounds good, I struggled to get that to work, let me take another stab at it.

amundsen_application/static/js/ducks/tableMetadata/api/tests/index.spec.ts
ttannis: It's not that it needs to be modified, more so that I don't see what needs to be tested based on the changes you're made. I'm not seeing any changes to the logic of `getTableDataFromResponseData`.

amundsen_application/static/js/components/TableDetail/NonEditableSection/index.tsx
samshuster: figured out one possible solution I think!

amundsen_application/static/js/ducks/tableMetadata/api/tests/index.spec.ts
samshuster: ok, I guess I wanted to make sure that programmatic_descriptions were being returned in case somebody removed it from that method, but I could see how static typing takes care of that concern for us? I will remove it then.

amundsen_application/api/utils/metadata_utils.py
ttannis: Column stat order would not be deprecated, but consider that programmatic descriptions are more like columns because they have no limit to their size. Meanwhile column stats is a fixed set  of values. So if we're grouping feature patterns based on characteristics, programmatic descriptions would be more akin to columns than the fixed set of column stats.

Another concept is that while I think there's value in separating what is metadata vs presentational information, we have to make design choices that are scalable. If we were dealing with a fixed set of values my thoughts would be different.

So now to focus on solutions:

1. I think I would be in favor of the alphabetical sort. Given that I liken these descriptions to columns, we do have a plan in our designs to put a sorting UI above the columns, which we could then put in for programmatic descriptions as well! However do you have use case for needing something other than alphabetical right now?

2. In terms of the display `title` being something that is transformed from the `source`, I'm currently not concerned about performance (given that it would take a LOT of data for this to have any noticeable affect on modern browsers). What I want to avoid is designing a feature where users will have to update the configuration every time they want to add a new description. Rather, if the information is ingested with certain rules in play -- such as the `source` following some common pattern like "snake_case" or "kebab-case"  -- then the frontend service will be able to transform new information for free.

amundsen_application/static/js/components/TableDetail/NonEditableSection/index.tsx
ttannis: Great. Feel free to ping me on slack if you need further assistance on the React components.

amundsen_application/api/utils/metadata_utils.py
samshuster: > What I want to avoid is designing a feature where users will have to update the configuration every time they want to add a new description. Rather, if the information is ingested with certain rules in play 
Oh! To clarify, the configuration is totally optional. By default, the programmatic_descriptions use the "description_source" as the title (I can make sure that it has the transform logic in place) and alphabetical sort. They don't need to be specified in the config file. These are just optional overrides. I put more information about that per @jornh suggestion in this code review in the docs/flask_config.py file. Maybe that will help clarify? If that is the case, would you be ok keeping the title overrides, or simpler to just get rid of all of it?

> Meanwhile column stats is a fixed set of values. 
Oh I thought that column stats were also unbounded as well?

> I think I would be in favor of the alphabetical sort. Given that I liken these descriptions to columns, we do have a plan in our designs to put a sorting UI above the columns, which we could then put in for programmatic descriptions as well! However do you have use case for needing something other than alphabetical right now?

Ok, we can definitely just keep it as only alphabetical sort for now.
In the future though, it would be nice to have control over the order of the descriptions because there are some reports we would like to be more immediately viewable, for example a quality badge about the dataset. Whereas, we have another job that populates some embedded graphs that can get quite long, so we would put that towards the end. I can't necessarily see it being useful to customize the sort differently in the UI for programmatic descriptions, but could be totally wrong about that!

Here is a screenshot that sort of demonstrates it, but not quite:

Screen Shot 2020-03-11 at 3 45 17 PM

Ideally we would want the charts to be at the end so that people don't have to scroll forever to get to other things. But it is not the end of the world and I am sure there is a better solution out there.

amundsen_application/static/js/components/TableDetail/EditableSection/tests/index.spec.tsx
```diff
@@ -48,6 +49,7 @@ describe("EditableSection", () => {
     const customTitle = "custom title";
     const { wrapper, props } = setup({ title: customTitle }, <TagInput/>);
 
+    // TODO how do I test for readOnly?
```
samshuster: heads up, I should still add a test for the readOnly=true case if I can figure it out! Will probably need to reach out to you on slack

amundsen_application/api/utils/metadata_utils.py
ttannis: I see. The configuration being optional makes things less painful and understanding that your use case deals with things like charts helps. I no longer think I mind having an optional sort order configuration, because I don't see a better way to address that use case. 

However for the display `title`, I currently still find it reasonable to expect some consistency in how what `source` is expected to look like. I think these sorts of consideration creates a more unified product, because in the worst case scenario if there is no consistency, then the optional configuration no longer becomes optional. 

We don't have to address everything in one go though, how about ping me other things are good to go and then we can re-evaluate. As a note if we do end up keeping the displayName configuration, with the existing precedents we would move that to the application configuration versus the Flask configuration. So if anything the Flask configuration would just have some concept of sort order and the application configuration would have the concept of the displayName

amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx
```diff
@@ -4,6 +4,7 @@ import './styles.scss';
 
 export interface EditableSectionProps {
   title: string;
+  readOnly: boolean;
```
ttannis: Lets make this optional with `readOnly?: boolean`. It will allow the default version of this component to be editable and we only have to throw `readOnly={true}` on the new non-editable cases.

amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx
```diff
@@ -16,6 +17,7 @@ export interface EditableSectionChildProps {
 }
 
 export class EditableSection extends React.Component<EditableSectionProps, EditableSectionState> {
+
```
ttannis: nit: don't need the extra line

amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx
```diff
@@ -24,6 +26,7 @@ export class EditableSection extends React.Component<EditableSectionProps, Edita
     }
   }
 
+
```
ttannis: nit: don't need the extra line

amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx
```diff
@@ -42,17 +45,27 @@ export class EditableSection extends React.Component<EditableSectionProps, Edita
         setEditMode: this.setEditMode,
       });
     });
-
+    if (!this.props.readOnly) {
```
ttannis: Notice how the only real difference here is whether or not the button is shown, so we can do:
```
return (
      <section className="editable-section">
        <div className="section-title title-3">
          { this.props.title }
          {
            !this.props.readOnly &&
            <button className={"btn btn-flat-icon edit-button" + (this.state.isEditing? " active": "")} onClick={ this.toggleEdit }>
              <img className={"icon icon-small icon-edit" + (this.state.isEditing? " icon-color" : "")} />
            </button>
           }
        </div>
        { childrenWithProps }
      </section>
    );
````

Comment (samshuster) [reply amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx]: ah ok thanks for pointing that out! fixing

File: amundsen_application/static/js/components/TableDetail/EditableSection/tests/index.spec.tsx (hunk 8 above)
Comment (ttannis) [reply amundsen_application/static/js/components/TableDetail/EditableSection/tests/index.spec.tsx]: It looks like this test suite currently didn't test for the button to be rendered. What we'd want to do is create a case where we have `readOnly` as false and verify the edit button exists, any another where we have `readOnly` as true and verify that the edit button doesn't exist.

File: amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx (hunk 9 above)
Comment (samshuster) [reply amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx]: ah ok thats how you do that :) great, fixing.

Comment (samshuster) [reply amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx]: do I need to set equal to false by default in the constructor? or will that be the default value?

Comment (samshuster) [reply amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx]: i believe the answer is that it is false by default. Everything appears to work ok as is, so resolving.

File: amundsen_application/static/js/components/TableDetail/EditableSection/tests/index.spec.tsx (hunk 8 above)
Comment (samshuster) [reply amundsen_application/static/js/components/TableDetail/EditableSection/tests/index.spec.tsx]: I think i got it up and running with 2 tests now. Did this:
```    it("renders button when readOnly=false", () => {
      expect(wrapper.find(".btn").length).toEqual(1);
    });

    it("does not render button when readOnly=true", () => {
      const { wrapper } = setup({readOnly: true}, <TagInput/>);
      expect(wrapper.find(".btn").length).toEqual(0);
    });```

File: amundsen_application/static/js/components/common/Flag/index.tsx [hunk 13]
```diff
@@ -20,7 +20,7 @@ export function convertText(str: string, caseType: string): string {
     case CaseType.LOWER_CASE:
       return str.toLowerCase();
     case CaseType.SENTENCE_CASE:
-      return `${str.charAt(0).toUpperCase()}${str.slice(1).toLowerCase()}`;
+      return str.split(new RegExp('\\s+')).map(x => x.charAt(0).toUpperCase() + x.slice(1).toLowerCase()).join(" ");
```
samshuster: Should I create a new CaseType for PascalCase? Or is it ok modifying SENTENCE_CASE to produce:

"test example" -> "Test Example"

samshuster:
Ok @ttannis thank you so much for your feedback, learning a lot!

At this point I believe I am ready for another review. I have also made the design changes per the recommendations. Please find a screenshot of what it looks like now:

Screen Shot 2020-03-12 at 8 16 20 AM

Per our long conversation about optional configuration. I have include the Flag convertText function by default.

amundsen_application/static/js/components/common/Flag/index.tsx
ttannis: I hadn't intended for this component to be modified. Rather this was an example of what `EditableSection` could do to create the `title` to ensure consistent titles in `Title Case`. The concept here is that:
1. We have `source` passed in from the backend, and an optional configuration for the display name
2. The UI requirement is that the `EditableSection` titles be rendered in `Title Case`
3. I want to decrease the need for configuring the display name, and one way to do this is to establish/expect a consistent pattern for `source` such that `EditableSection` can take values like "quality_service" and "s3_crawler", then automatically render "Quality Service" and "S3 Crawler" as the title without the engineer needing to configure anything.

samshuster: Ok so I think I have addressed the main concerns now:
- no more configurable display title
- source is used and made to conform to requirements

The only thing I am unsure if you are ok is if I can reuse that convert text function with a new case. Or would you prefer I just add a new function in the EditableSection index?

ttannis: How about add a new function and we can clean it up later

amundsen_application/static/js/components/TableDetail/EditableSection/tests/index.spec.tsx
```diff
@@ -49,7 +50,7 @@ describe("EditableSection", () => {
     const { wrapper, props } = setup({ title: customTitle }, <TagInput/>);
 
     it("sets the title from a prop", () => {
-      expect(wrapper.find(".section-title").text()).toBe(customTitle);
+      expect(wrapper.find(".section-title").text()).toBe("Custom Title");
```
samshuster: Note, I am currently applying to all section-titles in EditableSection. Is this ok?

samshuster:
ok @ttannis it is ready for you to take another look at it.

tests/unit/api/metadata/test_v0.py
```diff
@@ -312,6 +370,85 @@ def test_get_table_metadata_success(self) -> None:
             self.assertEqual(response.status_code, HTTPStatus.OK)
             self.assertCountEqual(data.get('tableData'), self.expected_parsed_metadata)
 
+    @responses.activate
+    def test_get_table_programmatic_metadata_success(self) -> None:
```
ttannis: Instead of needing a new test and a new object, you could add `programmatic_descriptions` to `self.mock_metadata` and `self.expected_parsed_metadata`. Beyond that I think the new tests you'd have to write wouldn't be in this file. 

The new functionality you added can be summarized as:
1. `_marshall_table_full` updates the `programmatic_descriptions` with ` _update_prog_descriptions`
2. ` _update_prog_descriptions` acts updates `progrmmatic_descriptions` according to certain criteria
3. `_sort_prog_descriptions` returns a order value according to certain criteria

Since the idea is to unit test new functionality I would expect to see tests for these new methods. However I notice we are lacking a test file for `metadata_utils`. Overall I'd say we could add `programmatic_descriptions` to the test input and output for `test_get_table_metadata_success`, but beyond that we would want to just make a `test_metadata_utils.py` in `tests/unit/utils` and add a test for those methods you added.

amundsen_application/static/js/components/TableDetail/EditableSection/tests/index.spec.tsx
ttannis: Yes, that's the intent to have a title that looks consistent.

amundsen_application/static/js/components/TableDetail/EditableSection/tests/index.spec.tsx
```diff
@@ -65,5 +66,18 @@ describe("EditableSection", () => {
       const { wrapper } = setup(null, child);
       expect(wrapper.childAt(1).text()).toBe(child);
     });
+
+    it("renders button when readOnly=false", () => {
+      expect(wrapper.find(".btn").length).toEqual(1);
```
ttannis:
```suggestion
      expect(wrapper.find(".edit-button").length).toEqual(1);
```
All buttons in this application have the class `.btn`, so I'm suggesting the className specifically for the edit button in case there is ever another button.

amundsen_application/static/js/components/TableDetail/EditableSection/tests/index.spec.tsx
```diff
@@ -65,5 +66,18 @@ describe("EditableSection", () => {
...
+    });
+
+    it("renders does not add button when readOnly=true", () => {
+      const { wrapper } = setup({readOnly: true}, <TagInput/>);
+      expect(wrapper.find(".btn").length).toEqual(0);
```
ttannis:
```suggestion
      expect(wrapper.find(".edit-button").length).toEqual(0);
```
All buttons in this application have the class `.btn`, so I'm suggesting the className specifically for the edit button in case there is ever another button.

amundsen_application/static/js/components/TableDetail/EditableSection/index.tsx
```diff
@@ -15,6 +16,10 @@ export interface EditableSectionChildProps {
   setEditMode?: (isEditing: boolean) => void;
 }
 
+export function convertText(str: string): string {
```
ttannis: nit: This can be a static method on the `EditableSection` class.

amundsen_application/static/js/components/TableDetail/EditableSection/tests/index.spec.tsx
```diff
@@ -65,5 +66,18 @@ describe("EditableSection", () => {
...
+    });
+
+    it( 'renders modifies title to have no underscores', () => {
```
ttannis: This unit test suffers from leveraging knowledge of how `convertText` works. Ideally the component is only concerned with checking that it called `convertText` with the correct parameters. If you aren't familiar with mocking functions in Jest,a fter this is merged I think I'll update this test in particular as an example that you can reference in the future.

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -63,9 +65,40 @@ def marshall_table_full(table_dict: Dict) -> Dict:
...
+
     return results
 
 
+def _update_prog_descriptions(prog_descriptions: List) -> None:
+    # We want to make sure there is a display title that is just source
+    for desc in prog_descriptions:
+        source = desc.get('source')
+        if not source:
+            logging.error("no source found in: " + str(desc))
```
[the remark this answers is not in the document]
samshuster: ok sounds good.

amundsen_application/static/js/components/TableDetail/EditableSection/tests/index.spec.tsx
samshuster: ok makes sense!

amundsen_application/static/js/components/TableDetail/EditableSection/tests/index.spec.tsx
samshuster: ok, yeah I just copied the test style from Flag. I am not familiar with jest. Will be a good learning experience to see that example in the future, thanks!

tests/unit/api/metadata/test_v0.py
samshuster: Alright, here is what I did, hopefully this is pretty much in line with your recommendations:

- I added programmatic_descriptions to mock_data
- I removed the two separate programmatic_description tests from test_v0
- I added a new test_metadata_utils file and added a single test for marshall_table_full at this time. My thought is that by testing this method, it tests the other private methods.

docs/flask_config.md
```diff
@@ -43,3 +43,36 @@ Here are the settings and what they should be set to
...
+           },
+           "doesnt_exist": {
+               "display_order": 2
+           }
+       }
+```
+description sources not mentioned in the configuration will be alphabetically placed at the end of the above list. If `PROGRAMMATIC_DISPLAY` is left at `None` all added fields are still showing up, so that display is entirely dynamically data-driven without configuration. Meaning configuration merely adds the (nice) benefit of setting display order.
+
+Here is a screenshot of what it would look like in the bottom left here:
+![programmatic_description](img/programmatic_descriptions.png)
```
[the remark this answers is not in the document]
samshuster: thanks for catching this!

tests/unit/api/metadata/test_v0.py
```diff
@@ -75,7 +79,7 @@ def setUp(self) -> None:
                 'name': 'test_name',
                 'id': 'test_id',
                 'description': 'This is a test'
-            },
```
ttannis: Trailing commas are okay in python and for a couple of nuanced reasons can be desirable. No need to put them back in (we should have a linter handle this at some point), but also no need to remove them in the future:
https://stackoverflow.com/questions/11597901/why-are-trailing-commas-allowed-in-a-list

tests/unit/utils/test_metadata_utils.py
```diff
@@ -0,0 +1,115 @@
+import unittest
+
+from amundsen_application.api.utils.metadata_utils import marshall_table_full
+from amundsen_application import create_app
+
+local_app = create_app('amundsen_application.config.TestConfig', 'tests/templates')
+
+
+class MetadataUtilsTest(unittest.TestCase):
```
ttannis: This isn't exactly what I meant but I can modify this after merging and tag you for reference. Ideally, we would have already had a `test_metadata_utils` set up with at least one example to make things easier.

tests/unit/api/metadata/test_v0.py
ttannis: > * My thought is that by testing this method, it tests the other private methods.
 While that is true, it's a good practice for unit tests to isolate logic and explicitly test behaviors. For example if method `A()` calls method `B()` and returns some output `C`, we *could* call method `A()`, check that `C` is returned and consider both `A()` and `B()` tested. 

However with this approach if the logic in `B()` ever changes then the test fails.  Then the question becomes why should a change in  the internal logic for `B()` be able to break a unit test for `A()`? We would address this my mocking `B()` with some fake logic in the unit test. I will tag you in a PR that demonstrates this. 

Thanks for your work on this feature!

PR #382, opened 2020-01-28
use schemas for metadata retrieval & update search field names

friendtocephalopods:
### Summary of Changes

Use schemas for RPC between metadatastore and frontend. Use search field names consistent with amundsencommon models.

Should be merged/deployed along with https://github.com/lyft/amundsenmetadatalibrary/pull/105 and https://github.com/lyft/amundsensearchlibrary/pull/81.

### Tests

Tweak tests to expect schema names for metadata and search responses. Use consistent naming in frontend and common.

### Documentation

None.

### CheckList


amundsen_application/api/utils/metadata_utils.py
```diff
@@ -1,73 +1,44 @@
...
-        'schema_name': schema_name,
-        'type': 'table',
-        'last_updated_epoch': table.get('last_updated_epoch', None),
-    }
+    schema = PopularTableSchema(strict=True)
+    # TODO: consider migrating to validate() instead of roundtripping
+    table: PopularTable = schema.load(table_dict).data
+    results = schema.dump(table).data
+    # TODO: fix popular tables to provide these? remove if we're not using them?
+    results['last_updated_timestamp'] = None
```
danwom: This might be out of scope but I think we should add the timestamp and type to `PopularTable` and rename it as `TableSummary`. Then we can use this for both popular tables, bookmarks, search results, etc. `TableSummary` would be the expected schema for anytime we need to display a `TableListItem` component and `Table` would be used to display the full details on the `table_detail` page.

If not I can do this in a follow up, as this has been on my personal todo list for quite some time.

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -1,73 +1,44 @@
...
+    results['type'] = 'table'
```
danwom: We actually need to preserve the `key` here
```
    # TODO: Add the 'key' or 'id' to the base PopularTableSchema
    results['key'] = '{0}://{1}.{2}/{3}'.format(table.db, table.cluster, table.schema_name, table.table_name)
```

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -1,73 +1,44 @@
...
-        'watermarks',
-        # 'last_updated_timestamp' Exists on the response from metadata but is not used.
-        # This should also be consolidated with 'last_updated_epoch' to have the same name and format.
-    ]
 
-    results = {field: table.get(field, None) for field in fields}
+    schema = TableSchema(strict=True)
+    # TODO: consider migrating to validate() instead of roundtripping
+    table: Table = schema.load(table_dict).data
+    results: Dict[str, Any] = schema.dump(table).data
```
danwom: Same thing here. Ideally we would add some `id` to all resource types in the base schema.
```
    # TODO: Add the 'key' or 'id' to the Table schema
    results['key'] = '{0}://{1}.{2}/{3}'.format(table.db, table.cluster, table.schema_name, table.table_name)
```

amundsen_application/api/utils/metadata_utils.py
jinhyukchang: @friendtocephalopods FYI, this is currently breaking UI and blocker of `v2.0.0`

amundsen_application/api/utils/metadata_utils.py
friendtocephalopods: I had the start of that here: https://github.com/lyft/amundsencommon/pull/14/files but we'd need to add it to popular tables as well.

amundsen_application/api/utils/metadata_utils.py
friendtocephalopods: Oh! Good catch! I haven't quite pushed 'key' as an attribute to the common models yet. Let me replace the key...

tests/unit/api/metadata/test_v0.py
```diff
@@ -20,11 +20,10 @@ def setUp(self) -> None:
...
                     'cluster': 'test_cluster',
                     'database': 'test_db',
-                    'key': 'test_db://test_cluster.test_schema/test_table',
                     'schema': 'test_schema',
-                    'table_description': 'This is a test',
-                    'table_name': 'test_table',
-                    'type': 'table',
+                    'description': 'This is a test',
+                    'name': 'test_table',
+                    'key': 'test_db://test_cluster.test_schema/test_table',
```
danwom: I think this should have both `type` and `key`

0001-SQ-tweaks.patch
```diff
@@ -0,0 +1,65 @@
+From e905c1c06ebb1dbd09f16cb7c81765ccc9c9f3a4 Mon Sep 17 00:00:00 2001
```
danwom: Was this added by mistake?

friendtocephalopods: /facepalm

tests/unit/api/metadata/test_v0.py
friendtocephalopods: ah, forgot to add the key to expected parsed user resources.

amundsen_application/api/utils/metadata_utils.py
ttannis: I agree with this. You could file an issue in the umbrella repo as a "help wanted" ticket.

requirements.txt
```diff
@@ -62,3 +62,7 @@ marshmallow==2.15.3
 # License: Apache 2.0
 # Upstream url: https://pypi.org/project/responses/
 responses==0.9.0
+
+amundsen-common==0.1.7
+marshmallow==2.15.3
```
ttannis: 1. We already have `marshmallow==2.15.3` on line 59
2. Could we add similar comments like we have listed above for `marshmallow-annotations`

ttannis:
My comments are non-blocking for this PR.

requirements.txt
friendtocephalopods: Oh! Good catch!

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -1,73 +1,46 @@
...
-        'name': table_name,
-        'schema_name': schema_name,
-        'type': 'table',
-        'last_updated_epoch': table.get('last_updated_epoch', None),
-    }
+    schema = PopularTableSchema(strict=True)
+    # TODO: consider migrating to validate() instead of roundtripping
+    table: PopularTable = schema.load(table_dict).data
+    results = schema.dump(table).data
+    # TODO: fix popular tables to provide these? remove if we're not using them?
```
danwom: `./amundsen_application/api/utils/metadata_utils.py:20: [W291] trailing whitespace`

danwom: Hopefully everything will pass the build after this :)

danwom:
@friendtocephalopods I have fixes in two commits [in a new branch](https://github.com/lyft/amundsenfrontendlibrary/commits/use-schema-for-table-api). Can you add these to this PR? I'm not sure how to add commits to this PR directly. 

Basically we get different subsets of the user schema from various sources (user metadata, user search, table metadata owners/readers) and we need to sanitize the data. As a longer term fix we should enforce each of those responses to send a complete user object. Issue: https://github.com/lyft/amundsen/issues/296

friendtocephalopods:
> @friendtocephalopods I have fixes in two commits [in a new branch](https://github.com/lyft/amundsenfrontendlibrary/commits/use-schema-for-table-api). Can you add these to this PR? I'm not sure how to add commits to this PR directly.
> 
> Basically we get different subsets of the user schema from various sources (user metadata, user search, table metadata owners/readers) and we need to sanitize the data. As a longer term fix we should enforce each of those responses to send a complete user object. Issue: [lyft/amundsen#296](https://github.com/lyft/amundsen/issues/296)

Ah, thanks for the commits! Is it worth trying to fix the search/store mismatches for User before 2.0? It might be easier to fix it now when we're already going to need a coordinated deploy.

danwom:
@friendtocephalopods That's a good point. In an ideal world I think we would be able to resolve all of the tech debt before launching 2.0, but I think it may be better to unblock this launch instead of snowballing more changes. At the very least we can start adding additional metadata to the user objects upstream and FE *should* graciously accept them as-is.

friendtocephalopods:
That's true, as long as we get the search changes in ahead of

amundsen_application/api/search/v0.py
```diff
@@ -186,8 +186,8 @@ def _map_table_result(result: Dict) -> Dict:
             'cluster': result.get('cluster', None),
             'description': result.get('description', None),
             'database': result.get('database', None),
-            'schema_name': result.get('schema_name', None),
-            'last_updated_epoch': result.get('last_updated_epoch', None),
+            'schema': result.get('schema', None),
+            'last_updated_timestamp': result.get('last_updated_timestamp', None),
```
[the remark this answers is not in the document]
danwom: In that case we would be better off renaming these to `epoch` instead of `timestamp`. I think there was another conversation elsewhere and we ended up going with `timestamp`. I don't recall seeing any non-epoch timestamps so far.

requirements.txt
```diff
@@ -58,7 +58,14 @@ requests==2.20.0
...
+# License: MIT
+# Upstream url: https://github.com/justanr/marshmallow-annotations
+marshmallow-annotations==2.4.0
+
 # A utility library for mocking out the requests Python library.
 # License: Apache 2.0
 # Upstream url: https://pypi.org/project/responses/
 responses==0.9.0
+
+amundsen-common==0.1.7
```
jinhyukchang: Could you change this to `0.2.2`?

jinhyukchang:
@friendtocephalopods Somehow the test is failing in Py3. Could you take a look?

PR #383, opened 2020-01-28
Update search flow (some more)

ttannis:
### Summary of Changes

This PR introduces a variety fo cleanup changes:
1. Add a Breadcrumb on HomePage that makes filters discoverable
image

2. updateSearchUrl` allows for navigation to just `/search`. If there is no search term or filters in the `searchParams` we now go to `/search`. I think this was the best place to fix a weird state the url would get into when someone used filters with no search term, then cleared all the filters. 

3. Allow for `InfoButton` to use SanitizedHTML. This was because product requested specfic formatting in the tooltips.

4. Not requiring the search term for search bar submit created some edge case issues. Specifically if someone deleted their search term in favor of entering filters but didnt hit enter, that search term would still apply but the UI  did not reflect that the search term was still in play. To fix this, I reverted the change to allow search bar submission with no search term. Instead:
* Updating the filters with an empty search term will still work and submit a search
* When a user deletes the entire search term (or clicks the 'X' ) on the search page, a search will automatically re-execute without the search term.
* The term is also optional in the url now, and made changes to `urlDidUpdateWorker` and `updateSearchUrl` to reflect this.

5. Fixed a bug caused by `urlDidUpdateWorker` evaluating the filters before it evaluated which resource to use.

### Tests

TODO on final feature branch review

### Documentation

TODO on final feature branch review

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/HomePage/index.tsx
+++ b/amundsen_application/static/js/components/HomePage/index.tsx
@@ -7,6 +7,7 @@ import { RouteComponentProps } from 'react-router';
 import './styles.scss';
 
 import MyBookmarks from 'components/common/Bookmark/MyBookmarks';
+import Breadcrumb from 'components/common/Breadcrumb';
 import PopularTables from 'components/common/PopularTables';
 import { SearchAllReset } from 'ducks/search/types';
 import { searchReset } from 'ducks/search/reducer';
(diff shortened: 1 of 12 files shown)


ttannis:
@danwom you don't have to review this PR directly because I actually just want to evaluate these changes within the full context of the feature branch and might change some things afterward. 

However I created this PR to give some context about the changes for you to reference in the final review, or if there's something that stands out to you right now that you want to comment on. 

Barring any major changes from design or serious bugs, I will now focus on cleaning up `feature/ui-filter`

PR #386, opened 2020-01-30
Added last_updated_timestamp to table details page

danwom:
### Summary of Changes

Added dateUtils function to format dates. Added 'last_updated_timestamp' to table details page.

### Tests

### Documentation

### CheckList


amundsen_application/api/utils/metadata_utils.py
```diff
@@ -42,6 +42,8 @@ def marshall_table_full(table: Dict) -> Dict:
         'database',
         'is_view',
         'key',
+        # TODO - Consolidate with 'last_updated_epoch' to the same name and format.
+        'last_updated_timestamp',
```
danwom: @jinhyukchang @feng-tao where do we stand with `epoch` vs `timestamp`? If I know what direction we're going, I can re-write FE code accept both formats and rename the field in Flask, then React can always refer to the correct name until this is renamed in the other services.

amundsen_application/static/js/components/Footer/index.tsx
```diff
@@ -29,8 +30,7 @@ export class Footer extends React.Component<FooterProps> {
   }
 
   generateDateTimeString = () => {
-    // 'moment.local' will utilize the client's local timezone.
-    return moment.unix(this.props.lastIndexed).local().format('MMMM Do YYYY [at] h:mm:ss a');
+    return formatEpochTime(this.props.lastIndexed, DATE_FORMAT_LONG);
```
danwom: @ttannis is there something I should be doing here with local timezones?

ttannis: Depends on the logic of `formatEpochTime`. The idea here is that users must see this time according to their local timezone to avoid confusion. For example if the actual time was 6AM pacific time, users in New York should ideally see this as 9AM eastern time.

danwom: I'm just using the `moment` constructor. I tested it while spoofing my location and it appears to update the timezone. Maybe the `moment.unix` function we were using before was trying to reset to GMT or something.

ttannis: Ah, so I think this was to account for daylight savings time as well.  I think the original logic could be summarized as "convert to GMT then transform to local timezone".  Whatever methods we use here just has to account for that so people don't get confused.

danwom:
I will wait until https://github.com/lyft/amundsenfrontendlibrary/pull/382 is merged in.

ttannis:
@danwom we may have some conflicting priorities is we wait for #382. 

This is surrounding the concept of us aiming to add badges as a filter.  I think this feature will be most effective if we complete the mocks as intended and show badges in the search results, meaning we'd need to complete putting this on the table detail page before then. Let me know what you think, but I'm now leaning towards us completing the task so we can unblock badges in search results.

danwom:
Looks like there hasn't been any update for that. I'm fine with merging this in first then after getting approval.

ttannis:
> Looks like there hasn't been any update for that. I'm fine with merging this in first then after getting approval.

Thanks! Definitely prefer to just unblock getting the badges in search results given our recent planning so that we can hopefully get that and badge filtering in the next month as well.

amundsen_application/static/js/components/Footer/index.tsx
ttannis: @danwom so does this new method account for daylight saving time and local timezones?

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -184,18 +185,23 @@ class TableDetail extends React.Component<TableDetailProps & RouteComponentProps
                       <WatermarkLabel watermarks={ data.watermarks }/>
                     </>
                   }
+                  {
+                    !!data.last_updated_timestamp &&
+                    <>
+                      <div className="section-title title-3">Last Updated</div>
+                      <div className="body-2">{ formatEpochTime(data.last_updated_timestamp) }</div>
```
ttannis: Our mocks seem to specify a new format: 
image

I'm fine with leaving it as is to complete the task, but I think we should check whether or not this new format was intentionally chosen by Design and understand if this new format is a must-have, nice-to-have, or if it was just placeholder text.

amundsen_application/static/js/utils/dateUtils.ts
```diff
@@ -0,0 +1,9 @@
+import * as moment from 'moment-timezone';
+
+export const DATE_FORMAT = 'MMM DD, YYYY';
+export const DATE_FORMAT_LONG = 'MMMM Do YYYY [at] h:mm:ss a';
```
ttannis: nit: I'd lean towards event more descriptive names because eventually we might support more formats and then it gets less and less intuitive to understand what format is which. For this one I might go with `DATE_TIME_FORMAT_LONG` since it will include the time.

amundsen_application/api/utils/metadata_utils.py
jinhyukchang: Yeah, we settled with timestamp.

cc: @feng-tao

amundsen_application/static/js/utils/tests/index.spec.ts
```diff
@@ -46,3 +47,21 @@ describe('Navigation utils', () => {
     });
   });
 });
+
+
+describe('dateUtils', () => {
+  describe('formatEpochTime', () => {
+    it('formats an epoch time to the default format', () => {
```
ttannis: Can we have some cases that include the time conversion as well since that's something that's important for our application? I assume Jest would output the timezone in `GMT`? I actually don't know and I'm curious if this sort of test would just work as expected.

danwom: Oh man, I vaguely remember adding some test config to specify a timezone to use, but I can't find anything about it anymore. Did we ever figure out this problem or am I just dreaming?

amundsen_application/static/js/components/TableDetail/index.tsx
danwom: This is an easy fix, but this would bring us to 3 unique date formats. I'd prefer to combine this with the one used in the footer. Does anyone really need to know the exact minute and second that the amundsen update job finished (or is that started?)

amundsen_application/static/js/utils/tests/index.spec.ts
ttannis: There was this commit where the change in `package.json` set Jest to run in UTC: https://github.com/lyft/amundsenfrontendlibrary/commit/b3ed6288bcc0dce5e1564d5bfe7ccc3f2aef927e#diff-fc699096242ec6e9acf55ca6bf7ecf09

amundsen_application/static/js/components/TableDetail/index.tsx
ttannis: My personal thought is that seconds may not matter, but I don't feel a need to force us into have the least amount of date formats possible. We could always end up with more for different use cases. In addition the code base could end up including a whole mapping of multiple date formats anyway if the community wants to introduce better internationalization support.

My vote would be to leave the footer as is, but run it by design or users if you feel strongly.

danwom: That's a good point about internationalization. In that case we should create a configs for a date formats and use that instead of putting it in a constants file. Ideally, we'd have just two date formats, maybe three. I'd just hate to see different date formats used randomly in various locations in the same product.

amundsen_application/static/js/components/Footer/index.tsx
danwom: Timezones should be accounted for in this version.

csteez:
> > Looks like there hasn't been any update for that. I'm fine with merging this in first then after getting approval.
> 
> Thanks! Definitely prefer to just unblock getting the badges in search results given our recent planning so that we can hopefully get that and badge filtering in the next month as well. Will take another look by end of day

so i took a look at the jira ticket for badges - are badges still going to replace last updated?

cc @danwom

danwom:
@csteez According to the search results mocks, yes, the badges would show up where the `last updated` currently resides. This PR adds that same information into the `table_detail` page. As a practical matter, I think there's enough real estate to show both `last updated` and `badges` in the search results on wider screens.

amundsen_application/static/js/config/config-types.ts
```diff
@@ -77,6 +79,16 @@ interface BadgeConfig {
   [badge: string]: BadgeStyleConfig;
 }
 
+/**
+ * DateConfig - Configure various date formats
+ *
+ */
+interface DateFormatConfig {
```
ttannis: We can table this idea for if we ever dedicate ourselves to internationalization, but what I think this would ideally look like at scale would be to consider writing code leveraging [Intl.DataTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat/format) and its various options. Then users would simply configure their country code like `en-US`, which would get applied in that logic.

amundsen_application/static/js/utils/dateUtils.ts
```diff
@@ -0,0 +1,41 @@
+import * as Moment from 'moment-timezone';
+import AppConfig from 'config/config';
+
+const timezone = Moment.tz.guess();
+
+interface DateConfig {
```
ttannis: nit: it seems like what we'd need is 3 different types:
```
interface SomeName {
    timestamp: number
}

interface SomeNameEpoch {
    epochTimestamp: number
}

interface SomeNameDateString {
    dateString: string
    dateStringFormat: string
}
```
This is an idea specifically for developer friendliness because since all the values are optional, folks still have to come look at the `getMomentDate` logic to understand how to use it, and they won't get any helpful error messages when compiling if they use this incorrectly. That being said I think this is fine to merge because I'd want to look up the right way to create something like this on TypeScript. One caveat for this idea would be understanding the right way to know which option was passed in. https://stackoverflow.com/questions/14425568/interface-type-check-with-typescript

danwom:
@ttannis tested this with v2 without issue.

PR #389, opened 2020-02-13
Allow users to create JIRA tickets associated with a table

csteez:
### Summary of Changes
Adds connection for Jira in the table detail page. Displays Jira tickets for a given table, and allows users to report data issues for a given table. Users can open a modal to create the tickets in Jira. Displays a configurable amount of tickets at a time. This is on the table detail view. 
Users can create new JIRA tickets associated with a table by clicking "Report Data Issue". Tickets found that contain the table key are automatically displayed on the left hand side. 
### Tests

Added as much test coverage as seems reasonable for code added. 

### Documentation

### CheckList


feng-tao:
do you think we could hide the screenshoot as it shows our internal table name...

feng-tao:
those are all internal infos which we shouldn't show in the open source pr....

csteez:
> do you think we could hide the screenshoot as it shows our internal table name...

Sorry, I removed it

amundsen_application/api/jira/v0.py
```diff
@@ -0,0 +1,85 @@
...
+    Given a table key, returns all JIRA tickets containing that key. Returns an empty array if none exist
+    :return: List of JIRA tickets
+    """
+    try:
+        table_key = get_query_param(request.args, 'key', 'Request requires a key')
+        missing_config_response = validate_jira_properties()
+        if len(missing_config_response) > 0:
+            return make_response(jsonify({'msg': missing_config_response}), HTTPStatus.NOT_IMPLEMENTED)
+
+        jira_client = JiraClient()
```
jinhyukchang: Do you think we can abstract Jira out and place it in proxy like what metadata service or search service have done?
https://github.com/lyft/amundsenmetadatalibrary/tree/master/metadata_service/proxy

It will open the door for other issue tracking system.

amundsen_application/api/jira/v0.py
```diff
@@ -0,0 +1,85 @@
...
+        response = jira_client.search(table_key)
```
jinhyukchang: Does this search method limits the result to 3?

amundsen_application/api/jira/v0.py
jinhyukchang: I wonder if client needs to be created everytime the API is called. 
e.g: https://github.com/lyft/amundsenmetadatalibrary/blob/master/metadata_service/api/table.py#L119

amundsen_application/api/jira/v0.py
```diff
@@ -0,0 +1,85 @@
...
+
+@jira_blueprint.route('/issues', methods=['GET'])
+def get_jira_issues() -> Response:
+    """
+    Given a table key, returns all JIRA tickets containing that key. Returns an empty array if none exist
+    :return: List of JIRA tickets
+    """
+    try:
+        table_key = get_query_param(request.args, 'key', 'Request requires a key')
+        missing_config_response = validate_jira_properties()
```
jinhyukchang: This seems like something JIRA client should validate, or should be validated while JIRA client is being instantiated. WDYT?

amundsen_application/jira/jira_client.py
```diff
@@ -0,0 +1,78 @@
...
+from flask import current_app as app
+
+import logging
+
+SEARCH_STUB = 'project={project_id} AND text ~ "{table_key}"'
+
+
+class JiraClient:
+
+    def __init__(self) -> None:
```
jinhyukchang: How about taking values from app from constructor so that we can have this client decouple from `flask`?

amundsen_application/jira/jira_client.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        Get the Jira client properly formatted prepared for hitting Lyft JIRA
+        :return: A Jira client.
+        """
+        return JIRA(
+            server=self.jira_url,
+            basic_auth=(self.jira_user, self.jira_password)
+        )
+
+    def search(self, table_key: str) -> Any:
```
jinhyukchang: Could we have structured output -- preferably a class? `Any` is really hard to read for other developers.

amundsen_application/jira/jira_client.py
```diff
@@ -0,0 +1,78 @@
...
+                maxResults=3)
+            result = []
+            for issue in issues:
+                result.append(self.get_issue_properties(issue))
+            return result
+        except JIRAError as e:
+            logging.exception(str(e))
+            raise e
+
+    def create_issue(self, description: str, key: str, title: str) -> Any:
```
jinhyukchang: Same comment as above. Shall we avoid `Any`?

amundsen_application/jira/jira_client.py
```diff
@@ -0,0 +1,78 @@
...
+            issue = self.jira_client.create_issue(fields=dict(project={
+                'id': self.jira_project_id
+            }, issuetype={
+                'id': 1,
+                'name': 'Bug',
+            }, summary=title, description=description + '\n Table Key: ' + key))
+
+            return [self.get_issue_properties(issue)]
+        except JIRAError as e:
+            logging.error(str(e))
```
jinhyukchang: Let's `s/error/exception` ?

amundsen_application/jira/jira_client.py
```diff
@@ -0,0 +1,78 @@
...
+            raise e
+
+    @staticmethod
+    def get_issue_properties(issue: Issue) -> Any:
```
jinhyukchang: If I understand correctly, this should be private, right? Shall we add `_` ?
`__get_issue_properties` or if you need it in unit test: `_get_issue_properties`

requirements.txt
```diff
@@ -62,3 +62,6 @@ marshmallow==2.15.3
 # License: Apache 2.0
 # Upstream url: https://pypi.org/project/responses/
 responses==0.9.0
+
+# SDK for integration with Jira
+jira==2.0.0
```
jinhyukchang: Could we have it in `extra_require` in setup.py as this is optional feature?
Example: https://github.com/lyft/amundsenmetadatalibrary/blob/master/setup.py#L24

amundsen_application/api/jira/v0.py
csteez: yeah i didn't think it was a good pattern but i wasn't sure how to maintain "state" in a flask endpoint. do you have any advice on that?

jinhyukchang: Yeah, Flask endpoint can be a instance like this one:
https://github.com/lyft/amundsenmetadatalibrary/blob/master/metadata_service/api/table.py#L113
https://github.com/lyft/amundsenmetadatalibrary/blob/master/metadata_service/__init__.py#L85

csteez: Actually question, does this seem like a good tradeoff? Since the existing app is not restful, and adding in one rest endpoint seems to be sort of weird given that it's not part of a bigger refactor.

amundsen_application/api/jira/v0.py
csteez: yes

amundsen_application/api/jira/v0.py
csteez: I also think that because the table key contains // and / this actually isn't a good pattern to follow - ie  'hive://gold.redshift/dimension_applicants' causes routing errors. 
given that for a get it should be like issues/hive://gold.redshift/dimension_applicants which won't map to a url. Would like to chat in person about this :)

amundsen_application/api/jira/v0.py
csteez: definitely agree - i had moved it out because it seemed like a weird pattern to raise an exception and return an error message instead.

amundsen_application/jira/jira_client.py
csteez: great catch, i switched the name

amundsen_application/jira/jira_client.py
csteez: good catch!

amundsen_application/jira/jira_client.py
csteez: very good point - i made a new class for this

csteez:
> Looks good and left several comments.
> 
> Quick question on search result only showing 3 issues. Are we going to imply that there are more issues such as `...`?

I think given that in a real use case, it's not likely there will be a giant list of tickets. I think maybe it would be best to remove the max results amount for now and then maybe tune it if the use case requires it?

amundsen_application/jira/jira_client.py
```diff
@@ -0,0 +1,78 @@
...
+SEARCH_STUB = 'project={project_id} AND text ~ "{table_key}"'
+
+
+class JiraClient:
+
+    def __init__(self) -> None:
+        self.jira_url = app.config['JIRA_URL']
+        self.jira_user = app.config['JIRA_USER']
+        self.jira_password = app.config['JIRA_PASSWORD']
+        self.jira_project_id = app.config['JIRA_PROJECT_ID']
```
feng-tao: does it mean it could only search issue for single project instead of multiple project for same table?

amundsen_application/jira/jira_client.py
```diff
@@ -0,0 +1,78 @@
...
+        :return: Metadata of matching issues
+        """
+        try:
+            issues = self.jira_client.search_issues(SEARCH_STUB.format(
+                project_id=self.jira_project_name,
+                table_key=table_key),
+                maxResults=3)
+            result = []
+            for issue in issues:
+                result.append(self.get_issue_properties(issue))
```
feng-tao: L41-L43 could be `result = [self.get_issue_properties(issue) for issue in issues]`

feng-tao: or just `return [self.get_issue_properties(issue) for issue in issues]`

amundsen_application/jira/jira_client.py
```diff
@@ -0,0 +1,78 @@
...
+        :param description: Description of the Jira issue
+        :param key: Table key
+        :param title: Title of the Jira ticket
+        :return: Metadata about the newly created issue
+        """
+        try:
+            issue = self.jira_client.create_issue(fields=dict(project={
+                'id': self.jira_project_id
+            }, issuetype={
+                'id': 1,
```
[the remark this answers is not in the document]
csteez: it's the value from jira for a bug "issue type". i think they're standardized (their examples seem to have the same values that i saw returned from the api). i moved that to a constant with a comment.

amundsen_application/jira/jira_client.py
```diff
@@ -0,0 +1,78 @@
...
+    def search(self, table_key: str) -> Any:
+        """
+        Runs a query against a given Jira project for tickets matching the key
+        :param table_key: Table key
+        :return: Metadata of matching issues
+        """
+        try:
+            issues = self.jira_client.search_issues(SEARCH_STUB.format(
+                project_id=self.jira_project_name,
+                table_key=table_key),
```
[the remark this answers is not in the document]
csteez: i think its table uri, if that is dbtype://cluster/tablename. is table_uri a preferred name? i don't see it used much in this code base

amundsen_application/api/jira/v0.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+@jira_blueprint.route('/issues', methods=['GET'])
+def get_jira_issues() -> Response:
+    """
+    Given a table key, returns all JIRA tickets containing that key. Returns an empty array if none exist
+    :return: List of JIRA tickets
+    """
+    try:
+        table_key = get_query_param(request.args, 'key', 'Request requires a key')
```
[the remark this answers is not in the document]
csteez: there isn't any specific logic that ties it only to a table - we're tying the jira ticket to the table by the table uri. so it could be extended to anything with a unique name or id. 

i added parameter descriptions, very good point thank you!

amundsen_application/api/jira/v0.py
```diff
@@ -0,0 +1,85 @@
...
+        missing_config_response = validate_jira_properties()
+        if len(missing_config_response) > 0:
```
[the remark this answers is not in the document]
csteez: i moved the validation into the client itself

amundsen_application/jira/jira_client.py
csteez: very good suggestion, thank you!

amundsen_application/jira/jira_client.py
csteez: hm interesting point - i could remove this entirely and search across all projects. it would only return tickets the service account has access to so that seems probably fine. i removed it from the query string and will remove it as a config setting

amundsen_application/static/js/components/TableDetail/TableIssues/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  logIssueClick = (event) => {
+    logClick(event); 
+  }
+
+  renderIssue = (issue: Issue, index: number) => {
+    return (
+      <div className="issue-banner" key={`issue-${index}`}>
+      <a className="issue-link" target="_blank" href={issue.url} onClick={this.logIssueClick}>
```
danwom: nit: indentation

For `logClick`, we rely try to infer logging metadata. Adding a descriptive `id` such as `table-issue-link-${index}` would help us know what kind of link was clicked. Also you don't need to wrap the `logClick` function unless you have other scripts to run.

amundsen_application/static/js/components/TableDetail/TableIssues/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+        { this.props.issues.map(this.renderIssue)}
+      </div>
+    );
+  }
+}
+
+export const mapStateToProps = (state: GlobalState, componentProps: ComponentProps) => {
+  return {
+    issues: state.issue.issues,
+    tableKey: componentProps.tableKey
```
danwom: Same comment here: can remove `tableKey`.

amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
```diff
@@ -0,0 +1,23 @@
...
+    height: 40px;
+    margin: 8px 0;
+    padding: 8px 16px;
+
+    &:first-child {
+      margin-top: 24px;
+    }
+  }
+  .icon-data-quality-warning {
+    background: #DB3615;
```
danwom: Can we use a color variable here? 

Our color palette is defined in `_colors.scss` (see: $sunset60)
But we try not to reference colors directly. Instead we define a color-role in `_variables-default.scss` and then assign that to the color palette value. For example you could use `$alert-color: $sunset60;`

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -167,6 +169,8 @@ class TableDetail extends React.Component<TableDetailProps & RouteComponentProps
                 TODO - Add a banner here if necessary
                 <section className="banner">optional banner</section>
               */}
+              <TableIssues tableKey={ this.key }/>
```
danwom: nit: We can remove the TODO above

amundsen_application/api/jira/v0.py
csteez: Added resource endpoints - this did require renaming part of a blueprint as the flask restful Api class is named Api automatically and caused collisions with our existing /api/ blueprint endpoint.

amundsen_application/static/js/ducks/issue/reducer.ts
```diff
@@ -0,0 +1,102 @@
...
+  issues: [], 
+  isLoading: false, 
+};
+
+
+/* maybe these should be separate reducers?*/
+export default function reducer(state: IssueReducerState = initialIssuestate, action): IssueReducerState {
+  switch (action.type) {
+    case GetIssues.REQUEST: 
+      return { issues: [], isLoading: false };
```
danwom: `isLoading` should be set to true.

amundsen_application/static/js/utils/tableissues-utils.ts
```diff
@@ -0,0 +1,10 @@
+import { ASSOCIATION_TEXT, MAX_TEXT_LENGTH } from '../components/TableDetail/TableIssues/constants';
+
+export function truncateText(issueTitle: string) : string {
+  if (!issueTitle) {
+    return issueTitle; 
+  }
+  const truncated = issueTitle.length > MAX_TEXT_LENGTH ? 
+    issueTitle.substring(0, MAX_TEXT_LENGTH)  + "...": issueTitle;  
+  return '"' + truncated + '"' +  ASSOCIATION_TEXT;
```
danwom: Instead of truncating the text displayed, we can use the `truncated` css class to add the `...` where it's needed. This will also automatically show/hide more characters if the width changes.

You can check the `TableListItem` for an example. (For some reason we double wrap the `truncated` text, but I'm not 100% sure that's necessary.)

amundsen_application/static/js/components/TableDetail/ReportTableIssue/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+          }
+      </>
+    );
+  }
+}
+export const mapStateToProps = (state: GlobalState, componentProps: ComponentProps) => {
+  return {
+    isLoading: state.issue.isLoading,
+    tableKey: componentProps.tableKey, 
+    tableName: componentProps.tableName
```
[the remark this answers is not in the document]
csteez: neat, i did not know that. tyvm!

amundsen_application/__init__.py
```diff
@@ -52,13 +55,22 @@ def create_app(config_module_class: str, template_folder: str = None) -> Flask:
...
     logging.info('Using search service at {}'.format(app.config.get('SEARCHSERVICE_BASE')))
 
+    api_bp = Blueprint('api', __name__)
+    api = Api(api_bp)
+
+    api.add_resource(IssuesAPI,
+                     '/api/issue/issues', endpoint='issues')
+    api.add_resource(IssueAPI,
+                     '/api/issue/issue', endpoint='issue')
+
```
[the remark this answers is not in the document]
csteez: blueprints are stateless, which means if there's a client used, it has to be created on every request. resources are stateful, so the client can be kept there. another option would have been to create a global, but that doesn't work in this case because the configuration settings are kept in the app.config, which aren't avail when globals are created (before start). 

i'm not a fan either of having both a MVC api and "restful" endpoints in the same project. i think it creates questions for future contributions as to what pattern to follow.

amundsen_application/static/js/components/TableDetail/ReportTableIssue/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+          <a href="javascript:void(0)"
+             className="report-table-issue-link"
+             onClick={this.toggle}
+          >
+            Report Data Issue
+          </a>
+          {
+            this.state.isOpen &&
+            <div className="report-table-issue-modal">
+              <h3 className="">
```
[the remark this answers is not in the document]
csteez: ty, good catch

amundsen_application/static/js/components/TableDetail/ReportTableIssue/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+                Report Data Issue
+              </h3>
+              <button type="button" className="btn btn-close" aria-label={"close"} onClick={this.toggle} />
+              <form id="report-table-issue-form" onSubmit={ this.submitForm }>
+                <input type="hidden" name="key" value={ this.props.tableKey }/>
+
+                <div className="form-group">
+                  <label>Title</label>
+                  <input name="title" className="form-control" required={true} maxLength={200} defaultValue={`Data Issue On: ${this.props.tableName}`}/>
```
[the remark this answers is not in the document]
csteez: when you say display text, do you mean the defaultValue text? or something else?

amundsen_application/static/js/components/TableDetail/TableIssues/index.tsx
csteez: thanks for the tip! i hadn't noticed the target id was empty

amundsen_application/static/css/_icons.scss
```diff
@@ -61,6 +61,11 @@ img.icon {
     mask-image: url('/static/images/icons/Database.svg');
   }
 
+  &.icon-data-quality-warning {
```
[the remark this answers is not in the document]
csteez: good point, i made it more generic!

amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
csteez: i updated to what you suggested but i would love it if you re-reviewed to make sure i did the right thing. i added a variable in _variable-default.scss and then used that - is that correct?

amundsen_application/static/js/utils/tableissues-utils.ts
csteez: good point, i hadn't done it initially because of the additional text required in the mocks. but i'll change it to use ellipsis instead

amundsen_application/api/jira/v0.py
```diff
@@ -0,0 +1,85 @@
...
+            return make_response(jsonify({'msg': missing_config_response}), HTTPStatus.NOT_IMPLEMENTED)
+
+        jira_client = JiraClient()
+        response = jira_client.search(table_key)
+        return make_response(jsonify({'jiraIssues': response}), HTTPStatus.OK)
+
+    except Exception as e:
```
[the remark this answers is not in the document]
csteez: Very good point, I updated the error handling to be specific exceptions (and added an exception class for configuration errors).

amundsen_application/api/jira/v0.py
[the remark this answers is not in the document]
csteez: i added configurations for enabled vs not, and which issue tracker is being used, thank you for the suggestions!

amundsen_application/api/jira/v0.py
csteez: i'm going to get user feedback to see if this was a good way to go or not

requirements.txt
csteez: sure - do all extra requires need a separate docker build? or is that specific to the oidc image because it's an auth library?

amundsen_application/api/jira/v0.py
jinhyukchang: Cool. Btw, apart from what number is proper to show, how about making the method to accept the number of results?

csteez: yeah, i'll make that configurable.

requirements.txt
jinhyukchang: Good point. I think this line can be changed to install this, but I think it's fine to leave it as is and let consumer decide.
https://github.com/lyft/amundsenfrontendlibrary/blob/master/public.Dockerfile#L20

If the questions was about us, I think we can choose it to install in amundsenfrontend-private.

jinhyukchang:
> > Looks good and left several comments.
> > Quick question on search result only showing 3 issues. Are we going to imply that there are more issues such as `...`?
> 
> I think given that in a real use case, it's not likely there will be a giant list of tickets. I think maybe it would be best to remove the max results amount for now and then maybe tune it if the use case requires it?

Yep, I agree. I think `...` would be still useful for the user who wonder why it's only showing 3 results. WDYT?

csteez:
> > > Looks good and left several comments.
> > > Quick question on search result only showing 3 issues. Are we going to imply that there are more issues such as `...`?
> > 
> > 
> > I think given that in a real use case, it's not likely there will be a giant list of tickets. I think maybe it would be best to remove the max results amount for now and then maybe tune it if the use case requires it?
> 
> Yep, I agree. I think `...` would be still useful for the user who wonder why it's only showing 3 results. WDYT?

I agree, I was concerned because I was told this needs to be done by Wednesday but I'm going to circle back with a designer and see if I could get a change to the designs to reflect that. Thanks for the feedback!

jinhyukchang:
> > > > Looks good and left several comments.
> > > > Quick question on search result only showing 3 issues. Are we going to imply that there are more issues such as `...`?
> > > 
> > > 
> > > I think given that in a real use case, it's not likely there will be a giant list of tickets. I think maybe it would be best to remove the max results amount for now and then maybe tune it if the use case requires it?
> > 
> > 
> > Yep, I agree. I think `...` would be still useful for the user who wonder why it's only showing 3 results. WDYT?
> 
> I agree, I was concerned because I was told this needs to be done by Wednesday but I'm going to circle back with a designer and see if I could get a change to the designs to reflect that. Thanks for the feedback!

I see. Didn't know about time pressure. This is not a blocker and I am more than fine with adding TODO as well.

requirements.txt
csteez: i kind of wonder if there's something weird going on because of how this is set up - regardless of what i try it, CI says 

`Processing /home/travis/build/lyft/amundsenfrontendlibrary
  amundsen-frontend 2.0.0 does not provide the extra 'jira'`

csteez: i think i'm going to pass on this change for now since it's blocking CI

amundsen_application/jira/jira_client.py
```diff
@@ -0,0 +1,78 @@
...
+                maxResults=3)
```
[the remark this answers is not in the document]
csteez: made a config setting!

amundsen_application/api/v0.py
```diff
@@ -13,7 +13,7 @@
 
 LOGGER = logging.getLogger(__name__)
 
-blueprint = Blueprint('api', __name__, url_prefix='/api')
+blueprint = Blueprint('main', __name__, url_prefix='/api')
```
csteez: This was needed as flask_restful creates API endpoints as 'api' and this being named api collided with the flask_restful contents. From testing, this doesn't appear to affect anything.

csteez:
> > > > > Looks good and left several comments.
> > > > > Quick question on search result only showing 3 issues. Are we going to imply that there are more issues such as `...`?
> > > > 
> > > > 
> > > > I think given that in a real use case, it's not likely there will be a giant list of tickets. I think maybe it would be best to remove the max results amount for now and then maybe tune it if the use case requires it?
> > > 
> > > 
> > > Yep, I agree. I think `...` would be still useful for the user who wonder why it's only showing 3 results. WDYT?
> > 
> > 
> > I agree, I was concerned because I was told this needs to be done by Wednesday but I'm going to circle back with a designer and see if I could get a change to the designs to reflect that. Thanks for the feedback!
> 
> I see. Didn't know about time pressure. This is not a blocker and I am more than fine with adding TODO as well.

I circled back with design and updated the ui/backend to reflect this change

amundsen_application/api/issue/issue.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        Given a table key, returns all tickets containing that key. Returns an empty array if none exist
+        :return: List of tickets
+        """
+        try:
+            self.client = get_issue_tracker_client()
+            if not app.config['ISSUE_TRACKER_CLIENT_ENABLED']:
+                message = 'Issuing tracking is not enabled. Request was accepted but no issue will be returned.'
+                logging.exception(message)
+                return make_response(jsonify({'msg': message}), HTTPStatus.ACCEPTED)
```
jinhyukchang: Should we return `404 NOT_FOUND`?

amundsen_application/api/issue/issue.py
```diff
@@ -0,0 +1,78 @@
...
+            self.reqparse.add_argument('key', 'Request requires a key', location='args')
+            args = self.reqparse.parse_args()
+            response = self.client.get_issues(args['key'])
+            return make_response(jsonify({'issues': response}), HTTPStatus.OK)
+
+        except IssueConfigurationException as e:
+            message = 'Encountered exception: ' + str(e)
+            logging.exception(message)
+            return make_response(jsonify({'msg': message}), HTTPStatus.NOT_IMPLEMENTED)
+        except (JIRAError, Exception) as e:
```
jinhyukchang: I don't think we need to expose `JIRAError` here, do we?

amundsen_application/api/issue/issue.py
```diff
@@ -0,0 +1,78 @@
...
+            response = self.client.create_issue(description=args['description'],
+                                                table_uri=args['key'],
+                                                title=args['title'])
+            return make_response(jsonify({'issue': response}), HTTPStatus.OK)
+
+        except IssueConfigurationException as e:
+            message = 'Encountered exception: ' + str(e)
+            logging.exception(message)
+            return make_response(jsonify({'msg': message}), HTTPStatus.NOT_IMPLEMENTED)
+        except (JIRAError, Exception) as e:
```
jinhyukchang: Same comment on `JIRAError`

amundsen_application/models/jira_issue.py
```diff
@@ -0,0 +1,15 @@
+from amundsen_application.base.base_issue import BaseIssue
+
+
+class JiraIssue(dict, BaseIssue):
```
jinhyukchang: I wonder if we are making this a bit complicated by having `BaseIssue` as abstraction class. How about just having plain `Issue` class? Other implementation can add optional fields later on..

amundsen_application/models/issue_results.py
```diff
@@ -0,0 +1,21 @@
+from amundsen_application.base.base_issue import BaseIssue
+
+
+class IssueResults(dict):
```
jinhyukchang: Do we need to inherit `dict`? (Is `IssueResults` `dict`?)

amundsen_application/models/jira_issue.py
```diff
@@ -0,0 +1,15 @@
...
+    def __init__(self,
+                 issue_key: str,
+                 title: str,
+                 url: str) -> None:
+        self.issue_key = issue_key
+        self.title = title
+        self.url = url
+
+        dict.__init__(self, issue_key=issue_key,
```
jinhyukchang: Same question as above. Do we need to inherit dict?

amundsen_application/api/issue/issue.py
csteez: whoops yeah good point!

amundsen_application/api/issue/issue.py
csteez: i was following the pattern from another method in the project where if it wasn't configured, it returned accepted. i think 404 would imply there just aren't any issues right? so i'm leaning accepted seems better

amundsen_application/models/jira_issue.py
csteez: yeah i think you're right. i'll remove it.

amundsen_application/models/issue_results.py
csteez: I did this as a workaround to make it serializable without calling a toJSON method. I found adding dict made it work. Very happy to do it differently! Any suggestions?

amundsen_application/api/issue/issue.py
jinhyukchang: Got it. Sounds good.

amundsen_application/models/issue_results.py
jinhyukchang: This will be serialized when it's called with `make_response`, right? 
How about using `vars` which returns dictionary based on attributes?
`vars(issue_results)`

csteez: i get Object of type 'Issue' is not JSON serializable when i try using vars, what I don't quite understand is how the rest of the code works on serialization

jinhyukchang: Is the exception thrown when `vars(issue_results)` is called?

How about:
Option 1 (implement serialize method that returns dictionary):
```
            response = self.client.get_issues(args['key'])
            return make_response(jsonify({'issues': response.serialize()}), HTTPStatus.OK)
```

Or Option 2 (little hacky):
```
            response = self.client.get_issues(args['key'])
            return make_response(jsonify({'issues': response.__dict__}), HTTPStatus.OK)

```

csteez: yeah it's when the vars() method is called. another option is to inherit from JSONEncoder - I might go that route since that seems the most normal? 

https://stackoverflow.com/questions/10252010/serializing-class-instance-to-json/10252138

jinhyukchang: Yep, sounds good to me.

csteez: I ended up doing the .serialize() method instead - the way I did it double serialized the 'data issue' class and I didn't want to mess with it further.

amundsen_application/static/js/components/TableDetail/index.tsx
danwom: nit: can remove the {} above

amundsen_application/static/js/components/TableDetail/TableIssues/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const ASSOCIATION_TEXT = " is associated with this table";
+export const MAX_TEXT_LENGTH = 32; 
+export const ADDITIONAL_ISSUES_TEXT = "additional issues"; 
+export const ADDITIONAL_ISSUES_END = "associated with this table"; 
+export const ADDITIONAL_ISSUES_START = "There are";
```
danwom: Seems like a lot. Can we just do something like `See all Issues` as the link?

I think Jin mentioned something about a `...` functionality to see more issues.

amundsen_application/static/js/components/TableDetail/TableIssues/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  componentDidMount() {
+    this.props.getIssues(this.props.tableKey);
+  }
+
+  renderIssue = (issue: Issue, index: number) => {
+    return (
+      <div className="issue-banner" key={`issue-${index}`}>
+        <a id={`table-issue-link-${index}`} className="table-issue-link" target="_blank" href={issue.url} onClick={logClick}>
+          <img className="icon icon-red-triangle-warning "/>
```
danwom: Do all issues automatically get a `red-triangle-warning` or is there a way to map some sort of severity to different colors? Also do we display issue `status`?

amundsen_application/static/js/ducks/issue/api/v0.ts
```diff
@@ -0,0 +1,19 @@
+import axios, { AxiosResponse } from 'axios';
+
+export const API_PATH = '/api/issue';
+
+export function getIssues(tableKey: string) {
+  return axios.get(`${API_PATH}/issues?key=${tableKey}`)
+  .then((response: AxiosResponse) => {
```
danwom: You can specify the type of the response payload by adding a type `AxiosResponse<Issue>`

amundsen_application/static/js/ducks/issue/reducer.ts
```diff
@@ -0,0 +1,113 @@
...
+/* maybe these should be separate reducers?*/
+export default function reducer(state: IssueReducerState = initialIssuestate, action): IssueReducerState {
+  switch (action.type) {
+    case GetIssues.REQUEST: 
+      return { issues: [], remainingIssuesUrl: null, remainingIssues: 0, isLoading: true }; 
+    case GetIssues.FAILURE: 
+      return { issues: [], remainingIssuesUrl: null, remainingIssues: 0, isLoading: false }; 
+    case GetIssues.SUCCESS: 
+      return {...state, 
+        issues: (<GetIssuesResponse> action).payload.issues,
```
danwom: Instead of casting each field separately, you can do

```
const payload = (<GetIssuesResponse>action).payload;
return {
  ...state,
  ...payload,
}

```

amundsen_application/models/data_issue.py
```diff
@@ -0,0 +1,15 @@
+
+class DataIssue:
+    def __init__(self,
+                 issue_key: str,
+                 title: str,
+                 url: str) -> None:
+        self.issue_key = issue_key
+        self.title = title
+        self.url = url
```
danwom: Can we add additional fields here like status, create date, last updated, etc?

amundsen_application/static/js/components/TableDetail/TableIssues/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+          { issue.issue_key }
+        </a>
+        <span className="issue-title-display-text">
+          <span className="issue-title-name">"
+            { issue.title }
+          </span>"
```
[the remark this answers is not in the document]
csteez: it's actually not - it closes the " from the line above. i'll move the " down a line so it's more visible though. the mocks have the text look like "really long title..." so the quotes have to be outside the div.

amundsen_application/static/js/ducks/issue/reducer.ts
csteez: oh cool! i hadn't thought of that. will do, thanks!

csteez:
> I wonder why we need **init**.py in the upper folder?

Wow must have put it there by accident - really good catch because that fixed the issues I'm running into with mypy

amundsen_application/__init__.py
```diff
@@ -4,16 +4,19 @@
 import logging.config
 import os
 
-from flask import Flask
-
 from amundsen_application.api import init_routes
+from flask import Flask, Blueprint
```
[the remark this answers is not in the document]
csteez: good catch, thanks!

amundsen_application/issue_tracker_clients/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+from threading import Lock
+from werkzeug.utils import import_string
+
+from amundsen_application.base.base_issue_tracker_client import BaseIssueTrackerClient
+
+_issue_tracker_client = None
+_issue_tracker_client_lock = Lock()
+
+
+def get_issue_tracker_client() -> BaseIssueTrackerClient:
```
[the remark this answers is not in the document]
csteez: good point, i'll move this. i'm a bit confused about your comment about the mail client - isn't there not an implementation of the mail client? should i make a module for it?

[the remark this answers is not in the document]
feng-tao: oh, don't worry about the mail client. I refer the mail client which is either our oss FE or in private module which we use for feedback/description request.

amundsen_application/config.py
```diff
@@ -35,6 +35,17 @@ class Config:
     # Initialize custom routes
     INIT_CUSTOM_ROUTES = None  # type: Callable[[Flask], None]
 
+    # Settings for Issue tracker integration
```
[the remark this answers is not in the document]
csteez: yeah very fair - i feel like this feature has an unusually high amount of configs unfortunately

csteez:
> > I wonder why we need **init**.py in the upper folder?
> 
> Wow must have put it there by accident - really good catch because that fixed the issues I'm running into with mypy

Turns out pycharm creates an __init__.py automatically if you move files 🤦‍♀ so that solves where it came from

amundsen_application/__init__.py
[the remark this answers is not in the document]
csteez: it's not that the old apis are inefficient - its that api resources allow for state, and blueprints don't. the existing blueprints have other ways of lazy loading other resources that isn't problematic because of how they access proxy classes (through entry points). i wasn't able to do the same because my client needed configuration on load. 

in general i think restful endpoints should be preferred, but we would also want to rename them to be restful and that's a pretty big refactor. i'll add a todo though.

amundsen_application/issue_tracker_clients/__init__.py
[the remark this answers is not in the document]
feng-tao: (maybe a question @ttanis) also are we plan to move the feedback backend with jira client as well?

amundsen_application/static/js/ducks/issue/api/v0.ts
csteez: Neither of the API responses here are a simple map to the models - would you recommend I create classes for each in that case?

csteez: ie this one returns 
`issues: {issues: Issue[], remaining: number, remaining_url: string}`

amundsen_application/__init__.py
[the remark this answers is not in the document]
ttannis: Its a different way of doing things that worked from our days of Amundsen as proof of concept, you'll see the blueprint pattern in a lot of flask docs and tutorials. It was sufficient when we did not need to maintain state, i.e. before we had any features that leveraged special clients and just had to make get api calls. I'm also in support for leveraging this pattern -- I noticed our other services were using it but it was just an issue of not having time to change something that was working. 

However rather than put a TODO in the code, I'd actually rather an issue be filed in the umbrella repo will the full context of what needs to be changed and why. It will provide better transparency to the community in case someone wants to pick up the task.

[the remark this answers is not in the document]
csteez: I'll remove the TODO and file an issue, thanks for the suggestion!

amundsen_application/static/js/ducks/issue/tests/index.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+        remainingIssues = globalState.issue.remainingIssues; 
+        remainingIssuesUrl = globalState.issue.remainingIssuesUrl;
+      });
+
+      it('gets issues', () => {
+        return expectSaga(getIssuesWorker, action)
+          .provide([
+            [matchers.call.fn(API.getIssues), {issues, remainingIssues, remainingIssuesUrl}],
+          ])
+          .put(getIssuesSuccess(issues, undefined, undefined))
```
danwom: Rather than putting `undefined, undefined` you can add a `?` to the function definition so they become optional parameters.

`getIssuesSuccess(issues: Issue[], remainingIssues?: number, remainingIssuesUrl?: string)`

feng-tao:
let's centralize the doc in lyft/amundsen repo which we could build a sphnix doc website on top of that.

amundsen_application/static/js/ducks/issue/tests/index.spec.ts
csteez: ty!

amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
```diff
@@ -0,0 +1,36 @@
...
+  }
+  .icon-red-triangle-warning  {
+    background: $red-triangle-warning;
+    margin: 0px 5px 0px 0px;
+  }
+  .table-issue-link {
+    margin: 0px 5px 0px 0px; 
+  }
+  .issue-title-name {
+    max-width: 390px;
```
feng-tao: n00b qq, for these css, does it normally get it from our UI designer?

csteez: yes and no - something things i took straight from design (like colors), but sometimes things like default width and size of a row, font in designs is different than what's already in use. 

some things are represented in design but aren't specified (like the margins around the text).

amundsen_application/static/js/ducks/issue/reducer.ts
```diff
@@ -0,0 +1,111 @@
...
+export function getIssuesFailure(): GetIssuesResponse {
+  return { 
+    type: GetIssues.FAILURE, 
+    payload: {
+      issues: [], 
+      remainingIssues: 0,
+      remainingIssuesUrl: null
+    }
+  }
+}
```
ttannis: There are many places where we can set the default values on failures, but we previously decided it should be in the sagas because that allows for the case where we might want to transform the response from the api call before it gets passed back to the application. Suggested change would have to be made together with other items suggested below.
```
export function getIssuesFailure(issues: Issue[], remainingIssues?: number, remainingIssuesUrl?: string): GetIssuesResponse {
  return { 
    type: GetIssues.FAILURE, 
    payload: {
      issues, 
      remainingIssues,
      remainingIssuesUrl
    }
  }
}
```

amundsen_application/static/js/ducks/issue/sagas.ts
```diff
@@ -0,0 +1,39 @@
...
+/** maybe just reload the issues content when there is a new issue created?*/
+
+export function* getIssuesWorker(action: GetIssuesRequest): SagaIterator {
+    const { key } = action.payload; 
+    let response;
+    try {
+        response = yield call(API.getIssues, key); 
+        yield put(getIssuesSuccess(response.issues, response.remaining, response.remaining_url)); 
+    } catch(e) {
+        yield put(getIssuesFailure());
```
ttannis: Related to suggestions above
```
yield put(getIssuesFailure([], 0, null)); 
```

amundsen_application/static/js/ducks/issue/reducer.ts
```diff
@@ -0,0 +1,111 @@
...
+};
+
+
+/* maybe these should be separate reducers?*/
+export default function reducer(state: IssueReducerState = initialIssuestate, action): IssueReducerState {
+  switch (action.type) {
+    case GetIssues.REQUEST: 
+      return { issues: [], remainingIssuesUrl: null, remainingIssues: 0, isLoading: true }; 
+    case GetIssues.FAILURE: 
+      return { issues: [], remainingIssuesUrl: null, remainingIssues: 0, isLoading: false };
```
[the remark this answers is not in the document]
ttannis: Edit: Just use `initialIssueState`.

amundsen_application/static/js/ducks/issue/reducer.ts
ttannis: Edit: Just return `initialIssueState` and this action won't need a payload unless we implement special logic for teh failure case.

amundsen_application/static/js/ducks/issue/sagas.ts
ttannis: Edit: Can leave as is

PR #392, opened 2020-02-27
Improve granularity of logging search actions

ttannis:
### Summary of Changes
This PR aims to introduce better granularity of logging search user actions. 

#### Why is this needed?
At the moment search logging is handled purely by the action logging tied to the search endpoints in Flask. We log the methods the endpoint was called with, which include `search_term` and `page_index`, as well as `filters` if we are using the query string endpoint. 

However because multiple users actions like search bar usage, filtering, & pagination can trigger a search, we do not have any insight into actual user behaviors when it comes to search. For example we cannot easily answer the question "Did the user re-execute a search, or did they just add a filter?" without looking at diffs between logs of the same user and applying assumptions.

#### Option 1
We can pass a new parameter `search_type` in every search request from the React application to be logged. This value will specify what kind of search was executed. i.e. `update_filter`, `update_page`, `update_term`, `automatic` (in the case of inlinesearch). This information would be passed from the redux middleware based on the action that was fired. 

_Pros:_
- Our log information will contain a full picture of what was executed as the keyword args will contain every piece of information passed to the search service, now also with a specific `search_type`
- Knowledge of what to log and when to log it will be abstracted away from the component logic, meaning that when writing new components or modifying existing ones does not require having context on what action to fire + what exactly needs to be logged. Components can continue to be written only with an understanding of what actions should be fired.

_Cons:_
- The only con I can think of if the implementation. Our search sagas could potentially be restructured for an easier implementation. 

#### Option 2
We can log an action in the component more granular commands such as `search:update_filter`, `search:pagination`.

_Pros:_
- This approach has straightforward implementation of calling `logAction` in the component whereever it is necessary.

_Cons:_
- Components only have knowledge of the information they were modifying so these logs will not provide context of the full search input. We have logs that have the full search information but no context of what the user did, and will be creating new that have context of what the user did but do not have the full search input information. 

### Tests

TBD based on approach

### Documentation

TBD based on approach

### CheckList


ttannis:
@danwom please prioritize the final filter feature branch PR review based on your bandwidth this week, but I could also use your thoughts on this to come to consensus before investing more work effort.

You will see here I found myself needing to include -- at minimum -- the information that was modified by the user interaction, such as `search_term`. This because some of our insights relies on having that information in the keyword arguments for the log. I then realized this approach limits the information we can log about the search, because the only place where we have access to the full search input is in the Redux layer and not the components. 

If the problem we have to solve is "log the full context of the search to know exactly what is happening", I am now much more convinced the best approach for this case is to update the update the saga logic such that a `search_type` is passed to the Flask endpoint to be logged. The caveat if that our sagas have to be structured properly and we must ensure that certain sagas are only to be triggered by certain flows.

ttannis:
Latest commit covers the second approach which I currently prefer. It:
1. Fully deprecates the old logic for for searchings tables
2. Logs a `search_type` on all resource endpoints
3. `searchAll` and `searchResource` sagas now require a `searchType` to be passed for logging, and the appropriate "first-layer" sagas have been updated to include an appropriate search type. Current values are `automatic`, `searchbar`, `filter`, `pagination`, and `external_page_view`.

Will wait for comment before pushing updated tests. Also slowly making attempt to establish a clear saga hierarchy in the code.

ttannis:
Use https://github.com/lyft/amundsenfrontendlibrary/pull/396

PR #396, opened 2020-02-29
Improve granularity of logging search actions

ttannis:
### Summary of Changes
This PR aims to introduce better granularity of logging search user actions. 

#### Why is this needed?
At the moment search logging is handled purely by the action logging tied to the search endpoints in Flask. We log the methods the endpoint was called with, which include `search_term` and `page_index`, as well as `filters` if we are using the query string endpoint. 

However because multiple users actions like search bar usage, filtering, & pagination can trigger a search, we do not have any insight into actual user behaviors when it comes to search. For example we cannot easily answer the question "Did the user re-execute a search, or did they just add a filter?" without looking at diffs between logs of the same user and applying assumptions.

#### Approach
We can pass a new parameter `search_type` in every search request from the React application to be logged. This information would be passed from the redux middleware based on the action that was fired. 

_Pros:_
- Our log information will contain a full picture of what was executed as the keyword args will contain every piece of information passed to the search service, now also with a specific `search_type`
- Knowledge of what to log and when to log it will be abstracted away from the component logic, meaning that when writing new components or modifying existing ones does not require having context on what action to fire + what exactly needs to be logged. Components can continue to be written only with an understanding of what actions should be fired.

_Cons:_
- The only con I can think of if the implementation. Our search sagas could potentially be restructured for an easier maintenance of this choice moving forward.

### Tests

Tests update according to modifications.

### Documentation

N/A. We currently do not document implementation details for logging.

### CheckList

### DIFF
--- a/amundsen_application/api/search/v0.py
+++ b/amundsen_application/api/search/v0.py
@@ -11,8 +11,7 @@ from flask.blueprints import Blueprint
 
 from amundsen_application.log.action_log import action_logging
 from amundsen_application.api.utils.request_utils import get_query_param, request_search
-from amundsen_application.api.utils.response_utils import create_error_response
-from amundsen_application.api.utils.search_utils import generate_query_json, map_table_result, valid_search_fields
+from amundsen_application.api.utils.search_utils import generate_query_json, map_table_result
 from amundsen_application.models.user import load_user, dump_user
 
 LOGGER = logging.getLogger(__name__)
@@ -42,11 +41,13 @@ def search_table() -> Response:
 
         filters = request_json.get('filters', {})
 
-        results_dict = _search_table(filters=filters, search_term=search_term, page_index=page_index, search_type=search_type)
+        results_dict = _search_table(filters=filters,
+                                     search_term=search_term,
+                                     page_index=page_index,
+                                     search_type=search_type)
         return make_response(jsonify(results_dict), results_dict.get('status_code', HTTPStatus.INTERNAL_SERVER_ERROR))
     except Exception as e:
         message = 'Encountered exception: ' + str(e)
-        results_dict['msg'] = message
         logging.exception(message)
         return make_response(jsonify(results_dict), HTTPStatus.INTERNAL_SERVER_ERROR)
 
@@ -80,9 +81,9 @@ def _search_table(*, search_term: str, page_index: int, filters: Dict, search_ty
     except Exception as e:
         message = 'Encountered exception generating query json: ' + str(e)
         results_dict['msg'] = message
+        results_dict['status_code'] = HTTPStatus.INTERNAL_SERVER_ERROR
         logging.exception(message)
-        return make_response(jsonify(results_dict), HTTPStatus.INTERNAL_SERVER_ERROR)
-
+        return results_dict
 
     try:
         url = app.config['SEARCHSERVICE_BASE'] + SEARCH_ENDPOINT
@@ -106,6 +107,7 @@ def _search_table(*, search_term: str, page_index: int, filters: Dict, search_ty
     except Exception as e:
         message = 'Encountered exception: ' + str(e)
         results_dict['msg'] = message
+        results_dict['status_code'] = HTTPStatus.INTERNAL_SERVER_ERROR
         logging.exception(message)
         return results_dict
 
@@ -114,7 +116,7 @@ def _search_table(*, search_term: str, page_index: int, filters: Dict, search_ty
 def search_user() -> Response:
     search_term = get_query_param(request.args, 'query', 'Endpoint takes a "query" parameter')
     page_index = get_query_param(request.args, 'page_index', 'Endpoint takes a "page_index" parameter')
-    search_type = request.args.get('searchType')
+    search_type = request.args.get('search_type')
 
     results_dict = _search_user(search_term=search_term, page_index=page_index, search_type=search_type)
 
@@ -173,6 +175,7 @@ def _search_user(*, search_term: str, page_index: int, search_type: str) -> Dict
     except Exception as e:
         message = 'Encountered exception: ' + str(e)
         results_dict['msg'] = message
+        results_dict['status_code'] = HTTPStatus.INTERNAL_SERVER_ERROR
         logging.exception(message)
         return results_dict
 
--- a/amundsen_application/api/search/v0.py
+++ b/amundsen_application/api/search/v0.py
@@ -20,16 +20,15 @@ REQUEST_SESSION_TIMEOUT_SEC = 3
 
 search_blueprint = Blueprint('search', __name__, url_prefix='/api/search/v0')
 
-SEARCH_ENDPOINT = '/search_table'
+SEARCH_TABLE_ENDPOINT = '/search_table'
 SEARCH_USER_ENDPOINT = '/search_user'
 
 
 @search_blueprint.route('/table', methods=['POST'])
 def search_table() -> Response:
     """
-    TODO (ttannis): Update this docstring after amundsensearch documentation is merged
-    Calls the search service to execute a search. The request data is transformed
-    to the json payload defined [link]
+    Parse the request arguments and call the helper method to execute a table search
+    :return: a Response created with the results from the helper method
     """
     try:
         request_json = request.get_json()
@@ -55,13 +54,11 @@ def search_table() -> Response:
 @action_logging
 def _search_table(*, search_term: str, page_index: int, filters: Dict, search_type: str) -> Dict[str, Any]:
     """
-    call the search service endpoint and return matching results
+    Call the search service endpoint and return matching results
+    Search service logic defined here:
+    https://github.com/lyft/amundsensearchlibrary/blob/master/search_service/api/table.py
+
     :return: a json output containing search results array as 'results'
-
-    Schema Defined Here:
-    https://github.com/lyft/amundsensearchlibrary/blob/master/search_service/api/search.py
-
-    TODO: Define an interface for envoy_client (ttannis fix for filter)
     """
     # Default results
     tables = {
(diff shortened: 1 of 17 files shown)


amundsen_application/static/js/ducks/search/sagas.ts
```diff
@@ -68,7 +68,7 @@ export function* filterWorker(): SagaIterator {
   const state = yield select(getSearchState);
   const { search_term, selectedTab, filters } = state;
   const pageIndex = getPageIndex(state)
-  yield put(searchResource(search_term, selectedTab, pageIndex));
+  yield put(searchResource("filter", search_term, selectedTab, pageIndex));
```
ttannis: TODO: Will make these enums in the next upload, and rethink better names. i.e. "external_page_view" is more accurately when the search occurs from pulling parameters from the url -- which could also happen on refresh or manually updating url -- so will consider a different name.

amundsen_application/static/js/ducks/search/reducer.ts
```diff
@@ -49,13 +49,14 @@ export interface SearchReducerState {
 };
 
 /* ACTIONS */
-export function searchAll(term: string, resource?: ResourceType, pageIndex?: number, useFilters: boolean = false): SearchAllRequest {
+export function searchAll(searchType: string, term: string, resource?: ResourceType, pageIndex?: number, useFilters: boolean = false): SearchAllRequest {
```
danwom: nit: `searchType` is the first reducer param but the last API param. Actually all of the fields are in a different order.

Would like to see some consistency, or convert ordered params into a map.

amundsen_application/static/js/ducks/search/sagas.ts
danwom: Can we use the redux action here? Most workers have only one redux action, but in this case it would have two: `UpdateSearchFilter.CLEAR_CATEGORY` and `UpdateSearchFilter.UPDATE_CATEGORY`. Not sure how to access the value in the worker.

amundsen_application/static/js/ducks/search/sagas.ts
```diff
@@ -90,16 +90,16 @@ export function* filterWatcher2(): SagaIterator {
...
 };
 
 export function* inlineSearchWorker(action: InlineSearchRequest): SagaIterator {
   const { term } = action.payload;
   try {
     const [tableResponse, userResponse] = yield all([
-      call(API.searchResource, 0, ResourceType.table, term),
-      call(API.searchResource, 0, ResourceType.user, term),
+      call(API.searchResource, 0, ResourceType.table, term, {}, "automatic"),
+      call(API.searchResource, 0, ResourceType.user, term, {}, "automatic"),
```
danwom: Not a fan of using `automatic`. Would prefer `InlineSearch.REQUEST` here.

amundsen_application/static/js/ducks/search/sagas.ts
```diff
@@ -264,7 +207,7 @@ export function* urlDidUpdateWorker(action: UrlDidUpdateRequest): SagaIterator {
 
   const state = yield select(getSearchState);
   if (!!term && state.search_term !== term) {
-    yield put(searchAll(term, resource, parsedIndex));
+    yield put(searchAll("external_page_view", term, resource, parsedIndex));
```
danwom: Not necessarily an `external_page_view`, but a browser navigation. Browser forward and backward, external URL, breadcrumbs (maybe?).

amundsen_application/api/search/v0.py
```diff
@@ -21,151 +20,56 @@
 
 search_blueprint = Blueprint('search', __name__, url_prefix='/api/search/v0')
 
-SEARCH_ENDPOINT = '/search'
+SEARCH_ENDPOINT = '/search_table'
```
danwom: It appears we don't have a generic search endpoint, but a different endpoint for each resource type. Better named as `SEARCH_TABLE_ENDPOINT`.

tests/unit/api/search/test_v0.py
```diff
@@ -49,15 +49,15 @@ class SearchTableQueryString(unittest.TestCase):
...
+        self.search_url = local_app.config['SEARCHSERVICE_BASE'] + SEARCH_ENDPOINT
 
     def test_fail_if_term_is_none(self) -> None:
         """
         Test request failure if 'term' is not provided in the request json
         :return:
         """
         with local_app.test_client() as test:
-            response = test.post('/api/search/v0/table_qs', json={'pageIndex': 0})
+            response = test.post('/api/search/v0/table', json={'pageIndex': 0})
```
danwom: Should `/table` be another constant?

amundsen_application/api/search/v0.py
ttannis: Will do.

amundsen_application/static/js/ducks/search/reducer.ts
ttannis: I'd like to do this with future work consolidating sagas, and yes this just follows existing tech debt where there are all out of order. I prefer maps moving forward, and maybe we can start that with the dashboard work and I can try to clean this up on existing code on `master` in parallel.

amundsen_application/static/js/ducks/search/sagas.ts
ttannis: I'm going to make this one resolve to "inline". For context, the types of searched I considered "automatic" were inline searches and the new search that happens under the hood when the user clears the search term. I am just going to call everything exactly what it is. Considering the following:
```
export enum SearchType {
  CLEAR_TERM = 'clearTerm',
  FILTER = 'filter',
  INLINE_SEARCH = 'inline',
  LOAD_FROM_URL = 'loadFromUrl',
  PAGINATION = 'pagination',
  SEARCH_BAR = 'searchbar',
}
```

amundsen_application/static/js/ducks/search/sagas.ts
ttannis: Considering `loadFromUrl` for this type

amundsen_application/static/js/ducks/search/sagas.ts
ttannis: Not sure I understand. `searchResource()` is an action.

tests/unit/api/search/test_v0.py
ttannis: It could be part of the test setup but I don't see a need for it to be a defined constant. The only functional code that needs to know this endpoint is in the TypeScript code.

ttannis:
@danwom besides code correctness, let me know what you think about the search type values. In the future even if we clean up how we send those values over, changing the actual strings will add extra work when updating our queries.

tests/unit/api/search/test_v0.py
```diff
@@ -49,15 +49,16 @@ class SearchTableQueryString(unittest.TestCase):
     def setUp(self) -> None:
         self.mock_table_results = MOCK_TABLE_RESULTS
         self.expected_parsed_table_results = MOCK_PARSED_TABLE_RESULTS
-        self.search_url = local_app.config['SEARCHSERVICE_BASE'] + SEARCH_ENDPOINT
+        self.search_service_url = local_app.config['SEARCHSERVICE_BASE'] + SEARCH_TABLE_ENDPOINT
+        self.fe_flask_endpoint = '/api/search/v0/table'
```
danwom: The tests are split between user search and table search. I think we clobbered some of the user searches.

amundsen_application/static/js/interfaces/Enums.ts
```diff
@@ -7,3 +7,13 @@ export enum FilterType {
...
+
+export enum SearchType {
+  CLEAR_TERM = 'clearTerm',
+  FILTER = 'filter',
+  INLINE_SEARCH = 'inlineSearch',
+  INLINE_SELECT = 'inlineSelect',
+  LOAD_FROM_URL = 'loadFromUrl',
+  PAGINATION = 'pagination',
+  SEARCH_BAR = 'searchbar',
+}
```
danwom: I guess we could try to make these slightly more descriptive. 

`clearTerm -> clearSearchBar` (I wasn't initially sure what this referred to)
`filter -> updateFilter`
`searchbar -> submitSearchBar`

tests/unit/api/search/test_v0.py
ttannis: Is this a misunderstanding of the diff? The user searches are in a separate test class with its own `setUp()` that uses `/api/search/v0/user`.

amundsen_application/static/js/interfaces/Enums.ts
ttannis: These are more descriptive but I'm also looking for consistency across the names. I'm going to try a pattern that loosely follows `::verbNoun::` (exception is "inlineSearch/Select").
```
export enum SearchType {
  CLEAR_TERM = 'clearSearchTerm',
  FILTER = 'updateFilter',
  INLINE_SEARCH = 'inlineSearch',
  INLINE_SELECT = 'inlineSelect',
  LOAD_URL = 'loadUrl',
  PAGINATION = 'updatePage',
  SUBMIT_TERM = 'submitSearchTerm',
}
```
Intentionally using `searchTerm` vs. `searchBar` because I consider the "thing" as the information we care about as opposed to the component that updates it.

danwom: I like to think about these in terms of user action and user intent. 

Also I don't think it's necessary to apply a pattern to these.

tests/unit/api/search/test_v0.py
danwom: I missed the user test setup on line 222.

amundsen_application/static/js/interfaces/Enums.ts
ttannis: Unrelated to the content of this PR I'd say we have some more overall guiding principles for frontend logging to flesh out and when the time comes Product should be involved. Something I find tricky to think through is that the purpose of these logs has always been to collect whatever metrics help us make product decisions, which we ourselves do not always know. Another thing I find tricky is that the end result of the actions are also important, but user actions and user intent will not always mirror the end result, because the application logic could be doing any amount of things under the hood and sometimes whats happening under the hood could matter. I don't have the answer or believe there is a clear one, this could very well be a case-by-case kind of issue.

danwom: I think both types of logging are useful, but for different purposes. For example when a user runs a search, we log API calls for `search_user`, `search_table`, and soon `search_dashboard`. This kind of logging can help track server load, cost, efficiency, etc. This can help us make better engineering decisions. For product related decisions, I think user intent and user actions are the key metrics we should be looking at.

PR #398, opened 2020-03-04
Update search results to include badges and display them

csteez:
### Summary of Changes

Some tables have badges which are displayed on the table detail view page. These badges contain useful information that can help users determine which table they want to look at. This change surfaces that information on search. 

### Tests

Modified tests to confirm that the results are displayed

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/api/utils/search_utils.py
+++ b/amundsen_application/api/utils/search_utils.py
@@ -1,8 +1,4 @@
-import logging
-
 from typing import Dict, List  # noqa: F401
-
-LOGGER = logging.getLogger(__name__)
 
 
 # These can move to a configuration when we have custom use cases outside of these default values
--- a/amundsen_application/api/utils/search_utils.py
+++ b/amundsen_application/api/utils/search_utils.py
@@ -20,7 +20,7 @@ def map_table_result(result: Dict) -> Dict:
         'description': result.get('description', None),
         'database': result.get('database', None),
         'schema': result.get('schema', None),
-        'badges': _map_badges(result.get('badges', None)),
+        'badges': result.get('badges', None),
         'last_updated_timestamp': result.get('last_updated_timestamp', None),
     }
 
@@ -54,8 +54,3 @@ def generate_query_json(*, filters: Dict = {}, page_index: int, search_term: str
         'query_term': search_term
     }
 
-
-def _map_badges(badges: List[str]) -> List[Dict[str, str]]:
-    if not badges:
-        return []
-    return [{"tag_name": badge, "tag_type": "badge"} for badge in badges]
--- a/amundsen_application/api/utils/search_utils.py
+++ b/amundsen_application/api/utils/search_utils.py
@@ -53,4 +53,3 @@ def generate_query_json(*, filters: Dict = {}, page_index: int, search_term: str
         },
         'query_term': search_term
     }
-
--- a/amundsen_application/api/utils/search_utils.py
+++ b/amundsen_application/api/utils/search_utils.py
@@ -1,10 +1,14 @@
+import logging
+
 from typing import Dict, List  # noqa: F401
+
+LOGGER = logging.getLogger(__name__)
+
 
 # These can move to a configuration when we have custom use cases outside of these default values
 valid_search_fields = {
     'column',
     'database',
-    'badge',
     'schema',
     'table',
     'tag'
--- a/amundsen_application/api/utils/search_utils.py
+++ b/amundsen_application/api/utils/search_utils.py
@@ -4,6 +4,7 @@ from typing import Dict, List  # noqa: F401
 valid_search_fields = {
     'column',
     'database',
+    'badge',
     'schema',
     'table',
     'tag'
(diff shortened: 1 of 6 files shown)


amundsen_application/api/utils/search_utils.py
```diff
@@ -51,3 +53,9 @@ def generate_query_json(*, filters: Dict = {}, page_index: int, search_term: str
         },
         'query_term': search_term
     }
+
+
+def _map_badges(badges: List[str]) -> List[Dict[str, str]]:
+    if not badges:
+        return []
+    return [{"tag_name": badge, "tag_type": "badge"} for badge in badges]
```
csteez: I did this because the metadata service returns tags as tags: [{tag_name: name, tag_type: type}] and it doesn't make sense to have the front end handle the mapping. 
I wasn't sure if it makes more sense to make a model here or not given how other results are handled.

ttannis: Did you mean the metadata service or the search service? If metadata service wondering how that applies to this feature. If search service, what are the repercussions using the information directly? 

Overall I would expect that the search service can return exactly what the frontend service needs. This is not currently always the case with some features as you may have noticed, but I think it's a problem that should more actively addressed in new features vs accumulating more workarounds.

csteez: I meant the metadata service - it's the only other entity returning badge information that is consumed by the front end. The reference to it was with regards to keeping badge data output consistent so that the FE doesn't have to worry about it. 

The search service already doesn't return exactly what the front end needs, but more a direct mapping of the search providers output (ie from ES/Atlas) to a table or user entity as defined by the search services models. The mapping of a badge model to what the front end expects seems more like an implementation detail for the front end service than for the search service from my perspective. 

But I'm happy to change things if it makes more sense to make the mapping further upstream.

ttannis: I'd prefer the mapping to be further upstream, and for amundsensearch to represent a badge using the full object that represents it. So far we've been lucky that all the values we've needed to show in the search results were strings -- like table names and descriptions -- so badges introduces a new case that should be handled differently if possible.

There is a need for the frontend to have the full representation of the badge object in the search results. While it's easy to throw in the `tag_type` hardcoded as "badge" right now, I'm not sure what we would do if and when we extend the badge object. For example, introducing a `category` value into badges is the only scalable way for the frontend to know what color different badges should be so that we can do a mapping from category to color.

feng-tao: IMO the concept of badge is not e2e 100% well designed/decided. In UI, it is not editable , but we never consider how we ingest the badge and who could ingest the badge. And the data model is 100% exactly the same as tag which it exactly fits into the tag model to avoid duplicate code (otherwise we will have so many duplicate code in metadata service/search).

feng-tao: but the change here is a bit hacky...I think we need to think through on how to allow FE to understand the badge concept without too much code by handling tag vs badge in the search service...

csteez: I was working on making the search service and databuilder handle tag vs badge per tamika's comments - though I feel like doing it at the databuilder level seems not great.

amundsen_application/api/utils/search_utils.py
```diff
@@ -1,5 +1,10 @@
+import logging
+
 from typing import Dict, List  # noqa: F401
 
+LOGGER = logging.getLogger(__name__)
```
ttannis: Do you intend to leave this here, or was this just for testing? Since these are utility methods that can be called by any APIs that need them, I'd expect any issues to be caught and logged from the API method as opposed to ever needing to log anything here.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/tests/index.spec.tsx
```diff
@@ -126,26 +127,36 @@ describe('TableListItem', () => {
         expect(resourceBadges.exists()).toBe(true);
       });
 
-      describe('if props.table has last_updated_timestamp', () => {
-        it('renders Last Updated title', () => {
-          expect(resourceBadges.children().at(0).children().at(0).text()).toEqual('Last Updated');
+      describe('if props.table has badges', () => {
+      
+        it('renders flag for badge', () => {
```
ttannis: nit: `'renders flag for badge'` --> `'renders BadgeList for badges'`

not a nit: For a better test we want to check not just that a component exists but verify that `TableListItem` gave it the right properties. So we could update this text to be:
```
expect(resourceBadges.find(BadgeList).props().badges).toEqual(props.table.badges);
```
That one assertion will end up failing if `BadgeList` doesn't exist but also if it doesn't get the right properties passed, making it a more useful test.

amundsen_application/api/utils/search_utils.py
csteez: good catch, that was from local testing. i'll remove it. thanks!

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/tests/index.spec.tsx
csteez: Super useful, thanks so much! I was trying to figure out how to test that the correct values were present. I hadn't thought of trying the props.

PR #401, opened 2020-03-06
Add reporter for JIRA Issue Creation

ttannis:
### Summary of Changes

Passing `reporter={'name': valid_jira_id}` sets the reporter for the issue. This current workaround can and should probably be improved in the future. It will not work for users who have aliases or any case where the `jira_id` is not the employee's email handle. 

### Tests

Updated relevant test.

### Documentation

N/A for now. If we keep this implementation we should call out the limitation above in the feature documentation.

### CheckList

### DIFF
--- a/amundsen_application/proxy/issue_tracker_clients/jira_client.py
+++ b/amundsen_application/proxy/issue_tracker_clients/jira_client.py
@@ -1,5 +1,8 @@
 from jira import JIRA, JIRAError, Issue
 from typing import List
+
+from flask import current_app as app
+
 from amundsen_application.base.base_issue_tracker_client import BaseIssueTrackerClient
 from amundsen_application.proxy.issue_tracker_clients.issue_exceptions import IssueConfigurationException
 from amundsen_application.models.data_issue import DataIssue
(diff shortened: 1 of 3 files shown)


amundsen_application/proxy/issue_tracker_clients/jira_client.py
```diff
@@ -67,12 +70,15 @@ def create_issue(self, table_uri: str, title: str, description: str) -> DataIssu
         :return: Metadata about the newly created issue
         """
         try:
+            user_email = app.config['AUTH_USER_METHOD'](app).email
```
csteez: should you check for null/empty before running split?

amundsen_application/proxy/issue_tracker_clients/jira_client.py
```diff
@@ -67,12 +70,15 @@ def create_issue(self, table_uri: str, title: str, description: str) -> DataIssu
...
+            jira_id = user_email.split('@')[0]
```
feng-tao: as discussed, create an issue in the jira sdk repo and link it here as a todo which we could remove this hack once the change is in (hopefully some day in the future).

ttannis: @csteez I don't have the full context of the sdk issue, are you able to file one and/or send a link to any of the conversations you found.

csteez: https://community.atlassian.com/t5/Answers-Developer-Questions/JIRA-Rest-API-find-JIRA-user-based-on-user-s-email-address/qaq-p/532715

amundsen_application/proxy/issue_tracker_clients/jira_client.py
ttannis: For this one I'm going to check for `app.config['AUTH_USER_METHOD']`. This method is typed as returning a `User` and that type guarantees an `email` attribute.

amundsen_application/proxy/issue_tracker_clients/jira_client.py
ttannis: Added.

ttannis:
> i think i would also alter the description part of it to include the reporter email since it defaults to "anonymous" otherwise.

Added.

PR #403, opened 2020-03-10
Add chart and query component to dashboard page

danwom:
### Summary of Changes

Added basic versions of the chart and query component. Since this actively designed, these are mostly placeholders.

### Tests

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/static/css/_list-group.scss
+++ b/amundsen_application/static/css/_list-group.scss
@@ -7,7 +7,7 @@
     border-right: none;
     padding: 0;
 
-    &:hover {
+    &.clickable:hover {
       box-shadow: 0 0 1px 0 rgba(0, 0, 0, 0.12), 0 2px 3px 0 rgba(0, 0, 0, 0.16);
       cursor: pointer;
       z-index: 1;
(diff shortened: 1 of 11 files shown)


amundsen_application/static/css/_list-group.scss
```diff
@@ -7,7 +7,7 @@
     border-right: none;
     padding: 0;
 
-    &:hover {
+    &.clickable:hover {
```
ttannis: Just a thought, but since it  like most of our list groups will actually be clickable, what do you think of making it such that items have to opt-out of this style as opposed to opt-in? Fine with either way for now.

danwom: I thought about that too-- even though most list-group-items in Amundsen happen to be clickable at the moment, list-group-items in general tend not to be clickable by default. 

I think it's more intuitive to add behavior with special classes rather than take them away with negative classes. For example bootstrap has the optional `list-group-item-action` class that you can add to normal `list-group-item`... Now that I think about it, we _could_ use that class but then we'd have to overwrite all of the default styles as well.

PR #404, opened 2020-03-12
Cast input to lowercase like we do for SearchBar

ttannis:
### Summary of Changes

Because search is currently case sensitive, the filter will not return results if someone enters uppercase. To address this `InputFilter` will do the same thing `SearchBar` has been doing, and cast the value to lower case automatically on input.

As a note, the ideal experience for this would be addressed by: https://github.com/lyft/amundsen/issues/283

### Tests

Updated applicable test

### Documentation

N/A

### CheckList


ttannis:
fyi @feng-tao another case for making search case insensitive. Unless you can think of a reason not to do this, I could follow up with the community members who say they made that change internally and see if they want to push those changes upstream for us to take a look.

PR #406, opened 2020-03-16
Fix overflow for issue tracking feature

csteez:
### Summary of Changes

Very long JIRA titles are overflowing on to the next line. This shrinks the width at which overflow is used for the list of issues so that doesn't happen. 

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
+++ b/amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
@@ -24,9 +24,6 @@
     min-width: fit-content;
   }
   .issue-title-name {
-    flex-basis: 50px;
-    flex-grow: 0.6;
-    flex-shrink: 4;
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
--- a/amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
+++ b/amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
@@ -8,6 +8,8 @@
     height: 40px;
     margin: 8px 0;
     padding: 8px 16px;
+    display: flex; 
+    overflow: hidden;
 
     &:first-child {
       margin-top: 24px;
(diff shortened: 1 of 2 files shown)


amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
```diff
@@ -21,7 +21,7 @@
     margin: 0px 5px 0px 0px; 
   }
   .issue-title-name {
-    max-width: 390px; 
+    max-width: 353px;
```
ttannis: @dwon is there a way of leveraging something that was defined in out layouts to have a better idea of what this could be? If and when layouts get updated I'm wondering if we can avoid having to manuall re-evaluate and update the max-widths of elements in the left panel.

csteez: i was wondering something similar - happy to change to something that would be more maintainable.

ttannis: @cstead If there's no other options that Daniel might be savvy to given his familiarity with the layout work, I'm fine with this fix but wanted to check first.

danwom: Thinking through some alternatives...

1) `max-width: calc(100% - { some constant } px` This may not work so well with variable width ticket numbers. You just allocate a set width the ticket numbers, but this could break with some long ticket names. (IE, `DATAHELP-1123` vs `IT-SUPPORTDESK-1232` at another company)
2) Use flex basis, shrink, and grow to automatically let the summary grow into the remaining space. You may need to play around with the structure to get the `...` text-overflow to work properly. There is a working example in the TableListItem where we truncate table names and descriptions.
3) Use the same width breakpoints with new constant values -- this is delicate and likely to break.

csteez: @danwom Thanks so much for your help and advice! I couldn't get it to work with truncated on the direct div, but on a parent it worked great.

ttannis: @csteez  I tried this out and experienced that the issue key would wrap on small screens too. If so, what seemed to work was wrapping the issue key in a `<span>` and adding `min-width: fit-content;` to `.table-issue-link`

csteez: @ttannis how of small a screen did you try it on? i'm not seeing it replicated on my local.

ttannis: My home laptop and shrinking the browser as skinny and Amundsen allows. I had other changes active while testing these styles though, so if you're not seeing that occur no problem, I'll just check if I can reproduce it in staging when this goes out.

csteez: sounds like what i did too. i did the span change anyway, and i already had `min-width: fit-content;` set

PR #407, opened 2020-03-17
Clean up some tests for programmatic descriptions feature

ttannis:
### Summary of Changes

Some cleanup on top of https://github.com/lyft/amundsenfrontendlibrary/pull/381
1. Python unit tests were updated to explicitly test the behavior of helper functions added in `metadata_utils`
2. Placed hardcoded text from `TableDetail` into a `constants` file. We are separating out hardcoded user-readable text for out components because it will make them easier to update in the future -- especially if we eventually allow for constants for be overwritten.
3. Updated the test for the `EditableSection` title to verify that convertText() is called and the result of that method is what gets rendered. 

### Tests

Included above.

### Documentation

N/A. Features have not been modified.

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/TableDetail/index.tsx
+++ b/amundsen_application/static/js/components/TableDetail/index.tsx
@@ -15,6 +15,7 @@ import BookmarkIcon from 'components/common/Bookmark/BookmarkIcon';
 import Breadcrumb from 'components/common/Breadcrumb';
 import DataPreviewButton from 'components/TableDetail/DataPreviewButton';
 import ColumnList from 'components/TableDetail/ColumnList';
+import EditableText from 'components/common/EditableText';
 import ExploreButton from 'components/TableDetail/ExploreButton';
 import Flag from 'components/common/Flag';
 import FrequentUsers from 'components/TableDetail/FrequentUsers';
@@ -40,7 +41,8 @@ import { formatDateTimeShort } from 'utils/dateUtils';
 import './styles';
 import RequestDescriptionText from './RequestDescriptionText';
 import RequestMetadataForm from './RequestMetadataForm';
-import EditableText from "components/common/EditableText";
+
+import { PROGRMMATIC_DESC_HEADER } from './constants';
 
 export interface StateFromProps {
   isLoading: boolean;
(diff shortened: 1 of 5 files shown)


ttannis:
@samshuster as promised, check out this PR for some minor cleanup I added on top of your PR in order to make tests a little more robust.

samshuster:
thanks @ttannis this was useful!

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -211,9 +213,10 @@ class TableDetail extends React.Component<TableDetailProps & RouteComponentProps
                   </EditableSection>
                 </section>
               </section>
-              {data.programmatic_descriptions.length > 0 &&
+              {
+                data.programmatic_descriptions.length > 0 &&
```
verdan, line 217: @ttannis @feng-tao same issue with the truthy conditions here as well. if this value is null then this is breaking the UI. 😢

ttannis: @verdan is this for the case of the quickstart as well? In practice the metadata service is supposed to be returning an empty array/list for this. 

We can fix all these cases but I do want to understand if this is either a confirmed quickstart issue, if it is affecting your use of Amundsen at ING, or if you're just calling this out to be addressed in the future. Will help prioritize when I go through and look for all these cases that may have been introduced in 2.1.0

verdan: @ttannis 
I am using the latest version of both frontend and metadata and it's failing. 

Two things here: 
1. This is only implemented on the Neo4j proxy in metadata, hence not available in atlas proxy for now, and returning default value. 
2. `Table` model in amundsen common says the default value of this variable is None. i.e., `programmatic_descriptions: Optional[List[ProgrammaticDescription]] = None`

However, in my opinion it should be 
`programmatic_descriptions: Optional[List[ProgrammaticDescription]] = []` 

or even this will work: 
`programmatic_descriptions: List[ProgrammaticDescription] = []`

samshuster: let me know if I can be of any help. It seems like the amundsen common change would be an easy solution?

ttannis: @samshuster I think that would be correct. I was under the impression the metadata service would be returning an empty list, and should have sanity checked the code in the metada library to ensure this was the case.

@verdan thanks for flagging this. Either of you can feel free to fix if you are able to get to it before I can, and we will be sure to get the PR through and create a new release to unblock as soon as possible.

samshuster:
thank you for catching and fixing @verdan and apologies about that.

PR #411, opened 2020-03-20
Initial dashboard search

ttannis:
### Summary of Changes
1. Hooks up the frontend dashboard endpoint to call the search service
2. Updates relevant Redux logic to include processing dashboard requests for search and inline search
3. Exposes dashboards in inline search and on the search page.

image

image

### Tests

TBD

### Documentation

Added docstrings and comments where applicable.

### CheckList


ttannis:
Beyond this existing work, there is another PR that will:
1. Update the content of the dashboard items shown in inline search and searchpage, we need to reevaluate what content we actually need to display and what the schema for dashboard search results will be. 
2. Addressing every "TODO ttannis" I added...makes it easier to know exactly where to come back to everything when addressing item 1.
3. Improving/completing tests

amundsen_application/static/js/components/common/ResourceListItem/DashboardListItem/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { Link } from 'react-router-dom';
+
+import { LoggingParams } from '../types';
+
+import BookmarkIcon from 'components/common/Bookmark/BookmarkIcon';
+
+import { toTitleCase } from 'utils/stringUtils';
+
+export interface DashboardListItemProps {
+  dashboard: any; // TODO ttannis: Add DashboardResource type when we know what it is
```
danwom, line 12: This is a type defined in `interfaces/Dashboard.ts` that's for a full-dashboard metadata. We could define a type `DashboardSummary` as a placeholder-- similarly I would like to create a `TableSummary` type to use in search results, `user/table/follows`, `user/table/reads`, `user/table/owns`, etc. This would also replace the `PopularTables` type as well.

amundsen_application/static/js/components/common/ResourceListItem/DashboardListItem/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+                    `${dashboard.dashboard_group} - ${dashboard.dashboard_name}`
+                  }
+                </div>
+                <BookmarkIcon bookmarkKey={ 'TODO ttannis: Add bookmarkKey' }/>
+              </div>
+              <div className="body-secondary-3 truncated">{ dashboard.dashboard_group_description }</div>
+            </div>
+          </div>
+          <div className="resource-type">
+            { toTitleCase(dashboard.product) }
```
danwom, line 48: Seems like this would work with our current analytics platforms but may not work for others. For example there could be one that is all-caps, all-lower, or stylized like `iPhone`. 

As much as I'd hate to make another config to map names that may be the most accurate method. Alternatively we can just put it in all caps.

amundsen_application/api/search/v0.py
```diff
@@ -184,6 +185,69 @@ def _map_user_result(result: Dict) -> Dict:
...
+
+    results_dict = {
+        'search_term': search_term,
+        'msg': '',
+        'dashboards': dashboards,
+    }
+
+    try:
+        url_base = app.config['SEARCHSERVICE_BASE'] + SEARCH_DASHBOARD_ENDPOINT
+        # TODO ttannis: Request consistency to use f'{url_base}?query_term={search_term}&page_index={page_index}'
```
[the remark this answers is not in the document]
ttannis: Yes the current endpoint from search is formatted differently, so even though things work I was going to ask if there was some ideal format or pattern that we intend to converge on, so that things aren't different every time we add a new endpoint. 

@feng-tao was it intentional to make the dashboard endpoint have a new pattern? If it was intentional is it because we eventually want to make all the endpoints match the new pattern?

amundsen_application/api/utils/search_utils.py
```diff
@@ -11,6 +11,17 @@
...
 
+def map_dashboard_result(result: Dict) -> Dict:
+    # TODO ttannis: Need the dashboard_uri/key. Update when we know what this is supposed to be
+    return {
+        'type': 'dashboard',
+        'dashboard_group': result.get('dashboard_group', None),
+        'dashboard_name': result.get('dashboard_name', None),
+        'dashboard_group_description': result.get('dashboard_group_description', None),
+        'product': result.get('product', None),
+    }
```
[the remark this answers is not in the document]
ttannis: Now that we know the field names, I will update this when the schema is finalized from search.

In regards to including `type` are you suggesting including it at the service level from search service. If so, I don't have a solid reason to prefer one option over the other. I'll let Tao make that decision and update this code based on that.

amundsen_application/static/js/components/common/ResourceListItem/DashboardListItem/index.tsx
ttannis: This makes sense. I'll update the code in this manner when the search endpoint is ready.

amundsen_application/static/js/components/common/ResourceListItem/DashboardListItem/index.tsx
ttannis: That's true. I'll leave this as is for now and revisit if that use case comes up.

amundsen_application/api/search/v0.py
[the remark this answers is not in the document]
feng-tao: will take a look tomorrow. if I haven't, please remind me in the standup as it is a bit late now:(

PR #412, opened 2020-03-20
Show status, priority for data issues

csteez:
### Summary of Changes

Updates the ui for issue tracking to include status and priority. Moves the list of issues below description. Updates to show resolved issues as well. 

### Tests
Tests were modified to test new functionality. 

### Documentation

N/A

### CheckList


amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
```diff
@@ -27,12 +23,49 @@
...
+    margin: 0px 0px 0px 5px;
+  }
+  .table-issue-priority {
+    margin: 0px 5px 0px 0px;
+    border: 1px solid $rose80; 
+    border-radius: 5px;
+    padding: 1px 2px;
+    flex: none; 
+  }
+  .blocker {
```
csteez: Are these kind of styles something that should be configurable? I wasn't sure how to properly handle configuration here since the "priorities" are set by the issue tracker rather than by the application.

amundsen_application/static/css/_variables-default.scss
```diff
@@ -103,6 +103,8 @@ $list-group-border-radius: 0 !default;
 // Labels
 $label-primary-bg: $brand-color-3 !default;
 
+//Priority
+$priority-text-blocker: $white;
```
ttannis: nit: Can you also define some `$priority-bg-color`, set it to `$rose80`, and reference that in the styles. Always nicer to have things defined in the same place.

amundsen_application/static/js/components/TableDetail/TableIssues/tests/index.spec.tsx
```diff
@@ -35,50 +38,56 @@ describe ('TableIssues', ()=> {
...
       const { props, wrapper } = setup({ issues: [{
         issue_key: 'issue_key', 
         title: 'title',
-        url: 'http://url'
+        url: 'http://url', 
+        status: 'Open', 
+        priority_display_name: 'P2', 
+        priority_name: 'major'
       }]}); 
       expect(wrapper.find('.table-issue-link').text()).toEqual('issue_key');
```
ttannis: nit: we can check for the new display name and status here as well.

amundsen_application/static/js/components/TableDetail/TableIssues/tests/index.spec.tsx
```diff
@@ -35,50 +38,56 @@ describe ('TableIssues', ()=> {
...
-        remainingIssues: 1, 
-        remainingIssuesUrl: 'url'
+        total: 1, 
+        allIssuesUrl: 'url'
       });
-      expect(wrapper.find('.table-issue-more-issues').text()).toEqual(SEE_ADDITIONAL_ISSUES_TEXT); 
+      expect(wrapper.find('.table-issue-more-issues').text()).toEqual('View all 1 issues'); 
     }); 
+
+    it('renders Report data issue if no issues ', ()=> {
```
ttannis: A bit confused by this test, we render `ReportTableIssue` all the time now, so if we're checking for it that's what we'd want to check.

amundsen_application/proxy/issue_tracker_clients/jira_client.py
```diff
@@ -148,5 +141,22 @@ def _generate_remaining_issues_url(self, table_uri: str, issues: List[DataIssue]
             return ''
         # jira expects a ticket key in the query to default to, so pick the first one
         first_issue_key = issues[0].issue_key
-        search_query = urllib.parse.quote(SEARCH_STUB.format(table_key=table_uri))
+        search_query = urllib.parse.quote(SEARCH_STUB_ALL_ISSUES.format(table_key=table_uri))
         return f'{self.jira_url}/browse/{first_issue_key}?jql={search_query}'
```
ttannis: Is this url correct? I tried locally and this didn't take me to a link where I could view all issues, instead it took me to the most recent issue.

amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
```diff
@@ -27,12 +23,48 @@
...
     max-width: fit-content;
   }
-  .table-issue-more-issues {
-    margin: 0px 2px 0px 2px; 
+  .table-issue-status {
+    display: flex;
+    justify-content: flex-end;
+    flex-grow: 1.1;
+    min-width: fit-content;
+    margin: 0px 0px 0px 5px;
```
ttannis: Can we use `$spacer-1` instead of 5px? 

Not sure if this was an oversight in Figma, but when we were using Abstract our designers had given us component breakdowns and one of the concepts was that the default spacer size was "8px (the value of  $spacer-1)" and that all other spacing would be multiples of 8. So I'd question is this was actually supposed to be 5.

amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
```diff
@@ -27,12 +23,48 @@
...
+  }
+  .table-issue-priority {
+    margin: 0px 5px 0px 0px;
```
ttannis: Can comment about using `$spacer-1` instead of 5.

amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
csteez: Yeah, the colors came straight from the recommendations from the LDL - I just followed what the designs had + recommendations on opacity of color.

amundsen_application/static/js/components/TableDetail/TableIssues/index.tsx
```diff
@@ -42,33 +43,55 @@ export class TableIssues extends React.Component<TableIssueProps> {
...
+    ); 
+  }
+  
+  renderIssueTitle = () => {
+    return (
+      <div className="section-title title-3">
+        Issues
+      </div>
     );
   }
```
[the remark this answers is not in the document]
csteez: yeah good point, removed!

amundsen_application/static/js/components/TableDetail/TableIssues/index.tsx
```diff
@@ -78,13 +101,26 @@ export class TableIssues extends React.Component<TableIssueProps> {
...
-      return null;
+      return (
+        <div>
+          {this.renderIssueTitle()}
+          <div className="table-issues">
+            <div className="issue-banner">
+              {NO_DATA_ISSUES_TEXT}
+            </div>
+          </div>
+          { this.renderMoreIssuesMessage(this.props.total, this.props.allIssuesUrl)}
```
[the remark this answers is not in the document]
csteez: i replaced it with just rendering the "Report Data Issue" link.

amundsen_application/static/css/_variables-default.scss
csteez: oh neat, i didnt know i could use a hex value for rgba!

csteez: ty!!

csteez:
> Let me know your thoughts on these comments but I think the implementation is set. Most important concern is that when I tested the url locally for generating all issues, it didn't take me where I was expecting.

Could you tell me what table you were testing with? i wonder if you're looking at a table with a single issue. The link generated should take you to JIRA with the first issue selected, and other issues visible beneath.

amundsen_application/static/css/_variables-default.scss
ttannis: Ah my comment was just about defining the use of $rose80 here. But if you want to use hex values for rgba, what would work is appending some more values according to the list defined here:
https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4

In those examples they added the alpha codes to the beginning of the regular hex value, but when I tried this out what worked was adding it to the end. If for some reason you run into problems though, leaving the rgba is fine.

ttannis:
> > Let me know your thoughts on these comments but I think the implementation is set. Most important concern is that when I tested the url locally for generating all issues, it didn't take me where I was expecting.
> 
> Could you tell me what table you were testing with? i wonder if you're looking at a table with a single issue. The link generated should take you to JIRA with the first issue selected, and other issues visible beneath.

I tested on our event table. It took me to the first issue. Looking around a bit more I finally noticed this on the top right hand side, which was the only indication I saw of other issues:
image
But I don't see a list of the other issues anywhere else on the page. When I click "Return to search" though, then I see all the issues.

amundsen_application/proxy/issue_tracker_clients/jira_client.py
```diff
@@ -148,5 +141,22 @@ def _generate_remaining_issues_url(self, table_uri: str, issues: List[DataIssue]
...
+
+    def _sort_issues(self, issues: List[Issue]) -> List[DataIssue]:
```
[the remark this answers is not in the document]
csteez: I'm not going to add a test here, as I was unable to easily figure out a way to mock out the Issue class (defined by JIRA) in a way that would work with mypy. (ie it needs a real Issue, but i can't create one).

amundsen_application/proxy/issue_tracker_clients/jira_client.py
ttannis: Updated link works for me

amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
```diff
@@ -11,12 +11,9 @@
...
     flex-direction: row;
   }
-  .icon-red-triangle-warning  {
-    background: $red-triangle-warning;
-    margin: 0px 5px 0px 0px;
-  }
   .table-issue-link {
-    margin: 0px 5px 0px 0px; 
+    margin-right: $spacer-1; 
+    margin-right: $spacer-1;
```
ttannis:
```suggestion
```
Delete duplicate style

amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
```diff
@@ -34,10 +31,10 @@
     justify-content: flex-end;
     flex-grow: 1.1;
     min-width: fit-content;
-    margin: 0px 0px 0px 5px;
+    margin-bottom: $spacer-1;
```
ttannis: Did you mean for this to be `margin-left`? The status is squished up against the text.
image

Also in this case I think `margin-left: $spacer-2` would make things even easier to read as well, but up to you based on what looks closer to the designs.

csteez: good catch - i wasnt testing overflow in this round

PR #414, opened 2020-03-25
Prevent entering of ':' into search term

ttannis:
#### Context
1. Some users still entered advanced search syntax into the search bar.
2. If a colon is in the search term, it will break the filter endpoint.
3. A fix was made in amundsensearchlibrary to return HTTPStatus.BAD_REQUEST in this case

This PR aimed to create a better experience for this on the frontend, by surfacing information about non-successful requests. 

#### Initial Goal
The initial plan to was show different messages on the `SearchPage` for 5xx and 4xx errors. However the work effort turned out to be too great given current priorities. Our search bar searches batches the 3 requests to each resource endpoint and we would have to restructure this to be able to capture error codes only for the resources that fail  and store it on the search state for that resource.

#### Current Solution
Because of the scope creep my current solution pulls things back to just prevent entering a ':' in the search bar. 

If we ever re-introduce advanced search syntax this workaround can be deleted, and I think we can still invest effort into showing better error messages on the `SearchPage` for each resource request when bandwidth allows. 

### Tests

Added test for use case

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/common/SearchBar/constants.ts
+++ b/amundsen_application/static/js/components/common/SearchBar/constants.ts
@@ -4,3 +4,5 @@ export const PLACEHOLDER_DEFAULT = 'search for data resources...';
 export const BUTTON_CLOSE_TEXT = 'Close';
 
 export const SIZE_SMALL = 'small';
+
+export const INVALID_SYNTAX_MESSAGE = 'Your search termc contains invalid syntax';
--- a/amundsen_application/static/js/components/common/SearchBar/constants.ts
+++ b/amundsen_application/static/js/components/common/SearchBar/constants.ts
@@ -5,4 +5,4 @@ export const BUTTON_CLOSE_TEXT = 'Close';
 
 export const SIZE_SMALL = 'small';
 
-export const INVALID_SYNTAX_MESSAGE = 'Your search termc contains invalid syntax';
+export const INVALID_SYNTAX_MESSAGE = "Your search term contains invalid syntax ':'.";
--- a/amundsen_application/static/js/components/common/SearchBar/index.tsx
+++ b/amundsen_application/static/js/components/common/SearchBar/index.tsx
@@ -92,12 +92,14 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
   handleValueChange = (event: React.SyntheticEvent<HTMLInputElement>) : void => {
     const searchTerm = (event.target as HTMLInputElement).value.toLowerCase();
 
-    if (searchTerm.length > 0) {
-      this.props.onInputChange(searchTerm);
-      this.setState({ searchTerm, showTypeAhead: true });
-    }
-    else {
-      this.clearSearchTerm();
+    if (searchTerm.indexOf(':') < 0) {
+      if (searchTerm.length > 0) {
+        this.props.onInputChange(searchTerm);
+        this.setState({ searchTerm, showTypeAhead: true });
+      }
+      else {
+        this.clearSearchTerm();
+      }
     }
   };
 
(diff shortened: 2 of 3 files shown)


csteez:
So does this just clear if the search term contains ":"? Might that be frustrating/confusing for users? Am I right that this is triggered when the user types ":"?

amundsen_application/static/js/components/common/SearchBar/index.tsx
```diff
@@ -92,12 +92,14 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
...
-    else {
-      this.clearSearchTerm();
+    if (searchTerm.indexOf(':') < 0) {
+      if (searchTerm.length > 0) {
+        this.props.onInputChange(searchTerm);
+        this.setState({ searchTerm, showTypeAhead: true });
+      }
+      else {
+        this.clearSearchTerm();
+      }
```
danwom: Might be simpler to use a keypress handler and check `event.code` to see if a `:` was typed and then `event.preventDefault` to stop the `:` character from being added. This approach adds the character, then removes it after the value is updated.

ttannis:
> So does this just clear if the search term contains ":"? Might that be frustrating/confusing for users? Am I right that this is triggered when the user types ":"? What about including an error message?

It won't clear the search term. If/when someone type ":" nothing will update.

I explored adding an error message, but then I was unsure about the concept of writing more code than necessary to optimize the user experience for something we've communicated is deprecated -- because deprecated means not providing support. The motivation for putting in some effort though came from the fact that if the work effort to show the different messages on the `SearchPage` (ideal case -- concept is to distinguish between no results for your search vs. your search generated a request error) was too great, then I would at least prevent them form doing the thing that causes the error. 

But I can push that code needed to show and error message and I'll include it if no one is opposed to keeping it.

amundsen_application/static/js/components/common/SearchBar/index.tsx
ttannis: Whats your thought on letting them type the ":" then showing an error message + blocking submit, vs the keypress approach?

ttannis:
Error Message Demo
error_msg

csteez:
> Error Message Demo
> error_msg

Personally i think this is better, since it gives the user feedback on what's going on. But I defer to other folks on their opinion.

amundsen_application/static/js/components/common/SearchBar/constants.ts
```diff
@@ -4,3 +4,5 @@ export const PLACEHOLDER_DEFAULT = 'search for data resources...';
 export const BUTTON_CLOSE_TEXT = 'Close';
 
 export const SIZE_SMALL = 'small';
+
+export const INVALID_SYNTAX_MESSAGE = 'Your search termc contains invalid syntax';
```
danwom: typo `termc`

danwom: Maybe we can explain what characters are invalid?

```
Your search term contains invalid syntax or characters.
Alphanumeric characters [a-z] [0 - 9] . 
```

Not sure what the exact messaging here should be.

amundsen_application/static/js/components/common/SearchBar/index.tsx
danwom: On second thought I like showing the error message better since it lets the user know what's going on rather than just blocking them and making them think it's a bug.

amundsen_application/static/js/components/common/SearchBar/constants.ts
ttannis: Thanks, I leaned away from listing of which characters are valid in this can because every other special character is currently valid for the search term. I can update to specify `:` is invalid.

PR #415, opened 2020-03-31
Update JIRA Issue query + TableIssue UI

ttannis:
### Summary of Changes

This PR:
1. Updates the jira issue query to use exact match on the full `Table Key: ....` that we provide. This fixes an issue where unrelated JIRA issues were being displayed due to fuzzy matching.
2. Update the `TableIssue` component to show the `LoadingSpinner` instead of `ReportTableIssue`. With the updated design, it's cleaner for the whole section to be replaced with the `LoadingSpinner` as oppose to just the `Report Issue` text which led also led to a misaligned spinner.

### Tests

Updated tests to reflect changes.

### Documentation

N/A but big fix can be called out in next patch release.

### CheckList

### DIFF
--- a/amundsen_application/proxy/issue_tracker_clients/jira_client.py
+++ b/amundsen_application/proxy/issue_tracker_clients/jira_client.py
@@ -11,7 +11,7 @@ from amundsen_application.models.issue_results import IssueResults
 import urllib.parse
 import logging
 
-SEARCH_STUB_ALL_ISSUES = 'text ~ "{table_key}" order by createdDate DESC'
+SEARCH_STUB_ALL_ISSUES = 'text ~ "\\"{table_key}\\"" order by createdDate DESC'
 # this is provided by jira as the type of a bug
 ISSUE_TYPE_ID = 1
 ISSUE_TYPE_NAME = 'Bug'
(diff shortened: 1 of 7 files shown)


amundsen_application/proxy/issue_tracker_clients/jira_client.py
```diff
@@ -11,7 +11,7 @@
 import urllib.parse
 import logging
 
-SEARCH_STUB_ALL_ISSUES = 'text ~ "{table_key}" order by createdDate DESC'
+SEARCH_STUB_ALL_ISSUES = 'text ~ "\\"Table Key: {table_key} [PLEASE DO NOT REMOVE]\\"" order by createdDate DESC'
```
csteez, line 14: what's the rationale behind including the [PLEASE DO NOT REMOVE]? if there are existing tickets without that text (it was added in a later pr) they would no longer be linked

ttannis: Some would actually be. When the query was exact match on: 
`Table Key: <our-table-key>` 
an unrelated ticket was included that happened to havee:
`Table { key: <our-table-key>`

I confirmed that jira queries are not case sensitive. I did not get to the bottom of what JIRA still did an exact match on the unwanted ticket even though it had the  `{` character in the string, but I can only assume maybe they ignore that character. 

The only thing that worked in the end was to include an even more specific string.

PR #416, opened 2020-04-01
Update ProfilePage & My Bookmarks to new design

ttannis:
### Summary of Changes

This PR updates the `ProfilePage` and `MyBookmarks` components to the new design for dashboards. This was done by also extending `ResourceList` with more optional parameters to account for the having an optional `title` and a "View All/View Less" interaction if we are not showing pagination.

This PR only makes UI changes and does not to cover the changes we will have to make for the components to request dashboard data.

### Tests

TODO after initial review

### Documentation

N/A

### CheckList


ttannis:
@danwom this is ready for review. Unit tests are failing but will be updated after resolving any concerns about the components.

ttannis:
Updated with base branches and changed merge target to `feature/dashboard`

ttannis:
@danwom any comments on this, if not I'll be updating the tests to merge down tomorrow to continue on the bookmarks work.

amundsen_application/static/js/components/ProfilePage/index.tsx
```diff
@@ -96,43 +106,68 @@ export class ProfilePage extends React.Component<ProfilePageProps, ProfilePageSt
...
-      />
+      <>
+        <ResourceList
+          allItems={ own }
+          itemsPerPage={ ITEMS_PER_PAGE }
+          paginate={ false }
+          source={ OWNED_SOURCE }
+          title={`${OWNED_TITLE_PREFIX} (${own.length})`}
+          customFooterText={`${FOOTER_TEXT_PREFIX} ${own.length} ${OWNED_LABEL} ${resourceLabel}`}
+          customEmptyText={`${EMPTY_TEXT_PREFIX} ${OWNED_LABEL} ${resourceLabel}.`}
```
danwom, line 121: I'm kind of on the fence about this, but a helper function might help here.

amundsen_application/static/js/components/common/Bookmark/MyBookmarks/index.tsx
```diff
@@ -24,29 +29,56 @@ export class MyBookmarks extends React.Component<MyBookmarksProps> {
...
 
+  generateTabContent = (resource: ResourceType) => {
+    const bookmarks = this.props.myBookmarks[resource] || [];
+    return (
+      <ResourceList
+        allItems={ bookmarks }
+        source={ MY_BOOKMARKS_SOURCE_NAME }
+        itemsPerPage={ BOOKMARKS_PER_PAGE }
+        customEmptyText={ EMPTY_BOOKMARK_MESSAGE }
+      />
```
danwom, line 40: It's unclear from mocks if this section should paginate or click to expand. I tagged you in a comment in Figma.

amundsen_application/static/js/components/common/ResourceList/index.tsx
```diff
@@ -49,40 +58,70 @@ class ResourceList extends React.Component<ResourceListProps, ResourceListState>
...
+          <>
+            <ul className="list-group">
+              {
+                itemsToRender.map((item, idx) => {
+                  const logging = { source, index: startIndex + idx };
+                  return <ResourceListItem item={ item } logging={ logging } key={ idx } />;
+                })
+              }
+            </ul>
+            {
```
danwom, line 99: Also on the fence about this, but it might be worth considering building a separate component for the paginated version and the click to expand version. If the bookmarks on the home page are changed to "click to expand" then we can probably deprecate the paginated version of this.

amundsen_application/static/js/components/ProfilePage/index.tsx
ttannis: On the fence as well. I thought about helper functions for the strings but all it would do is concatenate a list of values with no extra logic so that seemed like a waste. I also thought of a helper function to render the components for each relationship but that would have come with so much boolean logic regarding what constants to use.

I will leave it as is unless I see a clear way to do something better.

amundsen_application/static/js/components/common/Bookmark/MyBookmarks/index.tsx
ttannis: Noted. I will update this in the followup PR base don the clarification;

amundsen_application/static/js/components/common/ResourceList/index.tsx
ttannis: The paginated version is still used on `SearchPage`. 

I'm open to splitting them as well because this component is pretty much at the limit of how much branching logic I'd want to commit to maintaining. But I'm also inclined to accept this the way it is, under the expectation that we won't have any new concepts introduced in this area any time soon. 

I will clean up the render logic by splitting it into helper methods. This can at least make things a bit more intuitive to either maintain or copy-paste all this added logic into its own component.

PR #424, opened 2020-04-10
Update Dashboard Search UI

ttannis:
### Summary of Changes

This PR updates the search UI for dashboards to near its final state.

**Changes Include**
1. Update python code according to finalized searchlibrary work.
2. Add similar support for configuring the display names and icons for different dashboard sources. The previous `supportedDatabases` option has been renamed `supportedSources` and the same pattern is now used with dashboards. `config_utils` and related documentation has been updated accordingly. 
3. Updated `DashboardListItem`
image
4. Updated inline search results code to take a ReactNode as the title, allowing us to update the inline search result dashboard item for more control over the title.
image

### Tests

Updated tests according to changes.

### Documentation

Updated application configuration doc according to changes.

### CheckList
Make sure you have checked **all** steps below to ensure a timely review.
- [x] PR title addresses the issue accurately and concisely. Example: "Updates the version of Flask to v1.0.2"
    - In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).
- [x] PR includes a summary of changes, including screenshots of any UI changes. 
- [x] PR adds unit tests, updates existing unit tests, __OR__ documents why no test additions or modifications are needed.
- [x] In case of new functionality, my PR adds documentation that describes how to use it.
    - All the public python functions and the classes in the PR contain docstrings that explain what it does
- [x] PR passes all tests documented in the [developer guide](https://github.com/lyft/amundsenfrontendlibrary/blob/master/docs/developer_guide.md#testing)

PR #425, opened 2020-04-15
Add Dashboard Preview UI Support

ttannis:
### Summary of Changes

This PR adds the UI support for dashboard previews. There is also some code cleanup from https://github.com/lyft/amundsenfrontendlibrary/pull/424

#### Dashboard Preview UI
1. Created an `ImagePreview` component and added it to `DashboardPage`. This component supports showing the url if it exists, and supports showing error messages from the API if the request fails.
2. Added support in `ducks` to request the dashboard image preview. This is a special case where we already know what the url for each resource us going to be -- so the success response just returns that url, and the failure case has to do some extra parsing of the response to additionally return the error code and error message for the UI.

#### Code Cleanup
1. Updated the links to `DashboardPage` to put the `uri` in the search params
2. Resolved issue of `title` vs `titleNode` when the property is a `React.ReactNode`.

#### Future Work
1. `ImagePreview` loading state is intended to use the shimmer effect design.
2. The design or api or implementation that supports this should be revisited -- the issue is that because Redux calls the API to support knowing when a component is ready to load *and* the `` tag  requests its `src` when it renders we double request the image. When connections are slow, this results in a slight pause where neither the loading UI nor the image shows up and the UI will be blank for quick moment.

### Tests

Added test to cover main functionality of new code.

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/api/preview/dashboard/v0.py
+++ b/amundsen_application/api/preview/dashboard/v0.py
@@ -45,6 +45,7 @@ def get_preview_image(uri: str) -> Response:
                          mimetype='image/jpeg',
                          cache_timeout=app.config['DASHBOARD_PREVIEW_IMAGE_CACHE_MAX_AGE_SECONDS'])
     except FileNotFoundError as fne:
+        LOGGER.exception('FileNotFoundError on get_preview_image')
         return make_response(jsonify({'msg': fne.args[0]}), HTTPStatus.NOT_FOUND)
     except Exception as e:
         LOGGER.exception('Unexpected failure on get_preview_image')
--- a/amundsen_application/static/js/components/DashboardPage/ImagePreview/constants.ts
+++ b/amundsen_application/static/js/components/DashboardPage/ImagePreview/constants.ts
@@ -1,0 +1,1 @@
+export const DEFAULT_ERROR_MESSAGE = 'An error ocurred loading the image preview, please refresh the page';
--- a/amundsen_application/static/js/components/DashboardPage/ImagePreview/index.spec.tsx
+++ b/amundsen_application/static/js/components/DashboardPage/ImagePreview/index.spec.tsx
@@ -1,0 +1,128 @@
+import * as React from 'react';
+
+import { shallow } from 'enzyme';
+
+import { ImagePreview, ImagePreviewProps, mapStateToProps, mapDispatchToProps } from './';
+
+import LoadingSpinner from 'components/common/LoadingSpinner';
+
+import globalState from 'fixtures/globalState';
+import { errorPreviewStateNoMessage, errorPreviewStateWithMessage } from 'fixtures/dashboard/preview';
+
+import * as Constants from './constants';
+
+describe('ImagePreview', () => {
+  const setup = (propOverrides?: Partial<ImagePreviewProps>) => {
+    const props = {
+      uri: 'test:uri/value',
+      url: 'someUrl',
+      errorMessage: 'oops',
+      errorCode: undefined,
+      isLoading: false,
+      getPreviewImage: jest.fn(),
+      ...propOverrides,
+    };
+
+    const wrapper = shallow<ImagePreview>(<ImagePreview {...props} />)
+    return { props, wrapper };
+  };
+
+  describe('componentDidMount', () => {
+    it('triggers logic for preview image to load', () => {
+      const { props, wrapper } = setup();
+      expect(props.getPreviewImage).toHaveBeenCalledWith({ uri: props.uri });
+    })
+  })
+
+  describe('render', () => {
+    let props;
+    let wrapper;
+    let element;
+    beforeAll(() => {
+      const setupResult = setup();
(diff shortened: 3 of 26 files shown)


danwom:
This morning we discussed several approaches on how to load this image including

1) **Fetch image from XHR, and set the img src as a data url string.**
Pros: Can easily tell when the image is ready and manage state.
Cons: Can't use the browser's built in cache. Data url string will be very long. There may be some length restrictions depending on the browser.

2) **Rely on the img src only to fetch the image.**
Pros: Can use built in browser cache. Can use the "onload" event or "complete" property to check status.
Cons: Slightly less control of status code / error response.
[Tutorial](https://idiallo.com/javascript/img-detect-loaded) 
[Stack Overflow](https://stackoverflow.com/questions/14373683/how-to-show-image-only-when-it-is-completely-loaded)

Given this information I would go with approach 2.

ttannis:
> This morning we discussed several approaches on how to load this image including
> 
> 1. **Fetch image from XHR, and set the img src as a data url string.**
>    Pros: Can easily tell when the image is ready and manage state.
>    Cons: Can't use the browser's built in cache. Data url string will be very long. There may be some length restrictions depending on the browser.
> 2. **Rely on the img src only to fetch the image.**
>    Pros: Can use built in browser cache. Can use the "onload" event or "complete" property to check status to show image.
>    Cons: Slightly less control of status code / error response. Might interact strangely with React's render cycle. Might have to keep a reference to an image ref and update visibility.
>    [Tutorial](https://idiallo.com/javascript/img-detect-loaded)
>    [Stack Overflow](https://stackoverflow.com/questions/14373683/how-to-show-image-only-when-it-is-completely-loaded)
> 
> Given this information I would go with approach 2.

At this point I just have a strong preference that this be improved, but I don't believe we should make this decision right now in an Eng silo. So I don't have a preference on how this PR should be changed without circling back with Design & Product to understand if they have different opinions on which pros are stronger than which cons.

For example here are other perspectives: 
1. If we have can have more control over the user experience for loading state, does caching need to be optimized? My personal opinion leans towards yes but others could disagree.
2. Is the caching worth having less control over the UI if the request fails? My personal opinion leans towards not ruling this out without a sense of how often the request can fail. We currently only have anecdotal information my experience and Jin's experience, so instead we could also try to collect this data during the alpha stage before making a decision. We have seen from our issues with data preview that users do not have a great experience if they don't understand why things fail, even if that failure is not Amundsen's fault. 

Given all of that though, it might be a waste to deeply review this PR for anything besides immediate bugs. So perhaps instead we could align on an approach to the next steps, my current perspective is:
1. Consider leaving this implementation as part of alpha target week. At the end of the week we can check logs to have a better understanding of how many times an exception was logged on the preview endpoint. 
2. Use that data along with the Eng pros & cons to revisit the UX decisions on the feature.
3. Implement improved version before GA

danwom:
I believe we can satisfy all of the design requirements with method 2. You may need to make some modifications to the outline below, as I have not tested this myself. In theory it should work. LMK what you think. 

- On mount set `isLoading` to true and use this to show and hide the loading animation as well as the ``. The `` would need to use a `visibility: hidden` or equivalent so the image request is fired.
- Add an `onload` event to the `` and set the `isLoading` to false. This should hide the loading animation, and set `visibility: visible` on the ``. 
- Add an `onerror` to the `` and use this to show the error message if something happens.

ttannis:
> I believe we can satisfy all of the design requirements with method 2. You may need to make some modifications to the outline below, as I have not tested this myself. In theory it should work. LMK what you think.
> 
> * On mount set `isLoading` to true and use this to show and hide the loading animation as well as the ``. The `` would need to use a `visibility: hidden` or equivalent so the image request is fired.
> * Add an `onload` event to the `` and set the `isLoading` to false. This should hide the loading animation, and set `visibility: visible` on the ``.
> * Add an `onerror` to the `` and use this to show the error message if something happens.

I'm aware of these methods we can leverage but did not see if `onerror` isn't capable of fine grained error handling to obtain either the error message or error code from the response. So me the next question isn't about how to we'd code up the alternative but first understanding if fine grained error handling even matters so that we can avoid the same pitfalls from the past.

The pitfall I'm referencing is that initial data preview did not have fine grained error handling to start and as time went on we went through multiple iterations where it was requested that we show different information for different kinds of errors. We were reactive to this instead of proactive, and introduced whatever code made things work in our custom preview client and in the application.

So instead of choosing the method we prefer for dashboard preview right now and risking having to re-hack our implementation down the line, since we have the time before GA we could be proactive about trying to understand how often errors occur, what kinds of errors they might be, but most importantly just get an agreement between Eng, Product & Design on if we see a need for fine grained error processing (the data would help make that decision if anyone had any concerns).

I'm not opposed to method 2 by the way, more so I think we have enough time to try and get a better picture of a few more unknowns.

ttannis:
@danwom let me know if you have any thoughts on this before I merge. I'm also moving forward with the next steps that were discussed.

PR #426, opened 2020-04-16
Connect dashboard page with API

danwom:
### Summary of Changes

Connected the dashboard stub API to real API. This endpoint will also sanitize the tables and owners within the response. This PR also adds several fields to the dashboard page, including the recent view count, last successful run, and tables.

### Tests

### Documentation

### CheckList


amundsen_application/api/utils/metadata_utils.py
```diff
@@ -74,6 +74,20 @@ def marshall_table_full(table_dict: Dict) -> Dict:
     return results
 
 
+def marshall_dashboard_full(dashboard_dict: Dict) -> Dict:
+    """
+    Cleanup some fields in the dashboard response
+    :param dashboard_dict: Dashboard response from metadata service.
+    :return: Dashboard dictionary with sanitized fields, particularly the tables and owners.
+    """
+    # TODO - Cleanup https://github.com/lyft/amundsen/issues/296
```
ttannis, line 83: I notice this ticket is closed though, can you reopen or file a new one with the information so your ideas don't get lost.

amundsen_application/static/js/components/DashboardPage/index.spec.tsx
```diff
@@ -14,38 +14,47 @@ describe('DashboardPage', () => {
...
-          user_id: 'dwon@lyft.com',
-        }],
-        frequent_users: [{
-          read_count: 10,
-          user: {
-            email: 'dwon@lyft.com',
-            display_name: 'Daniel Won',
-            profile_url: "",
-            user_id: 'dwon@lyft.com',
+        badges: [],
```
ttannis, line 19: nit: creating the one time fixtures for the different chunks of data corresponding to at least the most important use cases to test would make tests across the app a bit less cumbersome to set up.

amundsen_application/static/js/components/DashboardPage/index.tsx
```diff
@@ -81,6 +92,14 @@ export class DashboardPage extends React.Component<DashboardPageProps, Dashboard
     if (isLoading) {
       return <LoadingSpinner/>;
     }
+    if (this.props.statusCode === 500) {
+      return (
+        <div className="container error-label">
+          <Breadcrumb />
+          <label>Something went wrong...</label>
```
ttannis, line 99: Feel free to add anything better here if you want. This is an artifact from the early days and you don't have to feel obligated to propagate it.

amundsen_application/static/js/components/DashboardPage/index.tsx
```diff
@@ -113,28 +132,48 @@ export class DashboardPage extends React.Component<DashboardPageProps, Dashboard
...
+                <div>
+                  { formatDateTimeShort({ epochTimestamp: dashboard.created_timestamp}) }
+                </div>
+                <div className="section-title title-3">Owners</div>
+                <div>
+                  {
+                    dashboard.owners.map(owner =>
+                      <Link
+                        key={owner.user_id}
+                        to={`/user/${owner.user_id}`}>
```
ttannis, line 151: Add the `source` parameter like we have on TableMatadata owners for tracking?

amundsen_application/static/js/ducks/dashboard/sagas.ts
```diff
@@ -8,8 +8,8 @@ import { GetDashboard } from './types';
 export function* getDashboardWorker(action): SagaIterator {
   try {
     const uri = action.payload.uri;
-    const dashboard = yield call(API.getDashboard, uri);
-    yield put(getDashboardSuccess(dashboard));
+    const { dashboard, statusCode } = yield call(API.getDashboard, uri);
```
ttannis, line 11: Is `statusCode` actually available on this result? If I'm reading this correctly this will be the `response.data` from the request and my understanding is that the status code would be found at `response.status`.

amundsen_application/api/dashboard/v0.py
```diff
@@ -1,11 +1,12 @@
 import logging
```
ttannis, line 1: nit: Are you opposed to moving all of this to `api/metadata/v0`? This is still a call to the metadata service, and all other dashboard support for the MVP will still be calling `api/metadata/v0` since the code can be reused

amundsen_application/api/utils/metadata_utils.py
danwom: I re-read the thread and I think this warrants further discussion. I have a half finished proposal doc about some of these fields. Either way I think this can be cleaned up at a later time. 

Some preliminary ideas about this:

1) Every resource should have a unique identifier. We can use the same field, `uri` for each, or use `uri` for dashboard, `key` for table and `user_id` for users. I prefer the former.

2) Deprecate `display_name` which is a generated field on FE Flask. In React, we could add a `getDisplayName(user)` function which will be something like
```
function getDisplayName(user) {
  if (!!user.full_name != null) return user.full_name;
  if (!!user.first_name && !!user.last_name) return ...
  else return user.email;
}
```

amundsen_application/static/js/components/DashboardPage/index.tsx
danwom: Don't have anything better to put at the moment :)

amundsen_application/static/js/ducks/dashboard/sagas.ts
danwom: Good catch. Will fix.

amundsen_application/api/utils/metadata_utils.py
ttannis: Makes sense, also for the identifier the solid case for the same field is bookmarks. You will see in the PR  I push later -- with no changes we'd be stuck writing boolean logic for processing dashboard bookmarks with `uri` and table bookmarks with `key`. I had to introduce another temporary workaround in the python layer for this. 

I have thought on the 2nd item as well but to keep this PR focus on track that can be discussed later in your proposal.

amundsen_application/api/dashboard/v0.py
danwom: I can move it back to `metadata`, though I would like to eventually revisit this file. It is becoming a monolithic file with the majority of our APIs here. For example we could move all Bookmark APIs into a separate file, and this would support fetching, adding, and removing bookmarks for any resource type, dashboard, table, etc.

ttannis: If and when things are revisited let's also not forget about https://github.com/lyft/amundsen/issues/307

amundsen_application/static/js/ducks/dashboard/reducer.ts
```diff
@@ -21,14 +21,20 @@ export function getDashboard(payload: { uri: string, searchIndex?: string, sourc
 export function getDashboardFailure() {
   return {
     type: GetDashboard.FAILURE,
-    payload: {}
+    payload: {
+      isLoading: false,
```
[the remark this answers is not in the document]
danwom: I copied the pattern used in by the table metadata API, which is probably outdated. Is the Data Preview our best example of how to handle failure? I'll try using that pattern instead.

amundsen_application/static/js/fixtures/globalState.ts
```diff
@@ -208,7 +208,9 @@ const globalState: GlobalState = {
         tag_count: 15,
       }
     ],
-    isLoading: false,
+    isLoadingAllTags: false,
+    isLoadingTags: false,
```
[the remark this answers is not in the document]
danwom: I am onboard with renaming to `resourceTags`. I'm not opposed to regrouping or nesting. How about something like

```
globalState = {
  ...
  allTags: { ... },
  resourceTags: { ... },
}
```
vs

```
globalState = {
  ...
  tags: {
    allTags: { ... },
    resourceTags: { ... },
  }
}
```

Personally, I am in favor of the second one.

amundsen_application/static/js/ducks/tags/api/v0.ts
```diff
@@ -0,0 +1,52 @@
...
+    .then((response: AxiosResponse<GetDashboardAPI>) => {
+      return getDashboardTagsFromResponseData(response.data)
+    });
+  }
+}
+
+/* TODO: Typing this method generates redux-saga related type errors that needs more dedicated debugging */
+export function updateTableTag(tagObject, resourceType: ResourceType, uriKey: string) {
+  let url = "";
+  if (resourceType === ResourceType.table) {
```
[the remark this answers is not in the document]
danwom: I agree, but I think I'll leave this for a follow up. I wonder if it also makes sense to approach the search APIs in a similar way. Probably need to spend more time thinking about it first, however.

amundsen_application/static/js/ducks/dashboard/reducer.ts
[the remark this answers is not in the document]
ttannis: Yeah we don't have consistency, but recalling we had a previous issue with the fact that logic for failures was handled in so many different places and settled on `sagas` as the ideal  place in case the failure response needs special logic. But if doesn't need special logic and the failure case is a default value of some kind I think it's also fair to hardcode that in the reducer, like setting something back to an initial or default state.

Yes I'd say the preview logic is our best example for the most complex failure case we currently have, which requires processing the failure response.

amundsen_application/static/js/fixtures/globalState.ts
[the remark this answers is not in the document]
ttannis: Fine with me.

amundsen_application/static/js/ducks/tags/api/v0.ts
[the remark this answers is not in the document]
ttannis: For the search APIs one thing that makes me hold off in that area is that the search service might also eventually refactor endpoints so that the same endpoints can be used for all resources to make things scale better as we add more resource or the community adds more resources to their custom versions.  I'd want to see exactly what that will look like before messing anything up here.

amundsen_application/static/js/ducks/dashboard/sagas.ts
```diff
@@ -8,10 +8,11 @@ import { GetDashboard, GetDashboardPreview, GetDashboardPreviewRequest } from '.
...
   try {
     const uri = action.payload.uri;
-    const dashboard = yield call(API.getDashboard, uri);
-    yield put(getDashboardSuccess(dashboard));
-  } catch (e) {
-    yield put(getDashboardFailure());
+    const { data, status } = yield call(API.getDashboard, uri);
+    yield put(getDashboardSuccess(data.dashboard, status));
+  } catch (error) {
+    const { status } = error;
```
ttannis, line 14: Is the status not on `error.response.status`?

amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
```diff
@@ -18,41 +18,20 @@ export type TableDataAPI= { tableData: TableData; } & MessageAPI;
...
-      data: {
-        key: tableKey,
-        tag: tagObject.tagName,
-      },
-    }
-  });
-  return updatePayloads.map(payload => { axios(payload) });
-}
 
+// TODO - Revisit the necessity of filtering out tags from main metadata
```
ttannis, line 26: The idea was to have as little bloat as possible in the application state. I didn't find it reasonable to set things up such that `globalState.tableMetadata.tableData.tags` and `globalState.<tag-state>.tags` would store the same information given that information could be a list of objects of unknown length.

amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
```diff
@@ -18,41 +18,20 @@ export type TableDataAPI= { tableData: TableData; } & MessageAPI;
...
 export function getTableData(tableKey: string, index?: string, source?: string ) {
   const queryParams = getTableQueryParams(tableKey, index, source);
   return axios.get(`${API_PATH}/table?${queryParams}`)
   .then((response: AxiosResponse<TableDataAPI>) => {
     return {
       data: getTableDataFromResponseData(response.data),
       owners: getTableOwnersFromResponseData(response.data),
-      tags: getTableTagsFromResponseData(response.data),
+      tags: response.data.tableData.tags,
```
ttannis, line 34: Still needs alphabetical sorting.

amundsen_application/static/js/ducks/dashboard/sagas.ts
ttannis: Example: https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/ducks/tableMetadata/api/v0.ts#L141-L159

I know we had at least a couple bugs in the past from not processing the `AxiosError` correctly. Type info: https://github.com/axios/axios/blob/master/index.d.ts#L85

amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
danwom: I think I can actually remove `tags` field here and add it back into the `data` which filters it out. Then in the Tags reducer and sagas we can read `response.data.tableData.tags` instead of `response.tags`. I'll add this as a follow up PR since there are a lot of places that need to be changed.

amundsen_application/static/js/ducks/dashboard/sagas.ts
danwom: I initially followed the response in `getPreviewDataWorker` which did 
```
catch (error) {
    const { data, status } = error;
```
But I did not see the code mentioned in your example which returned a cleaned error object.

amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
ttannis: See comment above: https://github.com/lyft/amundsenfrontendlibrary/pull/426#discussion_r413414960

Please consider the `globalState` if you make changes. It is recommended to set up application state such that each object is stored once in the globalState.

PR #430, opened 2020-04-22
Update dashboard preview UI

ttannis:
### Summary of Changes
This PR updates the dashboard preview logic and UI.

#### Python 
After exploring our failure logs for dashboard previews, I found that `'web_preview_image': None` is possible and all of our errors that occurred so far in staging were an uncaught `FileNotFoundError`. I update the logic to check for this case, and added tests for these two failure cases.

#### React App
The error investigation confirms that we are only dealing with file not found errors. This made a clearer case to update `ImagePreview` in the following ways:
1. Remove the Redux logic associated with `ImagePreview`
2. When the html `` requests its `src` we use the `onLoad` handler to know when to stop showing the loading state, and the `onError` handler to know when to show the default error message. `onError` does not receive any information about the network request so fine grained error handling will not be an option in the future for this component. However this is reasonable because our testing indicates we are dealing with 1 kind of error.
3. If an error occurs, the `ImagePreview` will instead present the dashboard url and instruct the user to view the dashboard directly.
4. Minor css updates to `ImagePreview`.

### Tests

Update tests to cover modified logic.

### Documentation

N/A

### CheckList


amundsen_application/dashboard_preview/mode_preview.py
```diff
@@ -69,4 +69,8 @@ def _get_preview_image_url(self, *, uri: str) -> str:
         if web_preview_image_key not in result:
             raise FileNotFoundError('No preview image available on {}'.format(uri))
 
-        return result[web_preview_image_key]
+        image_url = result[web_preview_image_key]
```
ttannis, line 72: cc @jinhyukchang I was able to reproduce some dashboard preview failures in staging that all reported the same error `Encountered exception: 'NoneType' object has no attribute 'replace'`. 

I found that this was an error coming from our internal requests library -- not Mode -- and the reason was that the `web_preview_image` key sometimes exists but has `None` as the value.

ttannis:
@danwom I have to pull some changes from the `connect_dashboard` branch after it gets merged to finish ` TODO, but this is ready for review.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
```diff
@@ -1,76 +1,64 @@
...
-      content = <LoadingSpinner/>;
-    }
-    else if (errorCode) {
-      content =  <div className="loading-error-text body-placeholder">{ errorMessage }</div>;
-    }
-    else {
-      content = <img src={this.props.url} height="auto" width="100%" />;
-    }
+  onError = (e: React.SyntheticEvent<HTMLImageElement>) => {
+    this.setState({ isLoading: false, hasError: true });
```
danwom, line 35: Would it be possible to check the error code from `e`?

ttannis: No we don't get any network information from these kind of events, just some of the usual DOM related things.

PR #432, opened 2020-04-27
Dashboard Page Updates

ttannis:
### Summary of Changes

This PR covers a set of updates to the dashboard page UI.

1. Show `last_run_state` as a `Flag`. The default badge colors have also been updated to match the updated guidelines for data colors.
    
    _Note: This support brittle in the way that it relies on everyone's databuilder implementation to ingest the exact text `succeeded`. Will file an issue in the main repo for further discussion, to propose coverting this value to a boolean._

2. Update the UI for when there are no owners. This change required some css changes and additional optional parameters to `AvatarLabel` to get the 3rd-party `Avatar` component to render according to the design. I intentionally did not attempt to make this its own component as I think this will be best to incorporate when we update `TableDetail/OwnerEditor`.
    

3. Add icon to the `DashboardPage` by leveraging the `getSourceDisplayIcon()` helper. This required updating some types and fixtures to include `product` on the dashboard detail interface.
    

4. Add a UI that directs users to the dashboard to add descriptions. Design has opted to only include the "Add" prompt to assist with the goal of getting more metadata, but there will be no prompt to edit the description. 
    

### Tests

Updated tests to cover additions and modifications.

### Documentation

N/A

### CheckList


ttannis:
@danwom some dashboard page updates

amundsen_application/static/js/components/DashboardPage/index.tsx
```diff
@@ -156,6 +182,15 @@ export class DashboardPage extends React.Component<DashboardPageProps, Dashboard
                       </Link>
                     )
                   }
+                  {
+                    dashboard.owners.length === 0 &&
+                    <AvatarLabel
+                      avatarColor='#CACAD9'
+                      avatarTextColor='#CACAD9'
+                      labelClass='text-placeholder'
+                      label={NO_OWNER_TEXT}
```
danwom, line 191: Nit: I'm not the biggest fan of passing color codes directly to components. I know the Avatar component directly adds color styles to the elements. We could get around that by adding some wrapper class `gray-avatar` (?) and assigning an `!important` color in css to one of the child elements.

amundsen_application/static/js/components/DashboardPage/index.tsx
```diff
@@ -184,8 +219,12 @@ export class DashboardPage extends React.Component<DashboardPageProps, Dashboard
                 <div className="section-title title-3">Last Run</div>
                 <div>
                   { formatDateTimeShort({ epochTimestamp: dashboard.last_run_timestamp }) }
-                  <div className="last-run-state title-3">
-                    { dashboard.last_run_state }
+                  <div className="last-run-state">
+                    <Flag
+                      caseType='sentenceCase'
+                      text={ dashboard.last_run_state }
+                      labelStyle={ dashboard.last_run_state === 'succeeded' ? 'success' : 'danger' }
```
danwom, line 226: nit: Could move `succeeded` to constants file. Also could consider creating a utility function to map run-status to label style. We've only seen two statuses so far, but it's possible there are more statuses especially if we integrate other vendors. For example `succeeded, failed, waiting, canceled, in progress`.

CC: @jinhyukchang

amundsen_application/static/js/components/DashboardPage/index.tsx
```diff
@@ -139,14 +149,30 @@ export class DashboardPage extends React.Component<DashboardPageProps, Dashboard
...
-            </div>
+            {
+              hasDescription &&
+              <div>
+                { dashboard.description }
+              </div>
+            }
+            {
+              !hasDescription &&
+              <span>
```
danwom, line 160: nit: Can we reuse the `div` from the `hasDescription &&` block or does this need to be a `span`?

amundsen_application/static/js/components/DashboardPage/index.tsx
ttannis: Since we have the new `labelClass` if I add an `avatarClass` and set and apply it correctly it will work.

amundsen_application/static/js/components/DashboardPage/index.tsx
ttannis: I will pull the `labelStyle` logic into its own method.

I'll file an issue for further discussion of some concerns around `last_run_state`, which IMO would be best to take action on only when we actually get to the point where we or a community member wants to integrate another vendor.

amundsen_application/static/js/components/DashboardPage/index.tsx
ttannis: `<span>` is probably not needed will remove.

PR #435, opened 2020-04-30
Add a configurable BASE_URL to allow Amundsen to be served outside of the root URL

dmateusp:
### Summary of Changes

Full description of the issue: https://github.com/lyft/amundsen/issues/424

On the Python (Flask) side, I leveraged `url_for` in index.html, a BASE_URL configuration variable and a middleware class. I also took the routes defined in `amundsen_application/api/__init__.py` and standardized them by using the same Blueprints mechanism as the rest was using.

On the Javascript side, I added an importable `axios` instance to inject the BASE_URL into the calls. Note that I have very little experience on front-end, I got some help from @rebeccarich at Earnest but I realized there's many things I don't understand there. For example I try to get the environment variable [here](https://github.com/lyft/amundsenfrontendlibrary/compare/master...EarnestResearch:424_add_configurable_root?expand=1#diff-23aa9972769ba4f31e19eccf198e6880R9), but  I imagine it's not getting picked up because it's client-side ? Any feedback on which way we should do this?

On the scss side, I'm injecting the variable, but in this case it's done at compile time as far as I understand, which is undesirable  (we probably don't want people building Amundsen themselves everytime they need BASE_URL to be set), also not sure about the right way of doing it

Happy to get initial feedback

### Tests

Updates the tests to use the new axios

### Documentation

TODO: add docs

### CheckList


dmateusp:
hey @ttannis I'm going to close this PR for now

PR #440, opened 2020-05-07
Make User ID the CTA for user operations

verdan:
### Summary of Changes
This PR covers the following things:
- Remove custom User model and use the one from amundsencommon. 
- Customize the label for the User ID in the frontend. (where people are not using email address, they can specify whatever they want. In case of ING the user id is the 6 Characters Corporate Key. 
- Makes sure to use the user id for all CTA operations for the user, instead of email address. 

### Tests

### Documentation

### CheckList


verdan:
@ttannis @feng-tao please do review and merge https://github.com/lyft/amundsencommon/pull/36 before this one.

verdan:
The follow up fix is to fix the parameters name in amundsenmetadata service. It says `user_email` as the parameter name. @mgorski

verdan:
@feng-tao @ttannis @danwom could you please review this PR? this is based on the new changes in amundsencommon.

tests/unit/api/search/test_v0.py
```diff
@@ -286,6 +287,7 @@ def test_search_user_success(self) -> None:
         Test request success
         :return:
         """
+        self.maxDiff = None
```
feng-tao, line 290: do we use this?

verdan: Nope. this was only to debug the test case. Will remove this.

amundsen_application/models/user.py
```diff
@@ -1,120 +1,15 @@
...
-        if self._str_no_value(data.get('user_id')):
-            raise ValidationError('"user_id" or "email" must be provided')
+from marshmallow import ValidationError
 
 
 def load_user(user_data: Dict) -> User:
     try:
         schema = UserSchema()
+        # To make sure we pass frontend configuration to amundsen common models
+        user_data['GET_PROFILE_URL'] = app.config['GET_PROFILE_URL']
```
ttannis, line 12: Just an FYI for this comment I left in amundsencommon, if we ever want top consider removing the need for this configuration -- which was initially another workaround. https://github.com/lyft/amundsencommon/pull/36/files#r425595217

amundsen_application/static/js/config/config-custom.ts
```diff
@@ -16,10 +16,11 @@ const configCustom: AppConfigCustom = {
     feedbackEnabled: false,
     notificationsEnabled: false,
   },
-  indexDashboards: {
-    enabled: false,
+  users: {
+    indexUsersEnabled: true,
+    idLabel: 'email address'
```
ttannis, line 21: I'm having trouble thinking of the best way to configure this `idLabel` concept. My thoughts are as follows:
1. This configuration exists not as an actual feature configuration about application functionality, rather it exists to configure a particular string in a particular piece of UI. There are more standard/scalable ways of doing this in an application, but amundsenfrontend currently lacks that. But the scope is not something that can be completed in a few days.
2. In addition, the need to configure this string comes from the fact that our `OwnerEditor` UI is still not our officially designed version -- at some point this UI will be updated and we may no longer need that configuration. The scope of that work is also not something that can be completed in a few days.
3. Since the only two options I can think of to avoid this configuration are not possible today, I understand why we have to add it for your use case. However because the configuration will become obsolete I prefer to make the least amount of modifications possible for this configuration. This preference also come back to my first point, regarding the fact that I don't prefer to couple configuration based on what they happen to be about ("users") vs. what the configurations functionality is("control application functionality" vs "UI string customization support")

In summary, I would be in favor of having a standalone `userIdLabel: 'email address'` until the application can better support your use case through more sophisticated means.

verdan: @ttannis I had the same confusion about this in my mind while implementing this. Few points that I thought:
- I was trying to group together the related config variables. Hence making the user namespace instead of introducing all variables on the root level. 
- I want to avoid creating noise and then removing the top level config variables, in the future. Creates confusion in my opinion. 
- Since the refactoring is something that can't be done without proper refactoring around the components, the quick and efficient fix. 

In the end, I don't think it makes any huge difference IMHO. Creating a namespace for users might be useful in the future though. Thoughts?

ttannis: I don't mind namespaces, `indexUsers` in itself functions that way right now. But it exists for the purpose of configuring application functionality related to the users feature. In my 3rd comment above I categorize what you want to accomplish as something that is actually different. 

What you accomplish with `idLabel` is UI string customization support. I suggest decoupling because this is a different concept that does not apply exclusively to the Amundsen People feature. If you had to configure any other string that just happened to be somewhere else, where would that configuration go? And then the next, and then all the ones after that? I hope that this example of considering how string configuration would work at scale helps to understand why I point out that this configuration is coupling concepts for two different features.

In addition because the UI you need to modify was never intended to be permanent. On top of not being convinced the two configurations should be coupled, the difference I think it makes is that we'd be making a destabilizing configuration change (by changing the nested structure) for a workaround. Instead, we can make the least destabilizing change knowing that we have more and more reason to invest in the proper support for customizing your text.

verdan: ah, like that. I got it what you mean. Makes sense. Will make the change. Thank you for the explanation. 🙂

verdan:
@ttannis ready for another pass

ttannis:
Also @verdan FYI there are a couple more bugs that we have noticed working through testing out our dashboard feature. In case this is relevant for your team to know, the next frontend release for your case could come as early as end of this week if we don't discover any more issues that affect the user experience.

verdan:
> Thanks @verdan. Also in the future when your team has another clear use case for needing to update specific text in the UI let us know. It becomes easier to prioritize the proper support I mentioned when we first off know that other people have other use cases, and a good understanding of what those use cases look like.

Sure thing @ttannis. Raised this issue a year ago btw. https://github.com/lyft/amundsen/issues/34 :neutral_face:

verdan:
No worries, we are deploying to production using the latest master now. Not using the pip version anymore because of the limitation of frontend configurations. So if this can be merged in master, that's already good enough for us.

PR #443, opened 2020-05-13
Fixes DataSource or Dashboard Icons not rendering on Firefox

Golodhros:
### Summary of Changes
- Updates resource icons with background image approach

Accessibility improvements:
- Adds a main landmark to the application
- Adds labels to filter inputs
- Adds language to the html tag and some other missing meta tags (to complete)

### Why?
As mentioned in https://jira.lyft.net/browse/AMD-1056, the current implementation of our icon images is based on inserting content into image tags. This method is hacky and not widely supported, so we wanted to use a better way.

In this PR, I propose a solution based on background images, swapping the previous image tags with fixed-size spans that use the same classes to set their background image. This approach is widely used and battle tested, although **I would recommend moving into a different solution in the future: inserting the SVG icons directly.**

I am also adding some a11y improvements, getting from 63 to 42 issues in our search pages (measured with the [Axe Chrome Extension](https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd)). Still a long way to go, but most of the next issues are pretty similar.

### Tests
Updates tests 

Rendering on Firefox:
Amundsen_-_Data_Discovery_Portal

### CheckList

### DIFF
--- a/amundsen_application/static/css/_icons.scss
+++ b/amundsen_application/static/css/_icons.scss
@@ -1,23 +1,92 @@
-@import 'variables';
+@import "variables";
+
+$icon-size: 24px;
+$icon-small-size: 16px;
+
+$default-db-logo-path: "/static/images/icons/Database.svg";
+$default-dashboard-logo-path: "/static/images/icons/dashboard.svg";
+
+// DBs
+$hive-logo-path: "/static/images/icons/logo-hive.svg";
+$bigquery-logo-path: "/static/images/icons/logo-bigquery.svg";
+$druid-logo-path: "/static/images/icons/logo-druid.svg";
+$hive-logo-path: "/static/images/icons/logo-hive.svg";
+$presto-logo-path: "/static/images/icons/logo-presto.svg";
+$postgres-logo-path: "/static/images/icons/logo-postgres.svg";
+$redshift-logo-path: "/static/images/icons/logo-redshift.svg";
+
+// Dashboards
+$mode-logo-path: "/static/images/icons/logo-mode.svg";
+
+
+.icon.resource-icon {
+  background-color: $icon-bg;
+  border: none;
+  height: $icon-size;
+  // margin: -3px 4px -3px 0px;
+  margin: auto 16px auto 0px;
+  min-width: $icon-size;
+  width: $icon-size;
+
+  // Default DB icon
+  &.icon-database {
+    background: transparent url($default-db-logo-path) center center / contain no-repeat;
+  }
+
+  // Default Dashboard icon
+  &.icon-dashboard {
+    background: transparent url($default-dashboard-logo-path) center center / contain no-repeat;
+  }
+
+  // Database Icons
+  &.icon-bigquery {
+    background: transparent url($bigquery-logo-path) center center / contain no-repeat;
+  }
+
+  &.icon-druid {
+    background: transparent url($druid-logo-path) center center / contain no-repeat;
+  }
+
+  &.icon-hive {
+    background: transparent url($hive-logo-path) center center / contain no-repeat;
+  }
+
+  &.icon-presto {
+    background: transparent url($presto-logo-path) center center / contain no-repeat;
+  }
+
+  &.icon-postgres {
+    background: transparent url($postgres-logo-path) center center / contain no-repeat;
+  }
+
+  &.icon-redshift {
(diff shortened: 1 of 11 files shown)


amundsen_application/static/js/components/common/ResourceListItem/DashboardListItem/index.tsx
```diff
@@ -29,13 +29,17 @@ class DashboardListItem extends React.Component<DashboardListItemProps, {}> {
     return `/dashboard?uri=${dashboard.uri}&index=${logging.index}&source=${logging.source}`;
   };
 
+  generateResourceIconClass = (dashboardId: string, dashboardType: ResourceType): string => {
+    return `icon resource-icon ${getSourceIconClass(dashboardId, dashboardType)}`;
+  };
+
```
Golodhros, line 35: Updated for consistency with Table page.

feng-tao:
nice! the icon not render for me previously if I am on firefox before :(

Golodhros:
Meow @ttannis @danwom @feng-tao

Golodhros:
Meow @ttannis @danwom

amundsen_application/static/css/_icons.scss
```diff
@@ -29,163 +73,118 @@ img.icon {
   }
 
   &.icon-alert {
-    -webkit-mask-image: url('/static/images/icons/Alert-Triangle.svg');
-    mask-image: url('/static/images/icons/Alert-Triangle.svg');
+    -webkit-mask-image: url("/static/images/icons/Alert-Triangle.svg");
```
ttannis, line 76: Note that we never made a strict lint choice for single vs double quotation marks, though I slightly favor double quotes if required to make a choice right at this moment. If you want to intentionally go this route, it would be great to have one PR that makes the changes everywhere to make diffs for future work easier to read through.

amundsen_application/static/js/components/DashboardPage/index.tsx
```diff
@@ -171,7 +171,7 @@ export class DashboardPage extends React.Component<DashboardPageProps, Dashboard
                className="btn btn-default btn-lg">Open Dashboard</a>
           </div>
         </header>
-        <main className="column-layout-1">
+        <article className="column-layout-1">
```
ttannis, line 174: So does `axe` highlight issues like this, or did you just happen to know what these should be ideally be? I'd be curious to see how you use the tool to work through updating a page.

amundsen_application/static/css/_icons.scss
Golodhros: Makes sense!
I think this was my prettier changing things, but I thought it was following the project settings, so I was wrong there.

I am open to undo the change or work on setting the project configuration to enforce these with prettier or ESLint!

Golodhros:
> To make sure I understand the full context behind the suggestion of embedding the svg's directly -- I'm familiar of performance benefits and interaction/animation support using SVG (though the latter is not a need for the tool as it exists now). Anything else beneficial to call out, so that I can retain that for the future?

Yeah, performance (saving some requests) and quality of image, as they will always look sharp, no matter the size or pixel density of the display.

amundsen_application/static/js/components/DashboardPage/index.tsx
Golodhros: Axe told me that there were a duplicated `<main>` tag after I added the global one on the top level component. 

Using an article here was my call on what would work better. You could [check here](https://www.smashingmagazine.com/2020/01/html5-article-section/) for more about the rationale.

With Axe, I just open its panel on the dev tools and see something like this:
Fixes_DataSource_or_Dashboard_Icons_not_rendering_on_Firefox_by_Golodhros_·_Pull_Request__443_·_lyft_amundsenfrontendlibrary

Then I circle through the issues (you can toggle the "Highlight" link on the top right to see which element its referring to) and check if I can fix them easily or would need more deep changes.

amundsen_application/static/css/_icons.scss
Golodhros: I will merge for now, but make a note about syncing about this later, thanks!

jornh:
Another Firefox 🔥 user here!

Sounds like this PR fixes https://github.com/lyft/amundsenfrontendlibrary/pull/443

PR #445, opened 2020-05-14
Split ResourceList Components

ttannis:
### Summary of Changes
Our UI currently has 3 use cases that need to be supported for showing a list of resources. We started with only 1, and as time went on we extended our initial `ResourceList` component to cover all of these cases with flexible properties and different combinations of boolean checks. `ResourceList` was on the border of becoming unmanageable, and this PR splits up the components while also fixing a bug with the `MyBookmarks` component.

Suggested method for review may be to ignore large diffs but review the new separated components -- `ResourceList`, `PaginatedResourceList`, and `PaginatedApiResourceList` -- for correctness within the context of their use cases. 

 #### `ResourceList`
**Parent Components:**  `DashboardPage`, `ProfilePage`
**Use Case:** Show a straight list `ResourceListItems`, where the parent component has access to the full list of items. The component initially renders up to `itemsPerPage` items, and has a toggle at the bottom of the list to "View all" or "View less". 

 #### `PaginatedResourceList`
**Parent Components:**  `MyBookmarks`,  `PopularTables`
**Use Case:** Show a list of paginated `ResourceListItems`, where the parent component has access to the full list of items, and pagination updates the `activePage` state and renders a different slice of the entire list according to the current `activePage`.
**Bug Fix:**  Leverage `componentDidUpdate` to reset the `activePage` if the component is updated with a smaller list of items. Reproduction steps:
1.  Have 5 bookmarks
2. Navigate to the second page of your bookmarks
3. Unbookmark your 5th bookmark. Before this fix the component would render nothing becaue the sliced array would be empty. With the fix the `activePage` is updated accordingly and the user will see their first page of bookmarks.

 #### `PaginatedApiResourceList`
**Parent Components:**  `SearchPage`
**Use Case:** Show a list of paginated `ResourceListItems`, where the parent component only has access to one page of items at a time. Pagination fires an API call for the new data and the component updates itself accordingly

### Tests

Updated tests according to changes.

### Documentation

Bug fix will be called out in notes for next release.

### CheckList


ttannis:
@Golodhros for review.

cc @dikshathakur3119 for learning. Feel free to ask questions on our PRs moving forward.

amundsen_application/static/js/components/ProfilePage/index.spec.tsx
```diff
@@ -9,7 +9,7 @@ import Breadcrumb from 'components/common/Breadcrumb';
 import Flag from 'components/common/Flag';
```
Golodhros, line 9: I really prefer this way of keeping the tests beside the tested components.
I assume we plan to move the rest of tests to this place, right?

amundsen_application/static/js/components/common/ResourceList/PaginatedApiResourceList/index.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  describe('onPagination', () => {
+    it('calls the onPagination prop', () => {
+      const setupResult = setup();
+      const wrapper = setupResult.wrapper;
+      const props = setupResult.props;
+      const onPaginationSpy = jest.spyOn(props, 'onPagination');
+      wrapper.instance().onPagination(3);
+      expect(onPaginationSpy).toHaveBeenCalledWith(2);
+    });
```
Golodhros, line 41: It is usually recommended to call `wrapper.unmount();` after every jest test case to keep the scope clean and avoid weird and hard to debug bugs.

Some teams prefer to do it on every test case, some prefer to make the `wrapper` variable a shared one and do it on a `beforeEach` clause.

amundsen_application/static/js/components/common/ResourceList/PaginatedApiResourceList/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+
+class PaginatedApiResourceList extends React.Component<PaginatedApiResourceListProps, {}> {
+  public static defaultProps: Partial<PaginatedApiResourceListProps> = {
+    emptyText: Constants.DEFAULT_EMPTY_TEXT,
+  };
+
+  constructor(props) {
+    super(props);
+  }
```
Golodhros, line 26: Probably not necessary, maybe something left from of a previous version?

amundsen_application/static/js/components/common/ResourceList/PaginatedApiResourceList/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+                activePage={ activePage + 1 }
+                itemsCountPerPage={ itemsPerPage }
+                totalItemsCount={ totalItemsCount }
+                pageRangeDisplayed={ Constants.PAGINATION_PAGE_RANGE }
+                onChange={ this.onPagination }
+              />
+            }
+          </>
+        }
+        <div className="resource-list-footer" />
```
Golodhros, line 68: Do we need this div?

amundsen_application/static/js/components/common/ResourceList/PaginatedResourceList/index.spec.tsx
```diff
@@ -0,0 +1,119 @@
...
+      const wrapper = setup({ itemsPerPage: 2 }).wrapper;
+      wrapper.setState({ activePage: 2 });
+      const componentDidUpdateSpy = jest.spyOn(wrapper.instance(), 'componentDidUpdate');
+      wrapper.setProps({
+        allItems: [
+          { type: ResourceType.table },
+          { type: ResourceType.table },
+        ]
+      });
+      expect(componentDidUpdateSpy).toHaveBeenCalled();
```
Golodhros, line 43: I am not sure of the utility of this expectation. Wouldn't this be called anyways?

amundsen_application/static/js/components/common/ResourceList/PaginatedResourceList/index.spec.tsx
```diff
@@ -0,0 +1,119 @@
...
+          { type: ResourceType.table },
+          { type: ResourceType.table },
+          { type: ResourceType.table },
+          { type: ResourceType.table },
+          { type: ResourceType.table },
+        ]
+      })
+      expect(wrapper.state().activePage).toEqual(2);
+    })
+  });
```
Golodhros, line 76: Given that the result of this logic is different values to the activePage state value, could we check this on the rendered results instead?

amundsen_application/static/js/components/common/ResourceList/PaginatedResourceList/index.spec.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  describe('onPagination', () => {
+    it('calls setState to update the activePage', () => {
+      setStateSpy.mockClear();
+      const wrapper = setup().wrapper;
+      wrapper.instance().onPagination(3);
+      expect(setStateSpy).toHaveBeenCalledWith({ activePage: 2 });
+    });
+  });
```
Golodhros, line 85: IMHO this test goes too much into the details.

We could do something similar by rendering the component, calling onPagination with the value 3, and then seeing how the rendered component changes.

amundsen_application/static/js/components/ProfilePage/index.spec.tsx
ttannis: Yes.

amundsen_application/static/js/components/common/ResourceList/PaginatedApiResourceList/index.spec.tsx
ttannis: This is new to me. I'd see the benefit if we were using `mount` for the tests,  but is it something that we have to do event with `shallow` rendering?

amundsen_application/static/js/components/common/ResourceList/PaginatedApiResourceList/index.tsx
ttannis: You're right not necessary. This isn't unique to this file though, "copy-paste" has led to at least a few more `constructor(props) { super(props) }` in a few other files I'll make a note to remove them on a future code-cleanup-only PR.

amundsen_application/static/js/components/common/ResourceList/PaginatedApiResourceList/index.tsx
ttannis: Nope. I  will also have to update some styles if I remove it to maintain the space we need at the bottom of the lists.

amundsen_application/static/js/components/common/ResourceList/PaginatedResourceList/index.spec.tsx
ttannis: That's fair, the valuable assertion to me here is `expect(wrapper.state().activePage).toEqual(0);`

amundsen_application/static/js/components/common/ResourceList/PaginatedResourceList/index.spec.tsx
ttannis: Could you elaborate I'm not sure I follow?

amundsen_application/static/js/components/common/ResourceList/PaginatedApiResourceList/index.spec.tsx
Golodhros: Good call!
It can definitely be done, as [in the docs they use it as an example](https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/unmount.html), but a quick search didn't give me any insights.

Not a big deal though!

amundsen_application/static/js/components/common/ResourceList/PaginatedApiResourceList/index.tsx
Golodhros: +1

amundsen_application/static/js/components/common/ResourceList/PaginatedResourceList/index.spec.tsx
ttannis: Would the suggestion be correctly summarized as "unit test `onPagination` by verifying its effects on the rendered result"?

If so, one of the things Daniel and I chose to move toward was the concept of writing tests that would not break if and when we made changes in the logic of other methods. This exact case is not the worst kind of case since te rendering for this component is basically just affected by a function of `activePage` and `itemsPerPage`, and `itemsPerPage` is always a constant. However for more complicated components I would see us more likely to be on the hook for modifying assertions in these methods if we get new designs or otherwise tweak the UI as we iterate through features. I'm open to talking more about unit test approach pros and cons though

Golodhros: Yes, you word it better than I did. The public effects of clicking on "onPagination" are seen on the rendering of the component (and by extension, of the Pagination component), while calling 'setState' can be considered an implementation detail.

You are right that in this case, checking those public effects would mean that we will search on the rendered result of the Pagination component, making this more like a small "integration test". Maybe a middle ground would be to check the props of the Pagination component after rendering it, something like:
`wrapper.find(Pagination).props().activePage`. I think this is less problematic than checking the state or spying on setState.

amundsen_application/static/js/components/common/ResourceList/PaginatedResourceList/index.spec.tsx
Golodhros: Yes, let me expand on it here.

In a similar way as in my comment below, I have experienced the issues that come from checking for implementation details on tests. 

In this case, we are rendering a component with a giving state, simulating a change of props and making sure that the new state is what we expect. This makes sense, and it is not a wrong test in any case. Actually, following the "London School of testing", this is exactly what they propose. They prefer deep tests that are strict but not flexible.

However, the way I have been using and recommend to use is a bit less detailed, but more flexible. This way would be to only check for the "public effects" of the actions. In this case, that would be what gets rendered in the DOM. That would mean that we would set up the component the same way, we would act solely on the 'onPagination' method and will check the results on the DOM. (In this case, as commented below, it would mean we depend on the Pagination component to render what we need). The benefits of this way ("The Chicago School" of testing) is that our tests would be more flexible and break less, at the cost of less strict checking to the implementation details. 

A good example we can think of now is the refactor to hooks. The test we are adding now would break with hooks, but if we check only the rendered result, they would still pass.

That said, I don't actually recommend to remove this tests. When I see this tests on a code base, I prefer them than no tests, and most of the times it is not worth to rewrite them. So what I suggest is to leave them here, but the moment that they break, we should remove them (the same for the next test).

Sorry for the wall of text!!

ttannis: This makes sense given the example of hooks, which as I think I mentioned I' m unfamiliar but have a goal to learn more about. Let's continue to talk about this offline. None of our methods are set in stone, we essentially learn as we go discussing different perspectives on the best way to approach things. And then if we do shift practices, for unit tests at least we just make those changes moving forward vs. trying to retroactively fix update hundreds of tests. 

So in short, let's keep discussing and feel free to write tests from your perspective.

amundsen_application/static/js/components/common/ResourceList/PaginatedResourceList/index.spec.tsx
ttannis: As mentioned above, let's make a goal to do some knowledge share on these topics!

amundsen_application/static/js/components/common/ResourceList/PaginatedResourceList/index.spec.tsx
Golodhros: Totally, makes sense! 
Thanks!

PR #453, opened 2020-05-16
Changes the favicon depending on the current environment

Golodhros:
### Summary of Changes
Adds more icons and settings on the metadata of our application HTML
Switches between favicons on dev and production
*Ready to do it on staging once I figure out the way

### CheckList


Golodhros:
🔨

amundsen_application/static/images/favicons/dev/safari-pinned-tab.svg
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" standalone="no"?>
```
ttannis, line 1: What's the use case for this file?

amundsen_application/static/images/favicons/dev/browserconfig.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<browserconfig>
+    <msapplication>
+        <tile>
+            <square150x150logo src="/static/images/favicons/dev/mstile-150x150.png"/>
+            <TileColor>#2d89ef</TileColor>
```
ttannis, line 6: What does this file end up achieving? Also why `#2d89ef`? This color isn't part of any of our themes.

amundsen_application/static/images/favicons/dev/site.webmanifest
```diff
@@ -0,0 +1,19 @@
+{
```
ttannis, line 1: Same question here to understand what this file achieves.

amundsen_application/static/templates/fragments/icons-dev.html
```diff
@@ -0,0 +1,10 @@
+<link rel="apple-touch-icon" sizes="180x180" href="/static/images/favicons/dev/apple-touch-icon.png">
```
ttannis, line 1: What do these particular cases cover? I'm not clear on why we need to generate this many images and when each one would get used.

Also, similar comment about anything that defines a color. What's the effect and if anything shouldn't these be Lyft and/or Amundsen theme colors?

amundsen_application/static/images/favicons/dev/browserconfig.xml
Golodhros: This one is for tiles on windows machines.
Yeah, it was par of a given set they give on the icon generator. It is not a big deal, I picked one that didn't look bad.

amundsen_application/static/images/favicons/dev/safari-pinned-tab.svg
Golodhros: I think this one is for the add to homescreen on safari mobile.

amundsen_application/static/images/favicons/dev/site.webmanifest
Golodhros: This one is the manifest file on the line of Progressive Web Apps.
It sets up the homescreen icon and start screen for when users add amundsen to their homescreen in their phones (also in windows desktops)

amundsen_application/static/templates/fragments/icons-dev.html
Golodhros: Yeah, the world of favicons and app icons and all that is really confusing, that's why it is better to just use a generator to create these and forget about the rest.

This is the generator I usully use: https://realfavicongenerator.net/. We just need to give it one icon and it creates the rest for you keeping the current best practices. If you are curious about the use of each of this icons, you can run the generator and learn more about them.

amundsen_application/static/images/favicons/dev/safari-pinned-tab.svg
ttannis: Officially we don't support mobile for this tool -- not just from the Eng perspective but our designs also don't take mobile into account. I'd suggest removing any of the mobile specific values because we can say at this point we would never use them, but my understanding is that they all came from a generator? (in which case I'd just let the generator do it's thing and not mess with its output)

amundsen_application/static/images/favicons/dev/site.webmanifest
ttannis: I see, also from the generator?

amundsen_application/static/images/favicons/dev/safari-pinned-tab.svg
Golodhros: Yeah, that is my take as well. We probably don't need half of this, but as it is just copy/paste from the generator output, it takes more work to prune it than to leave it here.

amundsen_application/static/images/favicons/dev/site.webmanifest
Golodhros: Yep!
This one is more useful though, it works also for windows desktop machines.

amundsen_application/static/templates/fragments/icons-dev.html
ttannis: I see I didn't pay attention to the `<!-- Generated with https://realfavicongenerator.net/ -->` comment. So all these files came from the generator, and you just made updates to the template and added environment variable logic?

amundsen_application/static/images/favicons/dev/browserconfig.xml
ttannis: Are we able to manually update these to match our theme for that bit of extra polish, or does the generator not allow it? I won't suggest any alternative colors right now without knowing what this looks like, but curious for the future.

amundsen_application/static/templates/fragments/icons-dev.html
Golodhros: Yeah, basically!

Oh, I just added the comment :)

amundsen_application/static/images/favicons/dev/browserconfig.xml
Golodhros: This generator didn't let me add the colors in this one, but gave me a bunch of options (10 or so).
We can always override it here!

PR #455, opened 2020-05-19
[AMD-1385] Changed 'Issue' border to match design

dikshathakur3119:
### Summary of Changes
Changed Border and font of 'Open Issues' to match design

### Tests
No test added or modified

### Documentation
NA

### DIFF
--- a/amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
+++ b/amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
@@ -3,13 +3,14 @@
 .table-issues {
 
   .issue-banner {
-    border-radius: 4px;
-    box-shadow: 0px 0px 1px 0px rgba(0, 0, 0, 0.12), 0px 2px 3px 0px rgba(0, 0, 0, 0.16);
     height: 40px;
     margin: 8px 0;
-    padding: 8px 16px;
+    padding: 8px;
     display: flex;
     flex-direction: row;
+    border-top: 2px solid $gray10;
+    border-bottom: 2px solid $gray10;
+    color: $gray60;
   }
   .table-issue-link {
     margin-right: $spacer-1;


dikshathakur3119:
👀 @ttannis @feng-tao @Golodhros

dikshathakur3119:
💨

amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
```diff
@@ -3,13 +3,14 @@
...
   .issue-banner {
-    border-radius: 4px;
-    box-shadow: 0px 0px 1px 0px rgba(0, 0, 0, 0.12), 0px 2px 3px 0px rgba(0, 0, 0, 0.16);
     height: 40px;
     margin: 8px 0;
-    padding: 8px 16px;
+    padding: 8px;
     display: flex;
     flex-direction: row;
+    border-top: 2px solid $gray10;
```
ttannis, line 11: @dikshathakur3119 If you could add as a follow up, I believe Figma lists these as `1px`.

amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
```diff
@@ -3,13 +3,14 @@
...
+    border-bottom: 2px solid $gray10;
+    color: $gray60;
```
ttannis, line 13: @dikshathakur3119 let's use the `$text-secondary` value here, which should have the same color. Something we do is create specific variables for commonly used design concepts, like text colors for example. That way, if we ever want to change the look of the application, or if our community members want to easily override the default theme, they just have to change the value `$text-secondary` in [one place ](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/css/_variables-default.scss#L26) versus looking for all the individual places where it needs to be changed.

PR #458, opened 2020-05-19
Optimizes JavaScript Bundle

Golodhros:
### Summary of Changes
- Optimizes load of lodash
- Adds moment-locales-webpack-plugin to prune included moment locales except 'en'
- Adds BundleAnalyzerPlugin (commented) to analyze bundle sizes
- Updates clean-webpack-plugin

Before:
static__19_May_2020_at_14_33_

After:
static__19_May_2020_at_14_24_

### Tests
Ran tests, checked sizes with the bundle analyzer

### CheckList

### DIFF
--- a/amundsen_application/static/package.json
+++ b/amundsen_application/static/package.json
@@ -53,7 +53,7 @@
     "@types/react-router": "^4.4.5",
     "@types/react-tagsinput": "^3.19.7",
     "@types/webpack": "^4.41.12",
-    "babel-jest": "^23.4.2",
+    "babel-jest": "^24.9.0",
     "babel-loader": "^8.1.0",
     "bootstrap-sass": "^3.4.1",
     "clean-webpack-plugin": "3.0.0",
--- a/amundsen_application/static/package.json
+++ b/amundsen_application/static/package.json
@@ -27,10 +27,24 @@
   "devDependencies": {
     "@babel/cli": "^7.8.4",
     "@babel/core": "^7.9.6",
-    "@babel/polyfill": "^7.8.7",
-    "@babel/preset-env": "^7.9.6",
+    "@babel/plugin-proposal-class-properties": "^7.0.0",
+    "@babel/plugin-proposal-decorators": "^7.0.0",
+    "@babel/plugin-proposal-do-expressions": "^7.0.0",
+    "@babel/plugin-proposal-export-default-from": "^7.0.0",
+    "@babel/plugin-proposal-export-namespace-from": "^7.0.0",
+    "@babel/plugin-proposal-function-bind": "^7.0.0",
+    "@babel/plugin-proposal-function-sent": "^7.0.0",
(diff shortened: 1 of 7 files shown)


amundsen_application/static/package.json
```diff
@@ -39,11 +53,10 @@
     "@types/react-router": "^4.4.5",
     "@types/react-tagsinput": "^3.19.7",
     "@types/webpack": "^4.41.12",
-    "babel-jest": "^24.9.0",
+    "babel-jest": "^23.4.2",
```
ttannis, line 56: Surprising to see a downgrade. Was this done automatically when making other changes or was there some issue with the previous version where you chose to intentionally downgrade?

Could you also provide context on the babel changes, did it all just come from the move to `core-js`? We've sadly only touched this stuff on as as-needed basis since the first working configuration we had and want to make sure I understand for future maintenance.

amundsen_application/static/webpack.common.ts
```diff
@@ -70,8 +76,11 @@ const config: webpack.Configuration = {
       ],
     },
     plugins: [
+      new CleanWebpackPlugin(),
```
ttannis, line 79: My understanding is that `CleanWebpackPlugin` removes things in `dist/`, but I think some part of our build script in the private repo leverages the `dist/` folder. I'm not recalling enough context at the moment to know if this is an issue, but could you take a look at the `partnerized-static-build.sh` to see what I'm referencing. However if all of the commands and and flags still work in that script with these changes then I assume we're in the clear.

amundsen_application/static/package.json
Golodhros: Oh, good catch!
Yes, this was the babel upgrade script, I will retrieve the previous one, it will probably work the same.

Yes, happy to extend on the babel updates. On version 7.4 of babel they deprecated @babel/polyfill, that it was using core-js internally. Now, instead of using the polyfills, we are going to use that core-js library directly. 

One of the main changes is the move from using the yearly presets and "stage-0" setup into the "preset-env", which uses a list of browsers instead. This list is on the package.json file. This change means that we need to bring the required plugins one by one. 

That said, if we are strict about only supporting Chrome, we could probably get rid of most of this plugins and save some space. Food for thought!

amundsen_application/static/webpack.common.ts
Golodhros: Yes, this plugins cleans up the /dist folder before creating a new version. 

Hmmm... I see in the script that most of the times we won't copy over the /dist folder, and we will build it again. It is working fine on dev for now, I have tried different runs of the script as well.

I will keep an eye on it just in case, and I will test it on staging before merging.

Thanks for the heads up!

amundsen_application/static/package.json
ttannis: Alright sounds good. We're not strict about only supporting Chrome. It's just the only thing currently promise our users because of the nature of not being able to guarantee resources to fix cross-functionality issues.

PR #460, opened 2020-05-20
Adds option to open Dashboard preview image into a modal

Golodhros:
### Summary of Changes
- Allows users to open the dashboard preview image into a modal

### Tests
Added tests 

### CheckList


amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
```diff
@@ -1,10 +1,11 @@
...
+import Linkify from "react-linkify";
+import { Modal } from "react-bootstrap";
 
-import ShimmeringDashboardLoader from '../ShimmeringDashboardLoader';
+import ShimmeringDashboardLoader from "../ShimmeringDashboardLoader";
 
-import * as Constants from './constants';
-import './styles.scss';
+import * as Constants from "./constants";
+import "./styles.scss";
```
Golodhros, line 8: Sorry, prettier changes.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
```diff
@@ -14,17 +15,43 @@ export interface ImagePreviewProps {
...
   hasError: boolean;
+  isModalVisible: boolean;
 }
 
-export class ImagePreview extends React.Component<ImagePreviewProps, ImagePreviewState> {
-  constructor(props) {
-    super(props);
+type PreviewModalProps = {
+  imageSrc: string,
+  onClose(): void,
```
ttannis, line 23: So this is just another way of doing: `onClose: () => void`?

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
```diff
@@ -14,17 +15,43 @@ export interface ImagePreviewProps {
...
+  const [show, setShow] = React.useState(true);
+  const handleClose = () => {
+    setShow(false);
+    onClose();
+  };
+
+  return (
+    <Modal show={show} onHide={handleClose} scrollable="true" className="dashboard-preview-modal">
+      <Modal.Header closeButton={true}>
+        <Modal.Title className="text-center">Dashboard Preview</Modal.Title>
```
ttannis, line 36: nit: Can we put `Dashboard Preview` in the constants file? A while ago we started trying to make sure we pull out all hardcoded English text as we go along, for the purpose of being able to easily find all cases if and when we introduce support for translations. While you're in that file, mind removing `LOADING_TEXT` which is no longer used?

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
```diff
@@ -34,36 +61,58 @@ export class ImagePreview extends React.Component<ImagePreviewProps, ImagePrevie
...
+    const { isLoading, hasError, isModalVisible } = this.state;
+    const imageSrc = `${Constants.PREVIEW_BASE}/${uri}/${Constants.PREVIEW_END}`;
+
     return (
-      <div className='image-preview'>
-        {
-          this.state.isLoading &&
+      <div className="image-preview">
+        {isLoading && (
           <div className="text-placeholder">
```
ttannis, line 80: Slipped my mind in your shimmer loader PR, but we no longer need the `<div>` to apply the `text-placeholder` style on this, since the loading state no longer has text.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.spec.tsx
```diff
@@ -71,29 +72,80 @@ describe('ImagePreview', () => {
...
           const setupResult = setup();
           props = setupResult.props
           wrapper = setupResult.wrapper;
-          wrapper.instance().setState({ isLoading: false, hasError:false });
+          wrapper.instance().setState({ isLoading: false, hasError: false });
         });
+
         it('renders visible img with correct props', () => {
           const elementProps = wrapper.find('img').props();
-          expect(elementProps.style).toEqual({ visibility: 'visible' });
```
ttannis, line 85: Is this redundant in some way? I find it valuable to verify any logic that affects what shows up on the page.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.spec.tsx
```diff
@@ -71,29 +72,80 @@ describe('ImagePreview', () => {
...
-    it('renders link if hasError', () => {
-      const { props, wrapper } = setup();
-      wrapper.instance().setState({ hasError: true });
-      expect(wrapper.find(Linkify).exists()).toBeTruthy();
+    describe('when there is an error', () => {
+      it('renders a link', () => {
+        const { props, wrapper } = setup();
+
+        wrapper.instance().setState({ hasError: true });
+        wrapper.update();
```
ttannis, line 108: Do we need this update, or is it a best practice? Does `setState` alone risk not functioning as expected?

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.spec.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
+import { Modal } from 'react-bootstrap';
 
-import { shallow } from 'enzyme';
+import { mount } from 'enzyme';
```
ttannis, line 4: Oh, why `mount`? The initial approach we took for this repo was to use `shallow` rendering unless we had a case that justified using `mount`. We were concerned about how expensive it would be to call `mount`, and how long it could take to run tests as the repository grew. 

Is it that `mount` allows for your preferred style of testing in the `lifecycle` section below? I'm not opposed to using `mount` where it has a benefit, but want to make sure I understand the benefit.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.spec.tsx
Golodhros: Yes, is a best practice to ensure you get a render before running the expectation. As setState is asynchronous, it could happen. Probably not in this case, but just in case.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.spec.tsx
Golodhros: It started breaking when I switched to mount. 
I didn't spend too much time trying to debug it though, I could do it now if you want.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.spec.tsx
Golodhros: This was also breaking when using "mount", as it wasn't finding the Linkify component on the rendered tree.

Somehow, shallow gives a false positive there, like it was rendered, but it was only "supposed to render".

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.spec.tsx
Golodhros: Yes, mount was necessary in order for the lyfecycle section to work, as it is rendering a child component (PreviewModal) and simulating events on it.

This was one of the things that surprised me at the beginning, as in my previous teams we never used shallow. It has some bad press, and even if it has improved over time (time ago you couldn't call some methods), it gives a false sense of security.

You are right about mount being slower, but it is a matter of milliseconds. Also, having to switch between one and the other depending on if you have nested components is a bit painful.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.spec.tsx
Golodhros: Ah, I only needed the `wrapper.update()`, I will bring it back!

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
Golodhros: Oh, it seems!
I was googling fast for a way of making the warning go. I can put your proposed one there if it is more starndard.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
Golodhros: Yeah! 
Totally agree, I will do the same on the next strings I found.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
Golodhros: Makes sense, removing!

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.spec.tsx
Golodhros: Here is a more opinionated response to the mount vs shallow: https://kentcdodds.com/blog/why-i-never-use-shallow-rendering

He has his own way, so take it with a grain of salt :)

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
ttannis: I can't claim its more standard because I don't know if that's actually true. Consistency would help for people who are newer the space and might not immediately know this is just another way of doing the same thing, but don't mind this as is.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.spec.tsx
ttannis: Didn't expect that. Good to know.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.spec.tsx
ttannis: I've read that article previously, along with others that basically take the opposite standpoint. I'll invest more research here at some point to be able to provide clearer directions for newcomers to the repository. Essentially even with all the pros and cons available in various posts and documents there's some effort on my part to revisit how much the application and components have changed since the initial decision, in order to try to draw some clearer connections between other's opinions and the needs of the app.

Internally our FE working group has actually been working on some more official guidance for unit testing -- that doc is almost finalized though and I've been waiting on that or guidance on improving our testing practices as well. I'll send you the link. (fyi: They do happen to lean towards `react-testing-library` and no shallow rendering.)

Golodhros: Totally, it makes sense to have a clear idea of where to go before jumping into refactoring tests!

I will check that doc, I received the email this week and added it to my reading list, I will prioritize it now. 

To be honest, I am not so much into the React Testing Library, I think enzyme is fine for most of the cases if used properly. That said, I don't know about hooks, maybe that changes everything.

Thanks!

PR #461, opened 2020-05-20
style: Design Fixes to resource detail

dikshathakur3119:
### Summary of Changes
 - Change font color to primary color


dikshathakur3119:
👀 @ttannis @Golodhros @feng-tao

amundsen_application/static/css/_layouts.scss
```diff
@@ -102,6 +102,10 @@ $resource-header-height: 84px;
 
       .editable-text {
         font-size: 16px;
+        
+        .markdown-wrapper{
+          font-size: 16px;
+        }
```
Golodhros, line 108: Out of curiosity, is this needed?
Usually the statement above will do, but it could be that there are some style overrides.

amundsen_application/static/js/components/DashboardPage/index.tsx
```diff
@@ -230,15 +230,15 @@ export class DashboardPage extends React.Component<DashboardPageProps, Dashboard
...
                 <div className="section-title title-3">Last Updated</div>
-                <div>
+                <div className="body-2 text-primary">
                   { formatDateTimeShort({ epochTimestamp: dashboard.updated_timestamp }) }
                 </div>
                 <div className="section-title title-3">Recent View Count</div>
-                <div>
+                <div className="body-2 text-primary">
                   { dashboard.recent_view_count }
                 </div>
```
Golodhros, line 243: Could you try to use semantic markup here? Maybe it works out of the box.

In this case, we would be something like:
```
<h3 className="section-title title-3">Created</h3>
<p className="body-2 text-primary">{ formatDateTimeShort....}</p>
...
```

Golodhros: Here for more about the rationale: https://medium.com/adalab/the-importance-of-semantic-html-78e74fb75ff0

dikshathakur3119: Thanks Marcos, will try semantic markup.

amundsen_application/static/css/_layouts.scss
dikshathakur3119: Yes it has been overridden, I will change font there and see if that is the correct way to do it.

ttannis: I'd say this option was a better fix. `EditableText` is used in our column descriptions as well, which I believe are to remain at `14px`. We can keep the default at `14px` for that component and instead only apply `16px` where needed -- which this version looks like it accomplishes.

ttannis: @dikshathakur3119 FYI see comment above.

feng-tao:
the available words are https://github.com/lyft/amundsenfrontendlibrary/blob/master/.github/prlint.json#L4

feng-tao:
prlint is fixed: so basically it needs to have : after the topic.

ttannis:
@dikshathakur3119 is this ready for another review?

amundsen_application/static/css/_layouts.scss
ttannis: Thanks for the update.

amundsen_application/static/css/_layouts.scss
```diff
@@ -102,6 +102,10 @@ $resource-header-height: 84px;
 
       .editable-text {
         font-size: 16px;
+
+        .markdown-wrapper{
```
ttannis, line 106:
```suggestion
        .markdown-wrapper {
```

amundsen_application/static/css/_typography-default.scss
```diff
@@ -149,3 +149,7 @@ body {
 .text-secondary {
   color: $text-secondary;
 }
+
+.text-primary{
```
ttannis, line 153:
```suggestion
.text-primary {
```

PR #465, opened 2020-05-26
fix: Fix SearchPage min-height

ttannis:
### Summary of Changes

At some point, our styles for the height of the `SearchPage` stopped working as intended. The page was no longer taking up the full height unless there was content to fill it -- this could be observed on an empty `SearchPage` where `Dashboard` or `People` is selected as the resource.

I was unable to trace when this actually broke, but updated the css accordingly to fix the issue.

_Before_
image

_After_
image

### Tests

N/A for this kind of fix

### Documentation

N/A for this kind of fix

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/SearchPage/styles.scss
+++ b/amundsen_application/static/js/components/SearchPage/styles.scss
@@ -2,7 +2,7 @@
 
 .search-page {
   display: flex;
-  min-height: calc(100% - #{$nav-bar-height + $footer-height});
+  min-height: calc(100vh - #{$nav-bar-height + $footer-height});
 
   .list-group {
      margin-top: 0px;


amundsen_application/static/js/components/SearchPage/styles.scss
```diff
@@ -2,7 +2,7 @@
 
 .search-page {
   display: flex;
-  min-height: calc(100% - #{$nav-bar-height + $footer-height});
+  min-height: calc(100vh - #{$nav-bar-height + $footer-height});
```
Golodhros, line 5: Yeah, height: 100% is tricky as it needs all its containers up to the <html> tag to have height 100% as well!

PR #468, opened 2020-05-28
feat: Adding schema description to search result items

dikshathakur3119:
### Summary of Changes

- Adding new component for schema name and attached description.
- On hovering over schema name, a tooltip will appear with schema description.

### Tests

NA

### Documentation

NA

### CheckList


dikshathakur3119:
@ttannis @Golodhros , this is work in progress PR, let me know if I am on right track. Also, Travis CI - Branch is failing, I am not sure why?

amundsen_application/static/js/components/common/ResourceListItem/SchemaInfo/index.tsx
```diff
@@ -23,7 +25,7 @@ const SchemaInfo: React.SFC<SchemaInfoProps>  = ({ schema, placement, desc }) =>
      placement={ placement }
      overlay={popoverHoverFocus}
      >
-      <span>{ schema }</span>
+      <span className="underline">{ schema }</span>
```
Golodhros, line 28: You might not need the span tag anymore, I would check it out.

amundsen_application/static/js/components/common/ResourceListItem/SchemaInfo/styles.scss
```diff
@@ -0,0 +1,5 @@
+@import 'variables';
+
+.underline{
+  border-bottom: solid #9191A8;
```
Golodhros, line 4: is there a variable for that color?

dikshathakur3119: I have bot checked yet, I was trying to put here working code. Do you think I should put this style in _layout.scss file as underline may be used at other places for standardization? or should I create a new stylesheet like above?

amundsen_application/static/js/components/common/ResourceListItem/SchemaInfo/index.tsx
dikshathakur3119: If I do not use span, it given an error on browser. Let me know if it works for you

Golodhros: I see, sure this is OK too

amundsen_application/static/js/components/common/ResourceListItem/SchemaInfo/styles.scss
Golodhros: Maybe you want to bubble it up to the ResourceListItem level, on its styles.scss file.

You will probably be able to find a variable with that color.

amundsen_application/static/js/components/common/ResourceListItem/SchemaInfo/index.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import { OverlayTrigger, Popover } from 'react-bootstrap';
+
+import './styles.scss';
+
+export interface SchemaInfoProps{
+  schema?: string;
```
ttannis, line 7: Currently using the `?` means that all of these values are optional. However it's worth considering scoping this to what our current use cases are in order to simplify implementation. Essentially while we like to make things flexible, making it too flexible has the effect of creating extra unnecessary work...like have to make sure you check for `null`.

I think it is fair to make `schema` required given that Amundsen is designed to expect this information elsewhere. If you choose to make it required, then you can also remove it from the `defaultProps` below since a required value won't need a default.

amundsen_application/static/js/components/common/ResourceListItem/SchemaInfo/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+     >
+      <span className="underline">{ schema }</span>
+    </OverlayTrigger>
+  );
+};
+
+SchemaInfo.defaultProps = {
+  schema: '',
+  placement: 'bottom',
+  desc: 'Main source of all events data, usually recommended for use, Data is usually 24-48 hours delayed.'
```
ttannis, line 36: I assume this is just being used for development purposes, but we won't want a default description here. We'd want to pass the `schema_description` from the API. 

So for the actual implementation there should be no default description. Instead we would have some boolean logic to render either the plain text or the text with the `OverlayTrigger`.

amundsen_application/static/js/components/common/ResourceListItem/SchemaInfo/styles.scss
ttannis: For some reason I can't find the actual hex color value in Figma, let's double check with design about what color this is. Then either find it or enter it in `_variables-default.scss`. I wonder if it's supposed to be the same color as $stroke-focus.

amundsen_application/static/js/components/common/ResourceListItem/SchemaInfo/styles.scss
```diff
@@ -0,0 +1,5 @@
+@import 'variables';
+
+.underline{
```
ttannis, line 3: nit: add a space before the `{`

amundsen_application/static/js/components/common/ResourceListItem/SchemaInfo/index.tsx
dikshathakur3119: yes it is just for development purpose, I will remove default values. I planning to keep only  placement = bottom as default.
yes you are right, I will put a boolean logic if we do not get any schema_description

amundsen_application/static/js/components/common/ResourceListItem/SchemaInfo/styles.scss
dikshathakur3119: ok

amundsen_application/static/js/interfaces/Resources.ts
```diff
@@ -38,6 +38,7 @@ export interface TableResource extends Resource {
   last_updated_timestamp?: number;
   name: string;
   schema: string;
+  schema_description: string;
```
ttannis, line 41: This should be an optional value using ` schema_description?: string;`. As you may have noticed this value only exists on the the search results, and there are other parts of the application where the `TableListItem` shows up where we won't be showing this. Making this optional will also prevent you from having to have added this into all the other files that aren't particularly related to this feature.

amundsen_application/static/js/fixtures/search/inlineResults.ts
```diff
@@ -68,6 +68,7 @@ export const allResourcesExample = {
         last_updated_timestamp: 946684799,
         name: 'testName1',
         schema: 'testSchema',
+        schema_description: 'testSchemaDescription',
```
ttannis, line 71: Can revert changes to this file after making `schema_description` optional.

amundsen_application/static/js/fixtures/globalState.ts
```diff
@@ -29,6 +29,7 @@ const globalState: GlobalState = {
           description: 'description',
           name: 'name',
           schema: 'schema',
+          schema_description: 'schemaDescription',
```
ttannis, line 32: Can revert changes to this file after making `schema_description` optional.

amundsen_application/static/js/ducks/search/tests/sagas.spec.ts
```diff
@@ -61,6 +61,7 @@ describe('search sagas', () => {
           last_updated_timestamp: 946684799,
           name: 'testName',
           schema: 'testSchema',
+          schema_description: 'testSchemaDescription',
```
ttannis, line 64: Can revert changes to this file after making `schema_description` optional.

amundsen_application/static/js/ducks/search/tests/reducer.spec.ts
```diff
@@ -67,6 +67,7 @@ describe('search reducer', () => {
           last_updated_timestamp: 946684799,
           name: 'testName',
           schema: 'testSchema',
+          schema_description: 'testSchemaDescription',
```
ttannis, line 70: Can revert changes to this file after making `schema_description` optional.

amundsen_application/static/js/ducks/bookmark/tests/index.spec.ts
```diff
@@ -44,6 +44,7 @@ describe('bookmark ducks', () => {
           description: 'description',
           name: 'name',
           schema: 'schema',
+          schema_description: 'schemaDescription',
```
ttannis, line 47: Can revert changes to this file after making `schema_description` optional.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
```diff
@@ -38,7 +39,7 @@ class TableListItem extends React.Component<TableListItemProps, {}> {
             <div className="resource-info-text my-auto">
               <div className="resource-name title-2">
                 <div className="truncated">
-                  { `${table.schema}.${table.name}`}
+                  <SchemaInfo schema={ table.schema } desc={ table.schema_description }placement="bottom"/>.{ table.name }
```
ttannis, line 42: We won't want to render a `SchemaInfo` if there is no `schema_description`. I would suggest:
1. If there is not `schema_description` return existing code -- `${table.schema}.${table.name}`.
2. Else render a `SchemaInfo`. In this case I think `SchemaInfo` should take `table.name` as a parameter as well, and own rendering the entire text.

amundsen_application/static/js/components/common/Bookmark/MyBookmarks/index.spec.tsx
```diff
@@ -36,6 +36,7 @@ describe('MyBookmarks', () => {
             description: 'description',
             name: 'name',
             schema: 'schema',
+            schema_description: 'schemaDescription',
```
ttannis, line 39: Can revert changes to this file after making schema_description optional.

amundsen_application/api/utils/search_utils.py
```diff
@@ -21,6 +21,7 @@ def map_table_result(result: Dict) -> Dict:
         'description': result.get('description', None),
         'database': result.get('database', None),
         'schema': result.get('schema', None),
+        'schema_description': result.get('schema_description', None),
```
ttannis, line 24: Regarding the failing test, you will want to edit this file https://github.com/lyft/amundsenfrontendlibrary/blob/master/tests/unit/api/search/test_v0.py. The test is failing because now we expect the search endpoint to process a `schema_description`, but the key is missing in `MOCK_PARSED_RESULTS`. So you can either:
1. Add some dummy value for this like `'schema_description': 'test',` to `MOCK_TABLE_RESULTS`, and `MOCK_PARSED_RESULTS`
2. Add `'schema_description': None,` to `MOCK_PARSED_RESULTS`.

amundsen_application/static/js/components/common/ResourceListItem/styles.scss
```diff
@@ -85,4 +85,8 @@
   .title-2 {
     color: $resource-title-color;
   }
+
+  .underline {
+    border-bottom: solid #9191A8;
```
ttannis, line 90: Something that is missing on the styling is the gray background that is supposed to show up on hover. I notice in Figma the styles for the rectangle are as follows:


So to achieve this we could change this to:
```
.underline {
    box-shadow: 0px 1px 0px $gray40; // can also make a new variable for this

    &:hover,
    &:focus {
      background: $btn-default-bg-hover;
    }
  }
```

Lastly, this file is appropriate for styles that will be shared across the components, but since this style is specific to `SchemaInfo`, I would actually create a styles file for `SchemaInfo` and put it there instead.

amundsen_application/static/js/interfaces/Resources.ts
dikshathakur3119: okay now I understand why I have to make changes into other files. Thanks!

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
dikshathakur3119: In my current implementation, `SchemaInfo` checks if `schema_description` exists or not and returns text accordingly. I will change this as per above suggestion.

ttannis: Actually you really could so it either way. I'd say the main concept here is having logic consolidated into the appropriate chunks. If you want `SchemaInfo` to own the logic for rendering either case, that is also fine. 

This is one if the reason I suggested `SchemaInfo` would take `table.name` as well. So you can also maintain your current implementation but address my 2nd comment.

amundsen_application/static/js/components/common/ResourceListItem/SchemaInfo/styles.scss
ttannis: Just to resolve this one, the color is $gray40.

amundsen_application/static/js/components/common/ResourceListItem/TableListItem/index.tsx
dikshathakur3119: I changed the as per your suggestion, I think by doing this logic is more clear and understandable. Thanks for your suggestion.

dikshathakur3119:
Hey @ttannis @Golodhros ,  How can I run codecov report after making commits explicitly?

dikshathakur3119:
👀 @ttannis @Golodhros

ttannis:
> Hey @ttannis @Golodhros , How can I run codecov report after making commits explicitly?

It should do it automatically, though sometimes there is some lag/delay and not guaranteed to run immediately.

amundsen_application/static/js/components/common/ResourceListItem/SchemaInfo/styles.scss
```diff
@@ -0,0 +1,10 @@
+@import 'variables';
+
+.underline {
+    box-shadow: 0px 1px 0px $stroke-underline;
+
+    &:hover,
+    &:focus {
+      background: $btn-default-bg-hover;
+    }
+  }
```
ttannis, line 10:
```suggestion
.underline {
  box-shadow: 0px 1px 0px $stroke-underline;

  &:hover,
  &:focus {
    background: $btn-default-bg-hover;
  }
}
```
nit: indentation is off

amundsen_application/static/js/components/common/ResourceListItem/styles.scss
```diff
@@ -85,4 +85,5 @@
   .title-2 {
     color: $resource-title-color;
   }
+
```
ttannis, line 88: nit: don't need the extra space

amundsen_application/static/js/components/common/ResourceListItem/SchemaInfo/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const SchemaInfo: React.SFC<SchemaInfoProps>  = ({ schema, table, desc, placement}) => {
+  const popoverHoverFocus = (
+   <Popover id="popover-trigger-hover-focus">
+     <strong>{ schema }:</strong> { desc }
+   </Popover>
+ );
+
+  return (
+    <div>
```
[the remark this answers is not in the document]
Golodhros: +1

dikshathakur3119:
👀 @ttannis @Golodhros

amundsen_application/static/js/components/common/ResourceListItem/SchemaInfo/tests/index.spec.tsx
```diff
@@ -0,0 +1,55 @@
...
+      );
+      expect(wrapper.find(OverlayTrigger).props().overlay).toEqual(expectedPopover);
+    });
+
+    it('renders OverlayTrigger with correct schema name', () => {
+      expect(wrapper.find(OverlayTrigger).find('.underline').text()).toEqual(props.schema);
+    });
+
+    it('renders correct table name', () => {
+      expect(wrapper.children().at(1).text()).toEqual('.');
```
ttannis, line 51: nit: For this particular text I personally wouldn't go so deep and instead just check that it renders with the correct text: "expect(wrapper.children().text()).toEqual(`${props.schema}.${props.table}`);"

I do see that the test above this ones verifies `props.schema`, but you could consider that that unit test verifies the specific styling that is part of this feature, while this unit test verifies the text for this feature.

dikshathakur3119: yes you are right

dikshathakur3119:
👀 @ttannis

dikshathakur3119:
💨

PR #467, opened 2020-05-28
feat: Related Dashboards endpoint on Flask app

Golodhros:
### Summary of Changes
Adds a new endpoint

### Tests
Adds API tests

### Documentation

_What documentation did you add or modify and why? Add any relevant links then remove this line_

### CheckList
Make sure you have checked **all** steps below to ensure a timely review.
- [ ] PR title addresses the issue accurately and concisely. Example: "Updates the version of Flask to v1.0.2"
    - In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).
- [ ] PR includes a summary of changes, including screenshots of any UI changes. 
- [ ] PR adds unit tests, updates existing unit tests, __OR__ documents why no test additions or modifications are needed.
- [ ] In case of new functionality, my PR adds documentation that describes how to use it.
    - All the public python functions and the classes in the PR contain docstrings that explain what it does
- [ ] PR passes all tests documented in the [developer guide](https://github.com/lyft/amundsenfrontendlibrary/blob/master/docs/developer_guide.md#testing)

PR #469, opened 2020-05-29
fix: Miscellaneous UI Fixes

ttannis:
### Summary of Changes
This PR make a series of UI and style fixes described below.  Each commit is scoped to a particular task for easier review.

#### 1. Fix Dashboard Preview Title
This fix is also included in https://github.com/lyft/amundsenfrontendlibrary/pull/467, but will just resolve on merge vs. reverting the commit.

#### 2. Improve Resource Header Vertical Spacing Balance
**Before:**
image
**After:**
image

#### 3. Min-height Metadata Sections
Design proposed to make the metadata sections in the left panel have a min-height, which will currently be 70px. The reason for this is so that most pages becomes more aesthetically pleasing by having column items align vs. before where many cases would be just slightly off. The design sentiment is that either items will align or be very clearly not aligned. 

**Before:** (items slightly off)
    

**After:**
    

#### 4. Update EditableSection
A readOnly `EditableSection` was rendering the edit button and the empty popover when not given the optional `editText` or `editUrl`. This fix adds a default `editText` prompt, and also does not render the button if there is no `editUrl` provided -- since the button will have no effect in that case.
**Before:**

**After:** (no button)

#### 5. Fix Resource Detail Height
This was another area where we needed to switch to `100vh` when calculating the height.

#### 6. Hover Effects on Service Links
Updated the component style on hover.
_Example_:

### Tests

Added a test to `EditableSection`.

### Documentation

N/A for UI fixes

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
+++ b/amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
@@ -33,7 +33,7 @@ const PreviewModal = ({ imageSrc, onClose }: PreviewModalProps) => {
   return (
     <Modal show={show} onHide={handleClose} scrollable="true" className="dashboard-preview-modal">
       <Modal.Header closeButton={true}>
-        <Modal.Title className="text-center">Constants.DASHBOARD_PREVIEW_MODAL_TITLE</Modal.Title>
+        <Modal.Title className="text-center">{Constants.DASHBOARD_PREVIEW_MODAL_TITLE}</Modal.Title>
       </Modal.Header>
       <Modal.Body>
         <img
--- a/amundsen_application/static/js/components/DashboardPage/index.tsx
+++ b/amundsen_application/static/js/components/DashboardPage/index.tsx
@@ -207,40 +207,48 @@ export class DashboardPage extends React.Component<DashboardPageProps, Dashboard
             </EditableSection>
             <section className="column-layout-2">
               <section className="left-panel">
-                <div className="section-title title-3">Owners</div>
-                <div>
-                  {
-                    dashboard.owners.length > 0 &&
-                    dashboard.owners.map(owner =>
-                      <Link
-                        key={owner.user_id}
-                        to={`/user/${owner.user_id}?source=${DASHBOARD_OWNER_SOURCE}`}>
-                        <AvatarLabel
-                          label={owner.display_name}/>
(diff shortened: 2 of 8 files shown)


ttannis:
A series of small fixes, each commit is scoped to a particular task for easier review.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
```diff
@@ -33,7 +33,7 @@ const PreviewModal = ({ imageSrc, onClose }: PreviewModalProps) => {
   return (
     <Modal show={show} onHide={handleClose} scrollable="true" className="dashboard-preview-modal">
       <Modal.Header closeButton={true}>
-        <Modal.Title className="text-center">Constants.DASHBOARD_PREVIEW_MODAL_TITLE</Modal.Title>
+        <Modal.Title className="text-center">{Constants.DASHBOARD_PREVIEW_MODAL_TITLE}</Modal.Title>
```
Golodhros, line 36: Ah, thanks! I had it on my current PR!

amundsen_application/static/js/components/DashboardPage/index.tsx
```diff
@@ -207,40 +207,48 @@ export class DashboardPage extends React.Component<DashboardPageProps, Dashboard
...
-                <div className="section-title title-3">Last Updated</div>
-                <div className="body-2 text-primary">
-                  { formatDateTimeShort({ epochTimestamp: dashboard.updated_timestamp }) }
-                </div>
-                <div className="section-title title-3">Recent View Count</div>
-                <div className="body-2 text-primary">
-                  { dashboard.recent_view_count }
-                </div>
+                <section className="metadata-section">
+                  <div className="section-title title-3">Owners</div>
```
Golodhros, line 211: Maybe we could try to use an h3 or h4 here?

ttannis: I'll revisit this. If anything it would be `h4` because the existing `h3` style specifications don't match. But I want double check with design about our typography specifications. 

A couple of things I'm confused by and can't remember the context of, such as the deal with our `title-#` styles -- for example, `<h4 className="section-title title-3">` makes me question if we also need to revisit the `title-#` styles to understand what styles should be applied by default to all headers across the app.

PR #473, opened 2020-06-03
fix: Updates the Dashboard Page URL

Golodhros:
### Summary of Changes
Updates the Dashboard Page URL to avoid depending on the query parameters
Flattened some components folder structure, removing /tests folders

### Tests
Updated tests

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/TableDetail/index.spec.tsx
+++ b/amundsen_application/static/js/components/TableDetail/index.spec.tsx
@@ -10,27 +10,27 @@ import LoadingSpinner from '../common/LoadingSpinner';
 import { TableDetail, TableDetailProps, MatchProps } from './';
 
 
-describe('TableDetail', () => {
-
-  const setup = (propOverrides?: Partial<TableDetailProps>, location?: Partial<History.Location>) => {
-    const routerProps = getMockRouterProps<MatchProps>({
-      "cluster":"gold",
-      "database":"hive",
-      "schema":"base",
-      "table":"rides"
-    }, location);
-    const props = {
-      isLoading: false,
-      statusCode: 200,
-      tableData: tableMetadata,
-      getTableData: jest.fn(),
-      ...routerProps,
-      ...propOverrides,
-    };
-    const wrapper = mount<TableDetail>(<TableDetail {...props} />);
-
-    return { props, wrapper };
+const setup = (propOverrides?: Partial<TableDetailProps>, location?: Partial<History.Location>) => {
+  const routerProps = getMockRouterProps<MatchProps>({
+    "cluster":"gold",
+    "database":"hive",
+    "schema":"base",
+    "table":"rides"
+  }, location);
+  const props = {
+    isLoading: false,
+    statusCode: 200,
+    tableData: tableMetadata,
+    getTableData: jest.fn(),
+    ...routerProps,
+    ...propOverrides,
   };
+  const wrapper = mount<TableDetail>(<TableDetail {...props} />);
+
+  return { props, wrapper };
+};
+
+describe('TableDetail', () => {
 
   describe('render', () => {
 
--- a/amundsen_application/static/js/components/common/ResourceListItem/DashboardListItem/index.tsx
+++ b/amundsen_application/static/js/components/common/ResourceListItem/DashboardListItem/index.tsx
@@ -23,7 +23,11 @@ export interface DashboardListItemProps {
 class DashboardListItem extends React.Component<DashboardListItemProps, {}> {
   getLink = () => {
     const { dashboard, logging } = this.props;
-    return `/dashboard?uri=${dashboard.uri}&index=${logging.index}&source=${logging.source}`;
+    const { cluster, product, group_url, uri } = dashboard;
+    const groupID = group_url.split('/').pop();
+    const dashboardName = uri.split('/').pop();
+
+    return `/dashboard/${cluster}/${product}/${groupID}/${dashboardName}?index=${logging.index}&source=${logging.source}`;
   };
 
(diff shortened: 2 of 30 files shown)


amundsen_application/static/js/components/TableDetail/index.spec.tsx
```diff
@@ -10,27 +10,27 @@ import LoadingSpinner from '../common/LoadingSpinner';
...
-      statusCode: 200,
-      tableData: tableMetadata,
-      getTableData: jest.fn(),
-      ...routerProps,
-      ...propOverrides,
-    };
-    const wrapper = mount<TableDetail>(<TableDetail {...props} />);
-
-    return { props, wrapper };
+const setup = (propOverrides?: Partial<TableDetailProps>, location?: Partial<History.Location>) => {
```
ttannis, line 13: Any benefit to moving this outside, or is it perhaps just more readable this way?

amundsen_application/static/js/components/common/ResourceListItem/DashboardListItem/index.tsx
```diff
@@ -23,7 +23,11 @@ export interface DashboardListItemProps {
 class DashboardListItem extends React.Component<DashboardListItemProps, {}> {
   getLink = () => {
     const { dashboard, logging } = this.props;
-    return `/dashboard?uri=${dashboard.uri}&index=${logging.index}&source=${logging.source}`;
+    const { cluster, product, group_url, uri } = dashboard;
+    const groupID = group_url.split('/').pop();
+    const dashboardName = uri.split('/').pop();
+
+    return `/dashboard/${cluster}/${product}/${groupID}/${dashboardName}?index=${logging.index}&source=${logging.source}`;
```
ttannis, line 30: How did we come up with this specific path? If we use new concepts that the frontend has to parse and generate, then we end up introducing hardcoded url & uri processing logic that is specific to Mode. If community members are using some other tool, this parsing logic could pose issues for them unless all BI tools somehow format their values the same exact way. 

I would propose that we only use some combination of the exact values that come from the metadata service, and use `group_name` and `name` directly from our metadata.

amundsen_application/static/js/components/TableDetail/index.spec.tsx
Golodhros: Yeah, I only see it as a readability matter. Is that OK?

ttannis: No problem

amundsen_application/static/js/components/common/ResourceListItem/DashboardListItem/index.tsx
Golodhros: This was a suggestion from @danwom, happy to adapt it.

I will follow your suggestion and move the URL creation into a helper so it is easy to change in the future.

Golodhros:
> We also need to update the inline search result item that shows up in the search bar: https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/components/common/SearchBar/InlineSearchResults/index.tsx#L99
> 
> Beyond that, one concern about how we are forming the new url.

Thanks a lot for the heads up!
I will work on it!

amundsen_application/static/js/utils/navigationUtils.ts
```diff
@@ -2,6 +2,7 @@ import * as qs from 'simple-query-string';
 import { createBrowserHistory } from 'history';
 
 import { ResourceType } from 'interfaces/Resources';
+import { MatchProps } from 'components/DashboardPage';
```
ttannis, line 5: nit: delete? don't see this type used here.

amundsen_application/static/js/components/common/SearchBar/InlineSearchResults/index.tsx
```diff
@@ -5,6 +5,7 @@ import SearchItemList from './SearchItemList';
 import ResultItemList from './ResultItemList';
 
 import { getSourceDisplayName, getSourceIconClass, indexDashboardsEnabled, indexUsersEnabled } from 'config/config-utils';
+import { buildDashboardURL } from '../../../../utils/navigationUtils';
```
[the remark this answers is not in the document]
Golodhros: much better, changing!
Thanks!

amundsen_application/static/js/utils/navigationUtils.ts
```diff
@@ -47,3 +48,16 @@ export const updateSearchUrl = (searchParams: SearchParams, replace: boolean = f
     BrowserHistory.push(newUrl);
   }
 };
+
+/**
+ * Creates the dashboard detail URL from the URI
+ * @param URI String  URI of the dashboard, it has this shape: uri = "<product>_dashboard://<cluster>.<groupID>/<dashboardID>"
+ * @return String     Dashboard Detail page URL
+ */
+export const buildDashboardURL = (URI: string) => {
```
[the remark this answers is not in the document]
Golodhros: Yep, writing one now.
Do you want me to add the product parameter already?

amundsen_application/static/js/utils/navigationUtils.ts
Golodhros: Good call, it was used before.

Thanks

amundsen_application/static/js/utils/navigationUtils.ts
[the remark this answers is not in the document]
ttannis: No need. We can add it when it's necessary, what's valuable is we made a decision that set us up to easily address the case without community members having to struggle too much to refactor anything.

PR #479, opened 2020-06-12
fix: Fixes scrolling on results pages and pagination

Golodhros:
### Summary of Changes
Makes the results and the sidebar of the search page scrollable
Adjusts spacing on the dashboard detail and table detail sidebar

### CheckList

### DIFF
--- a/amundsen_application/static/css/_layouts.scss
+++ b/amundsen_application/static/css/_layouts.scss
@@ -95,12 +95,12 @@ $aside-separation-space: 40px;
       border-right: 4px solid $divider;
       flex-basis: 600px;
       flex-shrink: 0;
-      overflow-y: scroll;
-      padding: $aside-separation-space 24px;
+      overflow-y: auto;
+      padding: 0 24px $aside-separation-space;
 
       > .banner {
         border: 1px solid $stroke;
-        height: $aside-separation-space;
+        height: 40px;
         margin: 24px 24px 0;
         padding: 8px;
       }
@@ -156,7 +156,6 @@ $aside-separation-space: 40px;
   // Inner column layout
   .column-layout-2 {
     display: flex;
-    padding-top: $aside-separation-space;
 
     > .left-panel {
       flex-basis: 262px;
--- a/amundsen_application/static/css/_layouts.scss
+++ b/amundsen_application/static/css/_layouts.scss
@@ -1,5 +1,7 @@
 @import 'variables';
+
 $resource-header-height: 84px;
+$aside-separation-space: 40px;
 
 .resource-detail-layout {
   height: calc(100vh - #{$nav-bar-height} - #{$footer-height});
@@ -94,11 +96,11 @@ $resource-header-height: 84px;
       flex-basis: 600px;
       flex-shrink: 0;
       overflow-y: scroll;
-      padding: 0 24px;
+      padding: $aside-separation-space 24px;
 
       > .banner {
         border: 1px solid $stroke;
-        height: 40px;
+        height: $aside-separation-space;
         margin: 24px 24px 0;
         padding: 8px;
       }
@@ -111,8 +113,11 @@ $resource-header-height: 84px;
       .editable-section,
       .metadata-section {
         display: inline-table;
-        margin-top: 40px;
         min-height: 70px;
+      }
+
+      .metadata-section {
+        margin-top: $aside-separation-space;
       }
 
       .editable-text {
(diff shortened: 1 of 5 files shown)


amundsen_application/static/css/_layouts.scss
```diff
@@ -1,5 +1,7 @@
 @import 'variables';
+
```
ttannis, line 2: What are the changes in the file addressing? The metadata sections are no longer aligned in the way that we want them to be.

Golodhros: Oh, I see, I will double check, thanks!

There was a huge space at the bottom of the sidebar that was triggering a y-scroll when it wasn't necessary.

Golodhros: Oh, yeah, that was pretty obvious there, I don't know how it went through!

Just sent a fix.

amundsen_application/static/css/_layouts.scss
```diff
@@ -154,7 +156,7 @@ $resource-header-height: 84px;
   // Inner column layout
   .column-layout-2 {
     display: flex;
-    flex-basis: 100%;
+    padding-top: $aside-separation-space;
```
ttannis, line 159: Any reason why the flex-basis was removed? I don't remember when or why it was added thought.
Also the comment above about the padding causing too much space at the top.

amundsen_application/static/css/_layouts.scss
```diff
@@ -110,8 +112,8 @@ $resource-header-height: 84px;
 
       .editable-section,
       .metadata-section {
+        margin-top: $aside-separation-space;
```
ttannis, line 115: Unrelated Question: Did the lint work not automatically put this in alphabetical order? Thought it would happen automatically on pre-commit.

amundsen_application/static/css/_layouts.scss
```diff
@@ -94,11 +96,11 @@ $resource-header-height: 84px;
       flex-basis: 600px;
       flex-shrink: 0;
       overflow-y: scroll;
-      padding: 0 24px;
+      padding: $aside-separation-space 24px;
```
ttannis, line 99: We now have double the amount of space between the header and the first section in the left-panel. So I think we'd want to remove this padding here and on line 159.

amundsen_application/static/css/_layouts.scss
ttannis: Taking another look, still have a few more comments. Essentially I would expect this PR to result in the page looking exactly the same except for what is needed to fix the issue you referenced here.

amundsen_application/static/css/_layouts.scss
```diff
@@ -94,11 +96,11 @@ $resource-header-height: 84px;
...
 
       > .banner {
         border: 1px solid $stroke;
-        height: 40px;
+        height: $aside-separation-space;
```
ttannis, line 103: nit: I think it's a coincidence that the spacing and banner height are the same value. I'd feel better making this it's own variable, because design could let's say change the spacing but not change the banner.

amundsen_application/static/css/_layouts.scss
Golodhros: That was the rule that was adding too much space and triggering a scroll.

amundsen_application/static/css/_layouts.scss
Golodhros: Are you sure? 
I am comparing it with Staging and looks the same to me!

amundsen_application/static/css/_layouts.scss
Golodhros: OK, I will rename it.

amundsen_application/static/css/_layouts.scss
Golodhros: Got it, fixing

amundsen_application/static/css/_layouts.scss
ttannis: I see, no problem them. I did some due diligence to try to figure out from the history if `flex-basis` was added for a reason (like to fix a bug) and it doesn't look like it.  May have just been a style we didn't need that was left over from the iterations we went through when we did the resource page redesign last year.

amundsen_application/static/css/_layouts.scss
Golodhros: Interesting, maybe the alphabetic ordering was part of the sass guidelines that I didn't keep.

Do you want me to find that rule and add it?

ttannis: I called it out in case this is a sign of something being broken vs the support not being there. If it's not there, I can always add it later.

PR #480, opened 2020-06-12
fix: Fixes sectioning and accessibility issues all over the application

Golodhros:
### Summary of Changes
Fixes sectioning markup
Fixes accessibility issues on all the application

### Tests
Updated tests

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/AnnouncementPage/index.tsx
+++ b/amundsen_application/static/js/components/AnnouncementPage/index.tsx
@@ -52,17 +52,17 @@ export class AnnouncementPage extends React.Component<AnnouncementPageProps> {
   render() {
     return (
       <DocumentTitle title="Announcements - Amundsen">
-        <div className="container announcement-container">
+        <main className="container announcement-container">
           <div className="row">
             <div className="col-xs-12">
-              <h3 id="announcement-header">Announcements</h3>
+              <h1 id="announcement-header" className="h3">Announcements</h1>
               <hr />
               <div id="announcement-content" className="announcement-content">
                 {this.createPosts()}
               </div>
             </div>
           </div>
-        </div>
+        </main>
       </DocumentTitle>
     );
   }
--- a/amundsen_application/static/js/components/AnnouncementPage/index.tsx
+++ b/amundsen_application/static/js/components/AnnouncementPage/index.tsx
@@ -55,7 +55,9 @@ export class AnnouncementPage extends React.Component<AnnouncementPageProps> {
         <main className="container announcement-container">
           <div className="row">
             <div className="col-xs-12">
-              <h1 id="announcement-header" className="h3">Announcements</h1>
+              <h1 id="announcement-header" className="h3">
+                Announcements
+              </h1>
               <hr />
               <div id="announcement-content" className="announcement-content">
                 {this.createPosts()}
--- a/amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
+++ b/amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
@@ -43,7 +43,7 @@ const PreviewModal = ({ imageSrc, onClose }: PreviewModalProps) => {
         </Modal.Title>
       </Modal.Header>
       <Modal.Body>
-        <img src={imageSrc} height="auto" width="100%" />
+        <img src={imageSrc} height="auto" width="100%" alt="" />
       </Modal.Body>
     </Modal>
   );
--- a/amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
+++ b/amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
@@ -89,6 +89,7 @@ export class ImagePreview extends React.Component<
             type="button"
             onClick={this.handlePreviewButton}
           >
+            <span className="sr-only">{Constants.PREVIEW_BUTTON_TEXT}</span>
             <img
               className="preview"
               style={
@@ -99,6 +100,7 @@ export class ImagePreview extends React.Component<
               onError={this.onError}
               height="auto"
               width="100%"
+              alt=""
             />
           </button>
         )}
--- a/amundsen_application/static/js/components/DashboardPage/index.tsx
+++ b/amundsen_application/static/js/components/DashboardPage/index.tsx
@@ -168,7 +168,7 @@ export class DashboardPage extends React.Component<
             />
           </div>
           <div className="header-section header-title">
-            <h3 className="header-title-text truncated">{dashboard.name}</h3>
+            <h1 className="h3 header-title-text truncated">{dashboard.name}</h1>
             <BookmarkIcon
               bookmarkKey={dashboard.uri}
               resourceType={ResourceType.dashboard}
@@ -203,7 +203,7 @@ export class DashboardPage extends React.Component<
           </div>
         </header>
         <article className="column-layout-1">
-          <section className="left-panel">
+          <aside className="left-panel">
             <EditableSection
               title="Description"
               readOnly
@@ -322,8 +322,8 @@ export class DashboardPage extends React.Component<
               </section>
             </section>
             <ImagePreview uri={this.state.uri} redirectUrl={dashboard.url} />
-          </section>
-          <section className="right-panel">{this.renderTabs()}</section>
+          </aside>
+          <main className="right-panel">{this.renderTabs()}</main>
         </article>
       </div>
     );
--- a/amundsen_application/static/js/components/DashboardPage/index.tsx
+++ b/amundsen_application/static/js/components/DashboardPage/index.tsx
@@ -257,19 +257,19 @@ export class DashboardPage extends React.Component<
                 </section>
                 <section className="metadata-section">
                   <div className="section-title title-3">Created</div>
-                  <div className="body-2 text-primary">
+                  <time className="body-2 text-primary">
                     {formatDateTimeShort({
                       epochTimestamp: dashboard.created_timestamp,
                     })}
-                  </div>
+                  </time>
                 </section>
                 <section className="metadata-section">
                   <div className="section-title title-3">Last Updated</div>
-                  <div className="body-2 text-primary">
+                  <time className="body-2 text-primary">
                     {formatDateTimeShort({
                       epochTimestamp: dashboard.updated_timestamp,
                     })}
-                  </div>
+                  </time>
                 </section>
                 <section className="metadata-section">
                   <div className="section-title title-3">Recent View Count</div>
@@ -289,25 +289,25 @@ export class DashboardPage extends React.Component<
                   <div className="section-title title-3">
                     Last Successful Run
                   </div>
-                  <div className="last-successful-run-timestamp body-2 text-primary">
+                  <time className="last-successful-run-timestamp body-2 text-primary">
                     {dashboard.last_successful_run_timestamp
                       ? formatDateTimeShort({
                           epochTimestamp:
                             dashboard.last_successful_run_timestamp,
                         })
                       : NO_TIMESTAMP_TEXT}
-                  </div>
+                  </time>
                 </section>
                 <section className="metadata-section">
                   <div className="section-title title-3">Last Run</div>
                   <div>
-                    <div className="last-run-timestamp body-2 text-primary">
+                    <time className="last-run-timestamp body-2 text-primary">
                       {dashboard.last_run_timestamp
                         ? formatDateTimeShort({
                             epochTimestamp: dashboard.last_run_timestamp,
                           })
                         : NO_TIMESTAMP_TEXT}
-                    </div>
+                    </time>
                     <div className="last-run-state">
                       <Flag
                         caseType="sentenceCase"
--- a/amundsen_application/static/js/components/Feedback/FeedbackForm/RatingFeedbackForm/index.tsx
+++ b/amundsen_application/static/js/components/Feedback/FeedbackForm/RatingFeedbackForm/index.tsx
@@ -34,8 +34,8 @@ export class RatingFeedbackForm extends AbstractFeedbackForm {
     return (
       <form id={AbstractFeedbackForm.FORM_ID} onSubmit={this.submitForm}>
         <input type="hidden" name="feedback-type" value="NPS Rating" />
-        <div className="form-group">
-          <label>{RATING_LABEL}</label>
+        <div className="form-group clearfix">
+          <label htmlFor="rating">{RATING_LABEL}</label>
           <div>
             <div className="radio-set">{radioButtonSet}</div>
             <div>
@@ -48,14 +48,19 @@ export class RatingFeedbackForm extends AbstractFeedbackForm {
             </div>
           </div>
         </div>
-        <textarea
-          className="form-control form-group"
-          name="comment"
-          form={AbstractFeedbackForm.FORM_ID}
-          rows={8}
-          maxLength={2000}
-          placeholder={COMMENTS_PLACEHOLDER}
-        />
+        <div className="form-group">
+          <label htmlFor="comment">{COMMENTS_PLACEHOLDER}</label>
+          <textarea
+            className="form-control form-group"
+            name="comment"
+            id="comment"
+            form={AbstractFeedbackForm.FORM_ID}
+            rows={8}
+            maxLength={2000}
+            placeholder={COMMENTS_PLACEHOLDER}
+          />
+        </div>
+
         <button className="btn btn-primary" type="submit">
           {SUBMIT_TEXT}
         </button>
--- a/amundsen_application/static/js/components/TableDetail/FrequentUsers/index.tsx
+++ b/amundsen_application/static/js/components/TableDetail/FrequentUsers/index.tsx
@@ -35,7 +35,6 @@ export function renderReader(
     >
       <Link
         className="avatar-overlap"
-        id="frequent-users"
         onClick={logClick}
         to={link}
         target={target}
--- a/amundsen_application/static/js/components/TableDetail/FrequentUsers/index.tsx
+++ b/amundsen_application/static/js/components/TableDetail/FrequentUsers/index.tsx
@@ -36,6 +36,7 @@ export function renderReader(
       <Link
         className="avatar-overlap"
         onClick={logClick}
+        data-type="frequent-users"
         to={link}
         target={target}
       >
--- a/amundsen_application/static/js/components/TableDetail/ReportTableIssue/index.spec.tsx
+++ b/amundsen_application/static/js/components/TableDetail/ReportTableIssue/index.spec.tsx
@@ -92,9 +92,9 @@ describe('ReportTableIssue', () => {
         const { wrapper } = setup();
         const previsOpenState = wrapper.state().isOpen;
 
-        wrapper
-          .instance()
-          .toggle({ currentTarget: { id: 'id', nodeName: 'button' } });
+        wrapper.instance().toggle({
+          currentTarget: { dataset: { type: 'id' }, nodeName: 'button' },
+        });
 
         expect(setStateSpy).toHaveBeenCalledWith({ isOpen: !previsOpenState });
       });
@@ -107,7 +107,7 @@ describe('ReportTableIssue', () => {
         // @ts-ignore: mocked events throw type errors
         wrapper.instance().submitForm({
           preventDefault: jest.fn(),
-          currentTarget: { id: 'id', nodeName: 'button' },
+          currentTarget: { dataset: { type: 'id' }, nodeName: 'button' },
         });
 
         expect(props.createIssue).toHaveBeenCalledWith(
@@ -123,7 +123,7 @@ describe('ReportTableIssue', () => {
         // @ts-ignore: mocked events throw type errors
         wrapper.instance().submitForm({
           preventDefault: jest.fn(),
-          currentTarget: { id: 'id', nodeName: 'button' },
+          currentTarget: { dataset: { type: 'id' }, nodeName: 'button' },
         });
 
         expect(wrapper.state().isOpen).toBe(false);
--- a/amundsen_application/static/js/components/TableDetail/ReportTableIssue/index.spec.tsx
+++ b/amundsen_application/static/js/components/TableDetail/ReportTableIssue/index.spec.tsx
@@ -93,7 +93,7 @@ describe('ReportTableIssue', () => {
         const previsOpenState = wrapper.state().isOpen;
 
         wrapper.instance().toggle({
-          currentTarget: { dataset: { type: 'id' }, nodeName: 'button' },
+          currentTarget: { id: 'id', nodeName: 'button' },
         });
 
         expect(setStateSpy).toHaveBeenCalledWith({ isOpen: !previsOpenState });
@@ -107,7 +107,7 @@ describe('ReportTableIssue', () => {
         // @ts-ignore: mocked events throw type errors
         wrapper.instance().submitForm({
           preventDefault: jest.fn(),
-          currentTarget: { dataset: { type: 'id' }, nodeName: 'button' },
+          currentTarget: { id: 'id', nodeName: 'button' },
         });
 
         expect(props.createIssue).toHaveBeenCalledWith(
@@ -123,7 +123,7 @@ describe('ReportTableIssue', () => {
         // @ts-ignore: mocked events throw type errors
         wrapper.instance().submitForm({
           preventDefault: jest.fn(),
-          currentTarget: { dataset: { type: 'id' }, nodeName: 'button' },
+          currentTarget: { id: 'id', nodeName: 'button' },
         });
 
         expect(wrapper.state().isOpen).toBe(false);
--- a/amundsen_application/static/js/components/TableDetail/WatermarkLabel/index.spec.tsx
+++ b/amundsen_application/static/js/components/TableDetail/WatermarkLabel/index.spec.tsx
@@ -83,14 +83,9 @@ describe('WatermarkLabel', () => {
         '2018-08-03',
         '2019-10-15'
       );
-      const startTime = mount(watermarkInfo)
-        .find('.date-range-value')
-        .at(0)
-        .text();
-      const endTime = mount(watermarkInfo)
-        .find('.date-range-value')
-        .at(1)
-        .text();
+      const watermark = mount(watermarkInfo);
+      const startTime = watermark.find('.date-range-value').at(0).text();
+      const endTime = watermark.find('.date-range-value').at(1).text();
 
       expect(startTime).toBe('Aug 03, 2018');
       expect(endTime).toBe('Oct 15, 2019');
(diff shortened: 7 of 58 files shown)


amundsen_application/static/js/components/Feedback/FeedbackForm/RatingFeedbackForm/index.tsx
```diff
@@ -48,14 +48,19 @@ export class RatingFeedbackForm extends AbstractFeedbackForm {
...
-        <textarea
-          className="form-control form-group"
-          name="comment"
-          form={AbstractFeedbackForm.FORM_ID}
-          rows={8}
-          maxLength={2000}
-          placeholder={COMMENTS_PLACEHOLDER}
-        />
+        <div className="form-group">
+          <label htmlFor="comment">{COMMENTS_PLACEHOLDER}</label>
```
ttannis, line 52: We could give an FYI to design about this by leaving a comment on the old design that didn't include any label. Will try to find the link for you assuming it is in Figma.

amundsen_application/static/js/components/SearchPage/SearchFilter/index.tsx
```diff
@@ -73,6 +73,7 @@ export const mapStateToProps = (state: GlobalState) => {
         type: categoryConfig.type,
       };
       if (categoryConfig.type === FilterType.CHECKBOX_SELECT) {
+        // eslint-disable-next-line @typescript-eslint/dot-notation
```
ttannis, line 76: What was the issue with this line? I've updated this in the filter PR and want to understand if the issue will still be applicable with the change I made.

amundsen_application/static/js/components/TableDetail/FrequentUsers/index.tsx
```diff
@@ -35,7 +35,6 @@ export function renderReader(
     >
       <Link
         className="avatar-overlap"
-        id="frequent-users"
```
ttannis, line 38: We need this id for logging purposes. https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/ducks/utilMethods.ts#L43

amundsen_application/static/js/components/TableDetail/WatermarkLabel/index.spec.tsx
```diff
@@ -83,9 +83,17 @@ describe('WatermarkLabel', () => {
         '2018-08-03',
         '2019-10-15'
       );
-      expect(mount(watermarkInfo).find('.range-dates').text()).toBe(
-        'Aug 03, 2018Oct 15, 2019'
-      );
+      const startTime = mount(watermarkInfo)
```
ttannis, line 86: nit: Create a variable instead `mount(watermarkInfo)` twice?

amundsen_application/static/js/components/TableDetail/WatermarkLabel/index.tsx
```diff
@@ -43,18 +43,24 @@ class WatermarkLabel extends React.Component<WatermarkLabelProps> {
...
-          {HIGH_WATERMARK_LABEL}
-        </div>
-        <div className="range-dates body-2">
-          {low && this.formatWatermarkDate(low)}
-          <br />
-          {high && this.formatWatermarkDate(high)}
-        </div>
-      </>
+      <div className="date-ranges">
+        {low && (
```
ttannis, line 47: Would want to verify the change this makes to the design is okay before changing it without it being requested.

amundsen_application/static/js/index.tsx
```diff
@@ -40,7 +40,7 @@ ReactDOM.render(
   <DocumentTitle title="Amundsen - Data Discovery Portal">
     <Provider store={store}>
       <Router history={BrowserHistory}>
-        <main id="main">
+        <div id="main">
```
ttannis, line 43: Why the change given that this was recently intentionally changed from `div` to `main`

amundsen_application/static/js/components/AnnouncementPage/index.tsx
```diff
@@ -52,17 +52,19 @@ export class AnnouncementPage extends React.Component<AnnouncementPageProps> {
   render() {
     return (
       <DocumentTitle title="Announcements - Amundsen">
-        <div className="container announcement-container">
+        <main className="container announcement-container">
           <div className="row">
             <div className="col-xs-12">
-              <h3 id="announcement-header">Announcements</h3>
+              <h1 id="announcement-header" className="h3">
```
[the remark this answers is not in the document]
Golodhros: Yes, you are right, we need to separate what is a styling matter (doing with classes) from what is a content structure matter (marked by the tags).

In general terms, all documents need to have at least one <h1>, and this is it for this page.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
```diff
@@ -43,7 +43,7 @@ const PreviewModal = ({ imageSrc, onClose }: PreviewModalProps) => {
         </Modal.Title>
       </Modal.Header>
       <Modal.Body>
-        <img src={imageSrc} height="auto" width="100%" />
+        <img src={imageSrc} height="auto" width="100%" alt="" />
```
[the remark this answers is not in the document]
Golodhros: Great question! Here is from w3.org:

"For example, the information provided by the image might already be given using adjacent text, or the image might be included to make the website more visually attractive. In these cases, a null (empty) alt text should be provided ( alt="" ) so that they can be ignored by assistive technologies, such as screen readers."

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
```diff
@@ -99,6 +100,7 @@ export class ImagePreview extends React.Component<
               onError={this.onError}
               height="auto"
               width="100%"
+              alt=""
```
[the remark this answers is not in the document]
Golodhros: The "sr-only" text is associated with the button, not really with the image. 
We can say that as a general rule, all buttons need to have some text inside, while the images can or cannot have an alternative text attribute.

amundsen_application/static/js/components/DashboardPage/index.tsx
```diff
@@ -203,7 +203,7 @@ export class DashboardPage extends React.Component<
           </div>
         </header>
         <article className="column-layout-1">
-          <section className="left-panel">
+          <aside className="left-panel">
```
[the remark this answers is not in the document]
Golodhros: Good question!

In general, we need to aim for having all this sections if possible: main, footer, header and aside.
They cannot be contained by each one, so the 'aside' needs to be outside the 'main', as well as the footer and the header.
On the other side, all 'div's don't really mean anything, they are kind of invisible, so we can add as many as we want.

amundsen_application/static/js/components/Feedback/FeedbackForm/RatingFeedbackForm/index.tsx
Golodhros: Yeah, I took the liberty of adding the label (I saw it wasn't on the design).

I get the impression (something we discussed some weeks ago I think) that designers think the placeholder attribute is enough for offering this information, but they are not aware the placeholder pattern doesn't really work for a11y.

I will add the comment, thanks!

amundsen_application/static/js/components/SearchPage/SearchFilter/index.tsx
Golodhros: Totally!
Our ESLint configuration includes a preference for the dot notation instead of the brackets, and running ESLint with the --fix flag keeps on chaning this line. 

However, when that change happens, it breaks the types. I am not sure if we are using the bracket notation just for that matter, so I commented it for now. I think that underlining this there is a type safety issue.

amundsen_application/static/js/components/TableDetail/FrequentUsers/index.tsx
Golodhros: Thanks for the heads up!!

I will use a data attribute instead, as these are repeating ids on several elements.

amundsen_application/static/js/components/TableDetail/WatermarkLabel/index.spec.tsx
Golodhros: Makes sense! Thanks!

amundsen_application/static/js/components/TableDetail/WatermarkLabel/index.tsx
Golodhros: It looks the same, but I can double check!

amundsen_application/static/js/index.tsx
Golodhros: Yep, that first change was a shortcut to have always a main. The problem is that aside, footer and header can't go inside the main element.
With this PR, the markup is in a much better place, although it took some extra work.

amundsen_application/static/js/components/AnnouncementPage/index.tsx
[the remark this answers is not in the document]
ttannis: > In general terms, all documents need to have at least one 'h1'

For what reason? I don't know the full context of where that recommendation comes from, but I think about how if our designs don't call for an `h1, then we wouldn't use an `h1`. 
I'm in favor of semantic html where it makes sense but I'm currently not understanding the benefit of transforming an `h1` to look like an `h3` just to have the `h1` present.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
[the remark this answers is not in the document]
ttannis: Ah noted that this is for the button. Is there a reason why you prefer to add an element with `sr-only` vs. using the `aria-label` button attribute? 

I'm partial to `aria-label` because it overrides whatever text gets nested in the button as well and gives us full control in every case I can currently think of, whereas with the `sr-only` approach I wonder what we would do if there was a case where it makes sense to have certain text in the button for sighted users whereas screen readers might actually need something with more detail. Thoughts?

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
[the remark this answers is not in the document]
ttannis: Yes, so in this case are you leaving it empty because the modal title says `Dashboard Preview`, therefore giving context about this image?

amundsen_application/static/js/components/DashboardPage/index.tsx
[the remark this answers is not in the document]
ttannis: > we need to aim for having all this sections if possible: main, footer, header and aside.

For this page in particular though, these changes *do* make sense so not requesting changes, but where does that recommendation come from? As you noted in a comment below, a11y concepts are not guaranteed to be focused on in our designs and I'm cautious of adhering to recommendations that our app might not have been designed for.

amundsen_application/static/js/components/Feedback/FeedbackForm/RatingFeedbackForm/index.tsx
ttannis: I found it in "Master -- Screens & UI Library", and it turns out recommendation in Figma does have a label. So we're set:
image

amundsen_application/static/js/components/SearchPage/SearchFilter/index.tsx
ttannis: Got it. I'd have to look into it myself as well to understand what the long term solution is, so disabling the lint for this line makes sense.

amundsen_application/static/js/index.tsx
ttannis: Got it. Noted that the `main` is now inside the pages themselves.

amundsen_application/static/js/components/TableDetail/WatermarkLabel/index.tsx
ttannis: It's minor to me (yet I don't want to make assumptions about what design cares about vs not) but its how the date texts align stacked on each other. 

**From Figma:**


**This PR:**

amundsen_application/static/js/components/AnnouncementPage/index.tsx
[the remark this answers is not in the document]
Golodhros: Just in that link you shared (http://web-accessibility.carnegiemuseums.org/foundations/semantic/) there is a mention: "Use one H1 per page, matching the page title." We didn't do the "matching" part though.

Again, this doesn't really have to do with the designs, as they deal with the aesthetics , not with the document structure. In this case, we are using that "h3" because its content is good enough for our h1. We could also create a hidden h1 as we have done in other pages, but I think this solution is more effective.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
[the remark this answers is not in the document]
Golodhros: That's a great question!

It seems that in general the 'sr-only' class is more reliable than the aria label, here is an excerpt from (https://accessible360.com/accessible360-blog/use-aria-label-screen-reader-text/): 
"the main upside to using screen reader only text over an aria-label is that the screen reader is not required to support ARIA, it’s simply reading text on the page that sighted users cannot see. While most screen readers support ARIA, you may not want to rely on it if you don’t have to."

In the case you mention, the 'sr-only' text is actually better, as sometimes the aria-label text gets completely ignored on the screen reader (check the last example on that linked website).

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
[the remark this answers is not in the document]
Golodhros: Well, in reality I was putting the empty alt to just be compliant. I don't think a visual impaired user would even open the modal, as it is a preview image of the dashboard.

amundsen_application/static/js/components/DashboardPage/index.tsx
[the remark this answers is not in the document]
Golodhros: This recommendation comes from the Axe extension. It is not a great deal if we cannot have all the sections, but if we can, we can easily do it!

amundsen_application/static/js/components/Feedback/FeedbackForm/RatingFeedbackForm/index.tsx
Golodhros: Awesome!

amundsen_application/static/js/components/TableDetail/WatermarkLabel/index.tsx
Golodhros: Oh, good point.  
I could fix it by setting a fixed width on the 'date-range-label', but that won't be flexible if we end up allowing translations.

Do you want me to fix it?

Golodhros:
> Haven't completed reviewing all the files yet but I too ended up with many questions about the patterns in order to review without leaving the same kinds of comments everywhere, so let's start with these and may take a couple passes.
> 
> I'd like to know more about certain decisions and whether or not some of these changes were intended to remove warnings vs. making everything accessible (e.g. empty alt attributes?)
> 
> I also think it would be good to document what rules we want to adhere to. Moving forward it's reasonable that once we resolve this tech debt, reviewers can be expected to ask for corrections on PRs but not everyone who has worked on/will ever work on this repo has these kinds of things in mind or is even aware of some of these concepts. So I think at minimum we could find of kind of basic checklist of these concepts and link somewhere in our doc (perhaps developer guide?) For example maybe [this](http://web-accessibility.carnegiemuseums.org/foundations/semantic/) or any other resource you think gives the best overview.

That sounds good!
I can add something on the developer guide about using those recommendations and running the Axe extension to make sure these rules are kept. We can also add more a11y rules on ESLint to enforce them!

amundsen_application/static/js/components/AnnouncementPage/index.tsx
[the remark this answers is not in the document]
ttannis: I had misinterpreted that line as a recommendation to only have one `h1` if you have them, not that there **must** be an `h1`. However, I found a better resource that covers the purpose and explains why it is a must. https://fae.disability.illinois.edu/rulesets/HEADING_1/. I'm all set on this concept.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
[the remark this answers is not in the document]
ttannis: I see, I'll look at each of these on a case by case basis to see if there are any where we probably should provide some text.

amundsen_application/static/js/components/DashboardPage/ImagePreview/index.tsx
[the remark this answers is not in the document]
ttannis: I see, I actually didn't know aria support had gaps. Also found this: https://developer.paciellogroup.com/blog/2018/07/short-note-on-aria-support/

> ARIA is not tested for screen reader support before the ARIA specification is published....

> The catch is that ARIA is tested for support in accessibility APIs, but not in screen readers...

😲

amundsen_application/static/js/components/TableDetail/ReportTableIssue/index.spec.tsx
```diff
@@ -92,9 +92,9 @@ describe('ReportTableIssue', () => {
         const { wrapper } = setup();
         const previsOpenState = wrapper.state().isOpen;
 
-        wrapper
-          .instance()
-          .toggle({ currentTarget: { id: 'id', nodeName: 'button' } });
+        wrapper.instance().toggle({
```
ttannis, line 95: Notice the test was updated but not the component. No issue though but as mentioned looks like we need to go through and formally update to using `dataset.type` for logging the location of a particular interaction we're interested in.

amundsen_application/static/js/components/TableDetail/WatermarkLabel/index.tsx
ttannis: Yes let's fix it for now to avoid making a design change just because we wanted to clean stuff up. Then if/when we allow translations we can deal with that issue by asking for an update design that considers this.

ttannis:
> > Haven't completed reviewing all the files yet but I too ended up with many questions about the patterns in order to review without leaving the same kinds of comments everywhere, so let's start with these and may take a couple passes.
> > I'd like to know more about certain decisions and whether or not some of these changes were intended to remove warnings vs. making everything accessible (e.g. empty alt attributes?)
> > I also think it would be good to document what rules we want to adhere to. Moving forward it's reasonable that once we resolve this tech debt, reviewers can be expected to ask for corrections on PRs but not everyone who has worked on/will ever work on this repo has these kinds of things in mind or is even aware of some of these concepts. So I think at minimum we could find of kind of basic checklist of these concepts and link somewhere in our doc (perhaps developer guide?) For example maybe [this](http://web-accessibility.carnegiemuseums.org/foundations/semantic/) or any other resource you think gives the best overview.
> 
> That sounds good!
> I can add something on the developer guide about using those recommendations and running the Axe extension to make sure these rules are kept. We can also add more a11y rules on ESLint to enforce them!

Yes please. Anything that helps contributors independently get most of the way on their own and understand basic concepts we will adhere to. It will take the burden off a code reviewer to catch everything.

amundsen_application/static/js/components/TableDetail/WatermarkLabel/index.tsx
Golodhros: OK, fixing it.

amundsen_application/static/js/components/TableDetail/ReportTableIssue/index.spec.tsx
Golodhros: Not sure I understand, I updated 'TableDetail/FrequentUsers/index.tsx' with a data attribute.

amundsen_application/static/js/components/common/Bookmark/BookmarkIcon/index.tsx
```diff
@@ -62,6 +62,7 @@ export class BookmarkIcon extends React.Component<BookmarkIconProps> {
             'icon ' +
             (this.props.isBookmarked ? 'icon-bookmark-filled' : 'icon-bookmark')
           }
+          alt=""
```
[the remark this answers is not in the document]
Golodhros: Yes, it should definitely be a button. We have more instances of this case, but it hasn't been detected by our ESLint rules.

I will create a ticket for this, I think it would make sense to move the star to use SVGs instead at the same time that we change it into a button.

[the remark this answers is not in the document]
Golodhros: This ticket: https://jira.lyft.net/browse/AMD-1513

amundsen_application/static/js/components/common/EditableSection/index.tsx
```diff
@@ -114,11 +118,15 @@ export class EditableSection extends React.Component<
...
-        {childrenWithProps}
+        <label className="editable-section-label">
+          <div className="editable-section-label-wrapper">
+            <span className="section-title title-3">
+              {EditableSection.convertText(title)}
+            </span>
+            {!readOnly ? this.renderButton() : this.renderReadOnlyButton()}
+          </div>
+          <div className="editable-section-content">{childrenWithProps}</div>
+        </label>
```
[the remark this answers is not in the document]
Golodhros: This feature is something that a lot of people don't know.
Since the beginning, the labels were supposed to wrap the inputs they label, and only when they don't is when we use the "for" attribute. Somehow this knowledge got lost in the last years of FE development, but it was the original idea.

In this case, as the inputs of this could be different, it made sense to make the label wrap the input so that we don't need to keep a dynamic id on the inputs and add it on the label as the "for" attribute value. I had to add some extra markup to keep the right styling, but it is worth to keep on having this dynamic editable section.

amundsen_application/static/js/components/TableDetail/ReportTableIssue/index.spec.tsx
ttannis: And this is the `TableDetail/ReportTableIssue` test. :)

amundsen_application/static/js/ducks/utilMethods.ts
```diff
@@ -40,7 +40,7 @@ export function logClick(
   const target = event.currentTarget;
   const inferredProps: ActionLogParams = {
     command: 'click',
-    target_id: target.id,
+    target_id: target.dataset.type,
```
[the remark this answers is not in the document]
Golodhros: Good call, changing it.

amundsen_application/static/js/components/TableDetail/ReportTableIssue/index.spec.tsx
Golodhros: Ohhhh.... bringing this back, thanks!!

amundsen_application/static/js/components/TableDetail/WatermarkLabel/styles.scss
```diff
@@ -8,9 +10,24 @@
     height: 40px;
     margin-right: 8px;
     width: 12px;
+    margin-top: 2px;
```
ttannis, line 13: Mind not having this margin?

amundsen_application/static/js/components/TableDetail/WatermarkLabel/styles.scss
```diff
@@ -8,9 +10,24 @@
...
   }
 
   .range-labels {
     flex-basis: 42px;
   }
+
+  .date-range {
+    margin-top: 0;
+    margin-bottom: 6px;
```
ttannis, line 22: Mind not having this margin or `margin-bottom: 0`

amundsen_application/static/js/components/common/EditableSection/styles.scss
```diff
@@ -1,13 +1,23 @@
 @import 'variables';
 
 .editable-section {
+  .editable-section-label {
+    display: block;
+    font-weight: 400;
```
ttannis, line 6: Mind having `margin-bottom: 0`?

amundsen_application/static/js/components/common/EditableSection/index.tsx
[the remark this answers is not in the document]
ttannis: This makes sense.

amundsen_application/static/js/ducks/utilMethods.ts
```diff
@@ -40,7 +40,7 @@ export function logClick(
   const target = event.currentTarget;
   const inferredProps: ActionLogParams = {
     command: 'click',
-    target_id: target.id,
+    target_id: target.dataset ? target.dataset.type : target.id,
```
ttannis, line 43: Oh also we must check `(target.dataset && target.dataset.type)`, I just tested a case that uses `id` and `target.dataset` is an empty object, so this got set to `undefined`.

amundsen_application/static/js/components/TableDetail/WatermarkLabel/styles.scss
Golodhros: It is important to have it properly aligned, you don't agree?

amundsen_application/static/js/components/TableDetail/WatermarkLabel/styles.scss
Golodhros: That would feel really tight. 
I can bring it down to 4px if you want.

amundsen_application/static/js/components/common/EditableSection/styles.scss
Golodhros: Sure.

amundsen_application/static/js/ducks/utilMethods.ts
Golodhros: OK.

Where is it that we are still using ids? It might be better to just change those.

ttannis: It would be but I didn't want this PR to increase in scope, especially for changes in logging. I think that's more appropriate for a dedicated commit that gets explicitly shepherded to production and verified. Any issues or bugs risk us just losing that user event data for however long the issues existed in production.

ttannis: Comment may have gotten lost in all the comments here, but would like to file a ticket for this to use `dataset.type` instead of `id` for logging. I myself forget all the places without going through and looking for each of them but if you search for places where we call `logClick`, every component that triggers the event where we call `logClick` is where we'd make this change.

amundsen_application/static/js/components/TableDetail/WatermarkLabel/styles.scss
ttannis: I guess our eyeballs just have a different perspective on what "alignment" looks like. But I can see how you find this helps better align. I'm fine with this.

amundsen_application/static/js/components/TableDetail/WatermarkLabel/styles.scss
ttannis: I can see that. `4px` would be good.

Golodhros: done!

amundsen_application/static/js/ducks/utilMethods.ts
Golodhros: OK, I will file a ticket!

PR #481, opened 2020-06-12
feat: Mode dashboard preview ACL

jinhyukchang:
### Summary of Changes

Added `access control` on Mode dashboard preview.

- Calls `metadata service` is verify if user is `is_active` and has `mode_user_id`. Because, all Mode dashboard is in shared space, this two conditions are enough to validate the access.
- If user is not registered in `Mode` (does not have permission to view the report), it returns `401` Unauthorized.

### Tests

Unit tests updated.
Integration test done.

### Documentation

docstring

### CheckList

### DIFF
--- a/amundsen_application/dashboard_preview/mode_preview.py
+++ b/amundsen_application/dashboard_preview/mode_preview.py
@@ -6,7 +6,10 @@ from flask import has_app_context, current_app as app
 from requests.auth import HTTPBasicAuth
 from retrying import retry
 
+from amundsen_application.api.metadata.v0 import USER_ENDPOINT
+from amundsen_application.api.utils.request_utils import request_metadata
 from amundsen_application.base.base_preview import BasePreview
+from amundsen_application.models.user import load_user
 
 LOGGER = logging.getLogger(__name__)
 DEFAULT_REPORT_URL_TEMPLATE = 'https://app.mode.com/api/{organization}/reports/{dashboard_id}'
@@ -16,6 +19,10 @@ def _validate_not_none(var: Any, var_name: str) -> Any:
     if not var:
         raise ValueError('{} is missing'.format(var_name))
     return var
+
+
+def _retry_on_retriable_error(exception: Exception) -> bool:
+    return not isinstance(exception, PermissionError)
 
 
 class ModePreview(BasePreview):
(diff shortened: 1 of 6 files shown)


amundsen_application/dashboard_preview/mode_preview.py
```diff
@@ -40,13 +47,26 @@ def __init__(self, *,
...
     def get_preview_image(self, *, uri: str) -> bytes:
         """
         Retrieves short lived URL that provides Mode report preview, downloads it and returns it's bytes
         :param uri:
         :return: image bytes
+        :raise: PermissionError when user is not allowed to access the dashboard
         """
+
+        if self._is_auth_enabled:
+            self._authorize_access(user_email=self._auth_user_method(app).email)
```
feng-tao, line 68: the code lgtm. But it only works for the case when we use email as user id as I believe we allows both email and user_id for primary key for user model.

tests/unit/dashboard_preview/test_mode_preview.py
```diff
@@ -56,3 +68,138 @@ def test_get_preview_image_url_failure_missing_key(self) -> None:
...
+                'first_name': 'test_first_name',
+                'team_name': 'test_team',
+                'email': 'test_email',
+            }
+
+            preview = ModePreview(access_token='token', password='password', organization='foo')
+            self.assertRaises(PermissionError, preview._authorize_access, user_email='test_email')
+
+
+if __name__ == '__main__':
```
feng-tao, line 204: do you need this in order to get the unittests called?

amundsen_application/dashboard_preview/mode_preview.py
jinhyukchang: > the code lgtm. But it only works for the case when we use email as user id as I believe we allows both email and user_id for primary key for user model.

Agreed. We probably could refactor this method to add user_id `attribute` later on.

tests/unit/dashboard_preview/test_mode_preview.py
jinhyukchang: > do you need this in order to get the unittests called?

No, unit test itself will be executed vis pytest. But I found it that IDE that I am using - Pycharm - is not doing well on `debugging` when executing as `pytest`.

amundsen_application/dashboard_preview/mode_preview.py
ttannis: So will we have a follow-up on our end to populate `user_id` in our user metadata and use it here instead?

feng-tao: For us, it is fine. We populate user_id with https://github.com/lyft/amundsencommon/blob/master/amundsen_common/models/user.py#L59 which will be the same as email. But it won't work for folks if they don't share the same user_id with email.

ttannis: 1. Do we have that working? Because when I check my user information my `user_id` is null.
2. If we do have `user_id` then, how come we wouldn't use it right at this moment?

jinhyukchang: @feng-tao Same question as Tamika. Do you know if this is currently working? I assumed that this requires additional effort.

ttannis: Okay, so what would be the list of next steps (for internal & external repos) for being able to switch this to `user_id`? It would be good to have the plan and file a ticket since it sounds like we have to update eventually in order to make this code work for others who request users by explicit `user_id`.

jinhyukchang: @feng-tao @ttannis Confirmed that `user_id` is already working. Posted a PR.
https://github.com/lyft/amundsenfrontendlibrary/pull/485

PR #483, opened 2020-06-15
feat: Adds Shimmer loader to the Table Issues and Footer

Golodhros:
### Summary of Changes
Adds a Shimmer loader on the Table Issues and Footer

### Tests
Updates tests

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/Footer/index.tsx
+++ b/amundsen_application/static/js/components/Footer/index.tsx
@@ -20,6 +20,14 @@ interface DispatchFromProps {
 
 export type FooterProps = StateFromProps & DispatchFromProps;
 
+const ShimmeringFooterLoader: React.FC = () => {
+  return (
+    <div className="shimmer-footer">
+      <div className="shimmer-footer-row is-shimmer-animated"></div>
+    </div>
+  );
+};
+
 export class Footer extends React.Component<FooterProps> {
   componentDidMount() {
     this.props.getLastIndexed();
@@ -30,12 +38,14 @@ export class Footer extends React.Component<FooterProps> {
   };
 
   render() {
-    let content;
+    let content = (<ShimmeringFooterLoader />);
+
     if (this.props.lastIndexed) {
       content = (
         <div>{`Amundsen was last indexed on ${this.generateDateTimeString()}`}</div>
       );
     }
+
     return (
       <footer>
         <div className="phantom-div" />
--- a/amundsen_application/static/js/components/Footer/index.tsx
+++ b/amundsen_application/static/js/components/Footer/index.tsx
@@ -23,7 +23,7 @@ export type FooterProps = StateFromProps & DispatchFromProps;
 const ShimmeringFooterLoader: React.FC = () => {
   return (
     <div className="shimmer-footer">
-      <div className="shimmer-footer-row is-shimmer-animated"></div>
+      <div className="shimmer-footer-row is-shimmer-animated" />
     </div>
   );
 };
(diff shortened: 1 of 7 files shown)


amundsen_application/static/package.json
```diff
@@ -514,7 +514,7 @@
     },
     "extends": "stylelint-config-standard"
   },
-  "husky": {
+  "__husky": {
```
Golodhros, line 517: Needs a fix added on https://github.com/lyft/amundsenfrontendlibrary/pull/480 to work properly

amundsen_application/static/js/components/Footer/index.tsx
```diff
@@ -30,12 +38,14 @@ export class Footer extends React.Component<FooterProps> {
   };
 
   render() {
-    let content;
+    let content = (<ShimmeringFooterLoader />);
```
ttannis, line 41: I actually don't think we want to show the loader if the value doesn't exist, we originally show nothing so that things look less "broken". I forgot that this component doesn't have a loading state, so perhaps it wasn't a good candidate for this yet. 

**But**, I think we could keep it and some future cleanup task would be to allow `Footer` to know when things are loading. So that the UI will show the loader and then either show the date string or no string -- or just not even render the Footer probably. Any opinions?

amundsen_application/static/package.json
ttannis: What's the context?

amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
```diff
@@ -1,5 +1,9 @@
 @import 'variables';
 
+$issue-banner-height: 40px;
+$shimmer-loader-height: 90px;
```
ttannis, line 4: nit nit: what do you think about increasing this to `120px` so that the shimmer loader lines will take up the same amount of space as the banner? I don't have some official methodology behind this, just a preference that sizing the shimmer items to the content improves that "feel" of content popping in nicely when ready.

amundsen_application/static/js/components/Footer/index.tsx
Golodhros: Oh, I see. I thought we would always have a lastIndexed. In that case I agree, we shouldn't show the loader if there won't be any value.

And yes, we probably won't want to load the footer, but that would require some extra tweaking. I am not sure if Knowl would actually asks us to remove the footer overall.

Not sure how to proceed at this moment.

amundsen_application/static/package.json
Golodhros: Some conflicts between ESLint and Prettier about whitespace on the TableIssues file.

amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
Golodhros: Yeah, I had some back and forth with Knowl about this. At the end we settled on not using this fixed size and use the fixed size of each line. I will update this code now.

amundsen_application/static/js/components/Footer/index.tsx
ttannis: We will always have it when things work as expected at Lyft, but this is just an edge case for an issue that can happen in the backend, so we just make the UI account for it as gracefully as possible. 

I'm okay with leaving this code as is though -- perhaps this is one of the cases where its not a terrible thing that something looks broken. Because if it actually is broken (in the backend) this hints that something isn't right.

amundsen_application/static/js/components/TableDetail/TableIssues/styles.scss
ttannis: Fine with me, based on whatever Knowl recommended.

Golodhros: We ended up having only two lines here.

PR #486, opened 2020-06-16
feat: Update TagList Loading State with a shimmer loader

Golodhros:
### Summary of Changes
Adds a TagList shimmer loader component
Adds it to our TagList component

### Tests
Updated tests

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/common/ShimmeringTagListLoader/index.tsx
+++ b/amundsen_application/static/js/components/common/ShimmeringTagListLoader/index.tsx
@@ -1,0 +1,38 @@
+import * as React from 'react';
+import * as times from 'lodash/times';
+
+import './styles.scss';
+
+const DEFAULT_REPETITION = 10;
+
+type ShimmeringTagItemProps = {
+  index: number;
+};
+
+export const ShimmeringTagItem: React.SFC<ShimmeringTagItemProps> = ({
+  index,
+}: ShimmeringTagItemProps) => {
+  return (
+    <span
+      className={`shimmer-tag-loader-item shimmer-tag-loader-item--${index} is-shimmer-animated`}
+    />
+  );
+};
+
+export interface ShimmeringTagListLoaderProps {
+  numItems?: number; // Max items would be 15
+}
+
+const ShimmeringTagListLoader: React.SFC<ShimmeringTagListLoaderProps> = ({
+  numItems = DEFAULT_REPETITION,
+}: ShimmeringTagListLoaderProps) => {
+  return (
+    <div className="shimmer-tag-list-loader">
+      {times(numItems, (idx) => (
+        <ShimmeringTagItem key={idx} index={idx} />
+      ))}
+    </div>
+  );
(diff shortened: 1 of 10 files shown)


amundsen_application/static/js/components/common/ShimmeringTagListLoader/styles.scss
```diff
@@ -0,0 +1,29 @@
+@import 'variables';
+
+$shimmer-loader-tag-height: 36px;
+$shimmer-loader-tag-border-radius: 4px;
```
ttannis, line 4: nit: I assume these two values were chosen to match `TagInfo`, so we perhaps we could add some shared tag-info variables in `_variables-default` and use them both here and in `TagInfo` for easier update if we ever change the size.

amundsen_application/static/js/components/common/ShimmeringTagListLoader/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+}: ShimmeringTagItemProps) => {
+  return (
+    <span
+      className={`shimmer-tag-loader-item shimmer-tag-loader-item--${index} is-shimmer-animated`}
+    />
+  );
+};
+
+export interface ShimmeringTagListLoaderProps {
+  numItems?: number; // Max items would be 15
```
ttannis, line 23: Is max items concept from design? If so we could also enforce that in the code with `Math.max()`.

amundsen_application/static/js/components/common/ShimmeringTagListLoader/styles.scss
Golodhros: Yeah, makes sense!
I will check those up, thanks!

amundsen_application/static/js/components/common/ShimmeringTagListLoader/index.tsx
Golodhros: I actually tried to enforce it, but didn't know the best way to go about it.
Throwing an error didn't work as expected, any ideas of how to approach it?

amundsen_application/static/js/components/common/ShimmeringTagListLoader/styles.scss
Golodhros: I was only able to reuse the border radius value.

amundsen_application/static/js/components/common/ShimmeringTagListLoader/index.tsx
ttannis: By "enforce" I don't think we need to go as far as making it such that the app throws an error but
1. What happens if we try `numItems?: 1 | 2 | 3 | ... | 15`?
2. If that doesn't do anything at best my original idea was to enforce a maximum of 15 in the render logic.

Golodhros: Oh, that makes sense!
Thanks!

PR #488, opened 2020-06-17
feat: Uses button for "clear" on filters

Golodhros:
### Summary of Changes
Adds a style for .btn-link
Uses that style on the filters section
Removes nesting on the filters section styling

rides_-_Amundsen_Search

### Tests
Updated tests

### CheckList

### DIFF
--- a/amundsen_application/static/css/_buttons-default.scss
+++ b/amundsen_application/static/css/_buttons-default.scss
@@ -144,6 +144,22 @@
     }
   }
 
+  &.btn-link {
+    color: $link-color;
+    text-decoration: none;
+    padding: 0 4px 0 4px;
+
+    &:hover,
+    &:focus {
+      color: $link-hover-color;
+    }
+
+    &:focus {
+      outline: 5px auto $link-hover-color;
+      outline-offset: -2px;
+    }
+  }
+
   &.disabled,
   &:disabled {
     -webkit-box-shadow: none;
--- a/amundsen_application/static/js/components/SearchPage/SearchFilter/styles.scss
+++ b/amundsen_application/static/js/components/SearchPage/SearchFilter/styles.scss
@@ -7,19 +7,19 @@
 
   .search-filter-section-header {
     display: flex;
+  }
 
-    .search-filter-section-title {
-      display: flex;
-      flex-grow: 1;
+  .search-filter-section-title {
+    display: flex;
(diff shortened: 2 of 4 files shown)


amundsen_application/static/js/components/SearchPage/SearchFilter/styles.scss
```diff
@@ -7,19 +7,19 @@
 
   .search-filter-section-header {
```
ttannis, line 8: Why would we un-nest styles? Once upon a time we intentionally moved towards nesting styles to decrease collisions given that we don't have anything else set up right now to assist with that, and have ended up with a few bugs because of that.

Golodhros: Yeah, some nesting is OK, and I personally keep it at two levels max.
The Sass guidelines suggest to limit it to just one.

The problem with nesting is that generates styles with really high specificity, so when you need to override them, you will be forced to use "!important" (as I just had to do here).

If you want, I can help on moving us into BEM(https://9elements.com/bem-cheat-sheet/), it is relatively easy and eliminates all collisions at once.

ttannis: I see. I don't mind this level of nesting. Though if the issue is that it creates problems when overriding then that makes a case for how we approach the styles for items in `common` that are meant to be reuseable, and wouldn't particularly invest the time into these highly specific components. 

In regards to the posted link, I won't be able to form an opinion until we do some kind of semi-formal evaluation of the applications needs and weigh them against other approaches regarding css collisions.

amundsen_application/static/css/_buttons-default.scss
```diff
@@ -144,6 +144,22 @@
     }
   }
 
+  &.btn-link {
+    color: $link-color;
```
ttannis, line 148: I know we have some inconsistent concepts in the application regarding the styles of links vs clickable text, but I didn't think the goal here was to change the color. From the recommendation to just remove the carat, I assumed we'd still be using the usual text color with indigo on highlight.

Golodhros: Yep, this was something I chat with Knowl. The current color is a grey that don't really make sense as an interactive element.

This color declaration could probably be gone, but I left it to be explicit.

PR #493, opened 2020-06-22
fix: Update user icon on ProfilePage

ttannis:
### Summary of Changes

Recent style changes caused a UI regression where the hover state over the user icon on the ProfilePage was not correct, showing an indigo background.
**Before**
image

To fix this issue I reverted to using the previous mask image styling for this use case.
**After**
image

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/css/_icons.scss
+++ b/amundsen_application/static/css/_icons.scss
@@ -183,6 +183,11 @@ img.icon {
     mask-image: url('/static/images/icons/Up.svg');
   }
 
+  &.icon-user {
+    -webkit-mask-image: url('/static/images/icons/users.svg');
+    mask-image: url('/static/images/icons/users.svg');
+  }
+
   &.icon-more {
     -webkit-mask-image: url('/static/images/icons/More.svg');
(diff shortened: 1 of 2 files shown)


ttannis:
cc @dikshathakur3119 for context and if you have any questions.

amundsen_application/static/css/_icons.scss
```diff
@@ -183,6 +183,11 @@ img.icon {
     mask-image: url('/static/images/icons/Up.svg');
   }
 
+  &.icon-user {
+    -webkit-mask-image: url('/static/images/icons/users.svg');
+    mask-image: url('/static/images/icons/users.svg');
```
feng-tao, line 188: what is the difference between webkit-mask-image vs mask-image ?

ttannis: Different browsers have small differences/complications with how they handle css styles, this is why sometimes UIs run into issues with some things working in one browser and not working in another. 

For the `mask-image` style, Chrome for whatever reason implemented support for this only when prefixed with `-webkit-`, so we need this option to work in Chrome.

PR #494, opened 2020-06-23
fix: Separate related dashboards component & request

ttannis:
### Summary of Changes

This PR fixes the issue of the `TableDetail` page being unable to render when the metadata endpoint for related dashboards is not properly implemented or otherwise has an issue and returns a non-200 code. The cause of the issue was coupling the request for table metadata with the request for related dashboards -- if the related dashboard request failed then the entire page failed to load. 

This PR separate the logic of the related dashboards feature into a separate component and decouples some API calls. Support also includes showing error text from the response when there is an issue with the related dashboard request. 

Further details implementation details have been included inline.

### Tests

Tests updated to reflect changes.

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/api/metadata/v0.py
+++ b/amundsen_application/api/metadata/v0.py
@@ -674,7 +674,7 @@ def _get_related_dashboards_metadata(*, url: str) -> Dict[str, Any]:
     results_dict['status_code'] = status_code
 
     if status_code != HTTPStatus.OK:
-        message = response.raise_for_status()
+        message = f'Encountered {status_code} Error: Related dashboard metadata request failed'
         results_dict['msg'] = message
         logging.error(message)
         return results_dict
--- a/amundsen_application/api/metadata/v0.py
+++ b/amundsen_application/api/metadata/v0.py
@@ -656,7 +656,7 @@ def get_related_dashboard_metadata(table_key: str) -> Response:
 def _get_related_dashboards_metadata(*, url: str) -> Dict[str, Any]:
 
     results_dict = {
-        'dashboards': {},
+        'dashboards': [],
         'msg': '',
     }
 
@@ -674,7 +674,7 @@ def _get_related_dashboards_metadata(*, url: str) -> Dict[str, Any]:
     results_dict['status_code'] = status_code
 
     if status_code != HTTPStatus.OK:
-        message = 'Encountered error: Related Dashboard Metadata request failed'
+        message = response.raise_for_status()
         results_dict['msg'] = message
         logging.error(message)
         return results_dict
--- a/amundsen_application/static/js/components/TableDetail/index.tsx
+++ b/amundsen_application/static/js/components/TableDetail/index.tsx
@@ -18,13 +18,13 @@ import LoadingSpinner from 'components/common/LoadingSpinner';
 import Flag from 'components/common/Flag';
 
 import ColumnList from 'components/TableDetail/ColumnList';
-import TableDashboardResourceList from 'components/TableDetail/TableDashboardResourceList';
 import DataPreviewButton from 'components/TableDetail/DataPreviewButton';
 import ExploreButton from 'components/TableDetail/ExploreButton';
 import FrequentUsers from 'components/TableDetail/FrequentUsers';
 import LineageLink from 'components/TableDetail/LineageLink';
 import OwnerEditor from 'components/TableDetail/OwnerEditor';
 import SourceLink from 'components/TableDetail/SourceLink';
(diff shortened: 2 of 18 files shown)


amundsen_application/api/metadata/v0.py
```diff
@@ -674,7 +674,7 @@ def _get_related_dashboards_metadata(*, url: str) -> Dict[str, Any]:
     results_dict['status_code'] = status_code
 
     if status_code != HTTPStatus.OK:
-        message = 'Encountered error: Related Dashboard Metadata request failed'
+        message = f'Encountered {status_code} Error: Related dashboard metadata request failed'
```
ttannis, line 677: This message will show in the UI if there is a failure from the metadata service.
**Example:**
image

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -138,18 +141,25 @@ export class TableDetail extends React.Component<
...
-          allItems={this.props.tableData.dashboards}
-          itemsPerPage={DASHBOARDS_PER_PAGE}
-          source={TABLE_SOURCE}
-        />
-      ),
-      key: 'dashboards',
-      title: `Dashboards (${this.props.tableData.dashboards.length})`,
-    });
+    if (indexDashboardsEnabled()) {
+      const loadingTitle = (
```
ttannis, line 145: This logic supports:
1. Not showing this tab at all if dashboards are not enabled.
2. Indicating the loading state then showing the total:
dashboard_tab_loading

amundsen_application/static/js/components/common/ResourceList/styles.scss
```diff
@@ -23,7 +23,8 @@
   .empty-message {
     display: flex;
     justify-content: center;
-    margin-top: $spacer-4;
+    margin: $spacer-4 $spacer-4 0;
```
ttannis, line 26: Cleanup: These styles just make the error text look nicer if it gets really long on a screen that isn't very wide.

amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
```diff
@@ -34,37 +34,53 @@ export type DescriptionAPI = { description: string } & MessageAPI;
 export type LastIndexedAPI = { timestamp: string } & MessageAPI;
 export type PreviewDataAPI = { previewData: PreviewData } & MessageAPI;
 export type TableDataAPI = { tableData: TableData } & MessageAPI;
-export type RelatedDashboardDataAPI = { dashboards: DashboardResource[] };
+export type RelatedDashboardDataAPI = {
+  dashboards: DashboardResource[];
+} & MessageAPI;
 
 export function getTableData(
```
ttannis, line 41: We reverted to having a `getTableData` call that just gets table metadata, and now also have a `getTableDashboards` call.

amundsen_application/static/js/ducks/tableMetadata/reducer.ts
```diff
@@ -227,6 +243,11 @@ export function getPreviewDataSuccess(
 
 /* REDUCER */
 export interface TableMetadataReducerState {
+  dashboards?: {
```
ttannis, line 246: Related dashboard information is optionally stored as its own object on the table metadata reducer state.

amundsen_application/static/js/ducks/tableMetadata/sagas.ts
```diff
@@ -34,15 +35,22 @@ import {
...
-    const { key, searchIndex, source } = action.payload;
     const { data, owners, statusCode, tags } = yield call(
       API.getTableData,
       key,
       searchIndex,
       source
     );
     yield put(getTableDataSuccess(data, owners, statusCode, tags));
+
+    try {
```
ttannis, line 48: If we succeed in getting the table metadata, this will now fetch the related dashboards.

ttannis:
cc @dikshathakur3119 @verdan for questions and comments

feng-tao:
hey @ttannis , please tag me as well for FE pr review in the future.

amundsen_application/api/metadata/v0.py
feng-tao: curious: do you use the open source version to mimic the issue or just override it with FE API endpoint?

ttannis: I updated the code in Flask to mimic returning an error instead of the correct response -- just like I would use it to return dummy data as well if there's some specific case we have to test but don't have the real data in our environment for.

PR #497, opened 2020-06-26
fix: Making description uneditable using config file

dikshathakur3119:
### Summary of Changes

Change to make Table and column description uneditable for given schemas and some table following specific pattern.

GitHub Issue:
https://github.com/lyft/amundsen/issues/507

### Tests
Have to add tests

### Documentation

NA

### CheckList


amundsen_application/static/js/components/common/EditableText/index.tsx
```diff
@@ -144,7 +144,7 @@ class EditableText extends React.Component<
           maxLength={this.props.maxLength}
           ref={this.textAreaRef}
           defaultValue={this.state.value}
-          disabled={this.state.isDisabled}
+          disabled={!this.props.editable}
```
ttannis, line 147: As discussed offline, the disabled feature isn't related to whether or not this is editable...it blocks interaction with the component for a use case where someone tries to edit an out of date description. (See logic in `componentDidUpdate` if curious.

For this fix, we would start on the `TableDetail` page. Notice that the text component is wrapped in an `EditableSection` component. `EditableSection` will control whether or not its children are editable with its `readyOnly` property. Let's look into a fix starting there.

ttannis: You'll also want to look into `EditableSection`'s properties to help with the task of allowing the edit button to take users to the source of truth where they can edit (if it exists)

dikshathakur3119:
👀 @ttannis @feng-tao

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -42,7 +43,17 @@ def marshall_table_full(table_dict: Dict) -> Dict:
     table: Table = schema.load(table_dict).data
     results: Dict[str, Any] = schema.dump(table).data
 
-    is_editable = results['schema'] not in app.config['UNEDITABLE_SCHEMAS']
+    # Check if schecma is uneditable
+    is_editable_schema = results['schema'] not in app.config['UNEDITABLE_SCHEMAS']
+
+    # Check if Table is uneditable
+    table_id = results['schema'] + "." + results['name']
```
ttannis, line 50: I have concerns about applying the regex to the pattern of `schema.name`. This is a pattern that we have hardcoded in certain places because this is how our current design creates a display name for the tables. The concern is that this pattern would not be applicable for others in open source who aren't using the `schema.name` pattern.

However, the proposal to add `is_editable` to the resource metadata itself as our long term approach would address this concern since it would deprecate this logic. So I'm raising this not to request changes, but to acknowledge another benefit of moving to the long term approach as soon as we are able to.

amundsen_application/config.py
```diff
@@ -21,6 +21,7 @@ class Config:
     COLUMN_STAT_ORDER = None  # type: Dict[str, int]
 
     UNEDITABLE_SCHEMAS = set()  # type: Set[str]
+    UNEDITABLE_TABLES_REGEX = r"$^"
```
ttannis, line 24: nit: Can add `type`

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -165,6 +165,8 @@ export class TableDetail extends React.Component<
       innerContent = <ErrorMessage />;
     } else {
       const data = tableData;
+      const editText = data.source ? data.source.source_type : null;
```
ttannis, line 168: 1. Let's have a more helpful piece of text like we do for dashboards:
image

2. `editText` and `editUrl` are typed as options strings, so let's use empty strings instead of `null` for the default. If we ever enable a stricter type checking option in the Typescript code,  these lines would end up throwing errors

amundsen_application/api/utils/metadata_utils.py
dikshathakur3119: @ttannis @feng-tao , how should I proceed with this? One way I can think of eliminating this company specific `table_id` from front-end is by adding three variables in config files : 
1 .one variable is table_metadata attributes which we want to be part of our matching criteria (table_id) e.g `table_id_constituents = ['schema' , 'name', 'source', 'last_updated_timestamp']`
2. Another variable in config which will define operators to combine above defined table_id_constituents, e.g `table_id_operators = ['.', '_', '.']`
Length of `table_id_operators` array will be one less than `table_id_constituents` array. 
3. Third variable would be regex string with which want to match our `table_id`.
It is not good solution as backend one but let me know your thoughts if I should try something like this, or is it better to start from backend itself.

dikshathakur3119:
> Some small comments. With this PR, we should also:
> 
> 1. Include some documentation about `UNEDITABLE_SCHEMAS` and `UNEDITABLE_TABLES_REGEX`. This would go somewhere in https://github.com/lyft/amundsenfrontendlibrary/blob/master/docs/flask_config.md, where we are trying to add at least a little information or quick explanation for each feature supported by `amundsen_application/config.py`
> 2. Add some unit tests to verify the additional logic in `metadata_utils` to verify that it sets `is_editable` accordingly.

@ttannis where should I add unit test cases since every company will have their own configurations for `UNEDITABLE_SCHEMAS` and `UNEDITABLE_TABLES_REGEX` . Should I write unit test cases in private repo??

amundsen_application/api/utils/metadata_utils.py
ttannis: I think to affectively address the use cases we might have to take a step back and do a new approach. I think we have to restructure the configuration so that users can have better control over what combinations of schemas and table_names they want to set as un-editable. The use cases are:
1. Set all tables with a given schema pattern as un-editable.
2. Set all tables with a specific table name  pattern in a given schema pattern as un-editable. This is the use case that creates the need for a new configuration structure, because users could want to apply different table patterns to different schemas.
3. Set all tables with a given table name pattern as editable.

In order to achieve the above via configurations, I'd propose switching to some kind of configuration where users would set a list of "match rules":
```
# MatchRuleObject could contain an optional regex for schemas and an option regex for table patterns

UNEDITABLE_TABLE_DESCRIPTION_MATCH_RULES: List[MatchRuleObject] = [
  # use case 1 - all tables in "schema" and "schema1" uneditable. 
  # An alternative to `UNEDITABLE_SCHEMAS`, but we could still keep that config for backwards compatibility 
  # until we switch to the long term solution
  { schema_regex: r”^(schema|schema1)" }, 
  
  # use case 2 - all "test_*" tables in "schema2" uneditable
  { schema_regex: r”^(schema2)", table_name_regex: r"^test_([a-zA-Z_0-9]+)" }, 
  
  # use case 2 - all "other_*" tables in "schema3" and "schema4" uneditable
  { schema_regex: r”^(schema3|schema4)", table_name_regex: r"^other_([a-zA-Z_0-9]+)" }, 
  
  # use case 3 - all "noedit_*" tables uneditable
  { table_name_regex: r"^noedit_([a-zA-Z_0-9]+)" }
]
```

What do you all think of this kind of configuration instead?

dikshathakur3119: This approach sounds good to me as far as we want to make decision only on basis of 'table name' and 'schema'. It is certainly better than fixing our pattern to 'schema.table_name', but it would not suffice if someone wants to make decision on the basis of other components of metadata like 'source'. We can move forward with this for the time till we move to long term solution.

ttannis: > but it would not suffice if someone wants to make decision on the basis of other components of metadata like 'source'

Yes. And that is okay for us because we acknowledged that this solution only needs to consider descriptions, beacause extended support to configure other editable structured metadata will deprecate these configurationsin favor of adding new attributes to the table metadata.

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -23,17 +25,43 @@ def marshall_table_partial(table_dict: Dict) -> Dict:
...
     results['type'] = 'table'
 
     return results
 
 
+def parse_editable_rule(rule: MatchRuleObject,
+                        schema: str,
+                        table: str) -> bool:
+    """
+    Matches table name and schema with corresponding regex in matching rule
```
ttannis, line 39: nit: Can you add descriptions for the parameters and return values like we do elsewhere? This may seem straightforward but understanding exactly what the parameters are or where they come from, or what the boolean returned by the method represents will provide useful context in the future for newcomers.

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -42,7 +70,17 @@ def marshall_table_full(table_dict: Dict) -> Dict:
     table: Table = schema.load(table_dict).data
     results: Dict[str, Any] = schema.dump(table).data
 
-    is_editable = results['schema'] not in app.config['UNEDITABLE_SCHEMAS']
+    # Check if schema is uneditable
+    is_editable_schema = results['schema'] not in app.config['UNEDITABLE_SCHEMAS']
+
+    # Check if Table Description is uneditable
+    # table_id = results['schema'] + "." + results['name']
```
ttannis, line 77: nit: delete line

amundsen_application/config.py
```diff
@@ -22,6 +31,28 @@ class Config:
 
     UNEDITABLE_SCHEMAS = set()  # type: Set[str]
 
+    # UNEDITABLE_TABLE_DESCRIPTION_MATCH_RULES is List of MatchRules
```
ttannis, line 34: I don't think we need these comments in the code and can clean this up. Setting `UNEDITABLE_TABLE_DESCRIPTION_MATCH_RULES = []  # type: List[MatchRuleObject]` give enough context about what this in.

Instead, this kind of information and examples would be written in the docs.

tests/unit/utils/test_metadata_utils.py
```diff
@@ -63,3 +65,47 @@ def test_sort_prog_descriptions_returns_default_value(self) -> None:
...
             not_in_config_value = {'source': 'test', 'text': 'I am a test'}
             self.assertEqual(_sort_prog_descriptions(mock_config, not_in_config_value), len(mock_config))
+
+
+class UneditableTableDescriptionTest(unittest.TestCase):
+    def setUp(self) -> None:
+        pass
+
+    def test_table_desc_match_rule_schema_only(self) -> None:
+        with local_app.app_context():
```
ttannis, line 75: Will these tests pass if you remove `with local_app.app_context():`? I believe we only have to include that when the code we are testing tries to access `app.config`.

tests/unit/utils/test_metadata_utils.py
```diff
@@ -63,3 +65,47 @@ def test_sort_prog_descriptions_returns_default_value(self) -> None:
...
+    def test_table_desc_match_rule_table_only(self) -> None:
+        with local_app.app_context():
+            # Mock match rule, table name and schema
+            test_match_rule = MatchRuleObject(table_name_regex=r"^noedit_([a-zA-Z_0-9]+)")
+
+            # assert result for given table name and match rule
+            self.assertEqual(_parse_editable_rule(test_match_rule, 'schema', 'noedit_test_table'), False)
+            self.assertEqual(_parse_editable_rule(test_match_rule, 'schema', 'editable_test_table'), True)
+
+    def test_table_desc_match_rule_multiple_schemas(self) -> None:
```
ttannis, line 92: nit: I'd say this test is not necessary. Our application logic doesn't place any logic on the regex itself, we just apply it and expect it to work. If we unit test the same exact logic with a different regex, we are essentially testing that the regex works as expected -- which we don't need to do because that is job of the `re` library.

tests/unit/utils/test_metadata_utils.py
dikshathakur3119: yeah you are right, we may not need this.

tests/unit/utils/test_metadata_utils.py
dikshathakur3119: yes I see your point. Thanks for pointing out!

dikshathakur3119:
👀 @ttannis @feng-tao

docs/flask_config.md
```diff
@@ -76,3 +76,50 @@ description sources not mentioned in the configuration will be alphabetically pl
 
 Here is a screenshot of what it would look like in the bottom left here:
 ![programmatic_description](img/programmatic_descriptions.png)
+
+##Uneditable Table Descriptions
+Amundsen supports configuring table and column description to be non-editable for selective tables. You may want to make table 
+descriptions non-editable due to various reasons such as table already has table description from source of truth.
+You can define matching rules in `config file` for selecting tables. This configuration is useful as table selection criteria can
```
ttannis, line 83: nit: `config.py`. When referencing files I have tried to either link the file directly or use the exact name people can use to find it via file search in their editors or github UI.

docs/flask_config.md
```diff
@@ -76,3 +76,50 @@ description sources not mentioned in the configuration will be alphabetically pl
...
+
+##Uneditable Table Descriptions
+Amundsen supports configuring table and column description to be non-editable for selective tables. You may want to make table 
+descriptions non-editable due to various reasons such as table already has table description from source of truth.
+You can define matching rules in `config file` for selecting tables. This configuration is useful as table selection criteria can 
+be company specific which will not directly integrated with Amundsen. 
+You can use different combinations of schema and table name for selecting tables.
+
+Here are some examples when this feature can be used: 
+1. You want to set all tables with a given schema pattern as un-editable.
```
ttannis, line 88: nit: `You want to set all tables with a given schema  or schema pattern as un-editable.`

docs/flask_config.md
```diff
@@ -76,3 +76,50 @@ description sources not mentioned in the configuration will be alphabetically pl
...
+2. You want to set all tables with a specific table name pattern in a given schema pattern as un-editable.
+3. You want to set all tables with a given table name pattern as un-editable.
+
+Amundsen has two variables in `config` file which can be used to define match rules:
```
ttannis, line 92: nit: config.py

docs/flask_config.md
```diff
@@ -76,3 +76,50 @@ description sources not mentioned in the configuration will be alphabetically pl
...
+If you want to restrict tables from a given schemas then you can use `UNEDITABLE_SCHEMAS` as follows:
+```python
+UNEDITABLE_SCHEMAS = set(['schema1', 'schema2'])
+```
+After above configuration, all tables in 'schema1' and 'schema2' will have non-editable table and column descriptions.
+
+If you have more complex matching rules you can use `UNEDITABLE_TABLE_DESCRIPTION_MATCH_RULES`. It provides you more flexibility 
+and control as you can create multiple match rules and use regex for matching schema nad table names.
+
+You can configure your match rules in `config file` as follow:
```
ttannis, line 108: nit: `config.py`

docs/flask_config.md
```diff
@@ -76,3 +76,50 @@ description sources not mentioned in the configuration will be alphabetically pl
...
+        # match rule for tables in schema4 with table name pattern 'noedit_*'
+        MatchRuleObject(schema_regex=r"^(schema4)", table_name_regex=r"^noedit_([a-zA-Z_0-9]+)"),
+        # match rule for tables in schema5, schema6 and schema7 with table name pattern 'noedit_*'
+        MatchRuleObject(schema_regex=r"^(schema5|schema6|schema7)", table_name_regex=r"^noedit_([a-zA-Z_0-9]+)"),
+        # match rule for all tables with table name pattern 'others_*'
+        MatchRuleObject(table_name_regex=r"^others_([a-zA-Z_0-9]+)")
+    ]
+```
+
+After configuring this, users will not be able to edit table and column descriptions of any table matching above match rules
```
ttannis, line 124: nit: I would actually delete the `and will be redirected to the source linked to the table (if any) provided by table metadata service`. The reason is that this in itself is a separate feature/UI decision, and we should limit this section to just describe how to make the table uneditable. What the UI does with it is a separate concern that *could* always change, and so if anything that would be included in some other documentation. Perhaps you could add a full tutorial in our doc site describing the full scope!

dikshathakur3119: Okay, got it. I will remove this form here.

PR #502, opened 2020-07-05
fix: Properly request number of entries for popular tables

jacobhjkim:
### Summary of Changes

The popular_tables GET request is not sending any parameters to the metadata API. Thus you can't request more than 10 entries of popular tables (which is the default number set in the [metadatalibrary](https://github.com/lyft/amundsenmetadatalibrary/blob/master/metadata_service/api/popular_tables.py#L21)). This PR properly requests number of entries for popular tables to the metadata API.

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/api/metadata/v0.py
+++ b/amundsen_application/api/metadata/v0.py
@@ -54,8 +54,8 @@ def popular_tables() -> Response:
     """
     try:
         url = app.config['METADATASERVICE_BASE'] + POPULAR_TABLES_ENDPOINT
-        data = {'limit': app.config['POPULAR_TABLE_COUNT']}
-        response = request_metadata(url=url, data=json.dumps(data))
+        params = {'limit': app.config['POPULAR_TABLE_COUNT']}
+        response = request_metadata(url=url, params=params)
         status_code = response.status_code
 
         if status_code == HTTPStatus.OK:
--- a/amundsen_application/api/metadata/v0.py
+++ b/amundsen_application/api/metadata/v0.py
@@ -54,15 +54,17 @@ def popular_tables() -> Response:
     """
     try:
         url = app.config['METADATASERVICE_BASE'] + POPULAR_TABLES_ENDPOINT
-        response = request_metadata(url=url)
+        data = {
+            'limit': app.config['POPULAR_TABLE_COUNT'],
+            'readers': app.config['POPULAR_TABLE_MINIMUM_READER_COUNT']
+        }
+        response = request_metadata(url=url, data=data)
         status_code = response.status_code
 
         if status_code == HTTPStatus.OK:
             message = 'Success'
             response_list = response.json().get('popular_tables')
-            top4 = response_list[0:min(len(response_list), app.config['POPULAR_TABLE_COUNT',
-                                                                      'POPULAR_TABLE_MINIMUM_READER_COUNT'])]
-            popular_tables = [marshall_table_partial(result) for result in top4]
+            popular_tables = [marshall_table_partial(result) for result in response_list]
         else:
             message = 'Encountered error: Request to metadata service failed with status code ' + str(status_code)
             logging.error(message)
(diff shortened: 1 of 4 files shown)


ttannis:
@jacobhjkim is this ready for review? If so can you update the title. If not, it will be easier for us to understand what needs our our attention vs not by opening your PR as a `Draft`, and only opening the official PR when its ready and tagging someone from the team to take a look.

jacobhjkim:
@ttannis My bad. The content of this PR changed a bit since I am not adding a new config to the frontend library, so I left it as WIP longer than I intended. Now It is ready for a review.

amundsen_application/api/metadata/v0.py
```diff
@@ -54,14 +54,14 @@ def popular_tables() -> Response:
     """
     try:
         url = app.config['METADATASERVICE_BASE'] + POPULAR_TABLES_ENDPOINT
-        response = request_metadata(url=url)
+        data = {'limit': app.config['POPULAR_TABLE_COUNT']}
```
ttannis, line 57: Can your PR link to the metadata implementation as well, I'm not seeing where this gets used in the endpoint.

Additionally `GET` requests don't send request data, instead I would expect we would include the parameters in the url in query string format, and the metadata service will parse that value to use it accordingly.

jacobhjkim: [This](https://github.com/lyft/amundsenmetadatalibrary/blob/master/metadata_service/api/swagger_doc/popular_tables_get.yml) is the swagger doc for popular tables metadata API, and thanks for pointing out the `GET` request stuff.

amundsen_application/api/metadata/v0.py
```diff
@@ -54,14 +54,14 @@ def popular_tables() -> Response:
     """
     try:
         url = app.config['METADATASERVICE_BASE'] + POPULAR_TABLES_ENDPOINT
-        response = request_metadata(url=url)
+        params = {'limit': app.config['POPULAR_TABLE_COUNT']}
+        response = request_metadata(url=url, params=params)
```
feng-tao, line 58: @ttannis FYI, this is the endpoint from metadata side: https://github.com/lyft/amundsenmetadatalibrary/blob/master/metadata_service/api/popular_tables.py#L21

amundsen_application/api/metadata/v0.py
ttannis: I'm not seeing the approach of using `params` work locally for us. I'm not sure why, but I have reservations about how widely supported is `params` in `GET` requests are for all different implementations of request clients.

Existing code in this repository also uses query strings and we don't plan on changing that precedent, so I'd suggest:
```
service_base = app.config['METADATASERVICE_BASE']
count = app.config['POPULAR_TABLE_COUNT']
url = f'{service_base}{POPULAR_TABLES_ENDPOINT}?limit={count}'
```
which also removes the need for any changes in `request_utils.py`.

tests/unit/api/metadata/test_v0.py
```diff
@@ -478,6 +478,17 @@ def test_popular_tables_success(self) -> None:
             self.assertEqual(response.status_code, HTTPStatus.OK)
             self.assertCountEqual(data.get('results'), self.expected_parsed_popular_tables)
 
+        with local_app.test_client() as test:
```
ttannis, line 481: The logic you've added for this PR add the `limit` to the popular tables get request, so these unit tests in the bigger scope of things don't actually unit test this logic since that logic is mocked. Meaning that these additions don't add any testing value, so I would prefer to remove them. What do you think?

jacobhjkim: Yes, I agree. It didn't add much value. 

How do you do an integration test with the metadata API? I run local Docker compose to see if microservices work together, but I hope there is some kind of unit test for that.

ttannis: We currently don't have integration tests, and we still have the task of manually testing api changes.

PR #498, opened 2020-06-30
fix: Update travis.yml

ttannis:
### Summary of Changes

The latest attempt at a release has been failing, even though the master for the most recent commit passes. In the [raw log](https://api.travis-ci.com/v3/job/354590611/log.txt) for the travis build I found the following error: `error: [Errno 11] write could not complete without blocking`.

Searching this error led to https://github.com/travis-ci/travis-ci/issues/8982. This PR tries out the solution from that discussion. 

If this does not help fix the issue, this change will be reverted.

### Tests

N/A

### Documentation

N/A

### CheckList
Make sure you have checked **all** steps below to ensure a timely review.
- [x] PR title addresses the issue accurately and concisely. Example: "Updates the version of Flask to v1.0.2"
    - In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).
- [x] PR includes a summary of changes, including screenshots of any UI changes. 
- [x] PR adds unit tests, updates existing unit tests, __OR__ documents why no test additions or modifications are needed.
- [x] In case of new functionality, my PR adds documentation that describes how to use it.
    - All the public python functions and the classes in the PR contain docstrings that explain what it does
- [x] PR passes all tests documented in the [developer guide](https://github.com/lyft/amundsenfrontendlibrary/blob/master/docs/developer_guide.md#testing)

PR #504, opened 2020-07-07
fix: Remove travis workarounds; Temporarily revert code blocks for release.

ttannis:
#### Remove Travis Workarounds
This PR reverts the travis workarounds for #498, #499, and #500 that did not yield results. More details about [the issue](https://github.com/lyft/amundsen/issues/537) here.

#### Temporarily revert code blocks for release
This PR also temporarily reverts our dashboard query feature to only show the query name instead of the actual query. This allows us to remove the `react-code-blocks` dependency. This library is *NOT* the cause of the reason Travis fails to release to pypi, removing it is just a hack that makes the issue less likely to occur so that we can complete a release. Verified this will lead to a successful release by doing a test: https://pypi.org/project/amundsen-frontend/2.2.0rc3

### Tests

N/A

### Documentation

N/A

### CheckList


feng-tao:
should we produce to a feature branch (2.2.0) instead of master which allows to keep the feature in master while produce pypi version through feature branch?

ttannis:
> should we create a feature branch (2.2.0) instead of master which allows to keep the feature in master while produce pypi version through feature branch?

Yeah you're right I was going to rename this branch and release from this branch to avoid removing the relevant code from master.

feng-tao:
Also is it the issue due to Travis? Ryan suggests on GitHub action which I could spend sometime to get it working once finish the spec and see if it helps.

ttannis:
> Also is it the issue due to Travis? Ryan suggests on GitHub action which I could spend sometime to get it working once finish the spec and see if it helps.

I'm still not sure. I don't know if Travis is the actual "problem", or if there is something we can still do with any part of our build configuration (either in `travis.yml` or `setup.py`) to resolve the issue. I'm putting helpful details in the issue linked above. I'd also be a fan of exploring github actions, I haven't looked at it either but I wonder if at some point they might have a smoother process.

ttannis:
Closing as I will not be merging this to master, but details are documented. Release will. be created from https://github.com/lyft/amundsenfrontendlibrary/tree/v2.2.0

PR #507, opened 2020-07-10
fix: added max-width style property with 560px value

allisonsuarez:
### Summary of Changes

Added max-width style property with 560px value to /static/js/components/NavBar/styles.scss.
Screen Shot 2020-07-10 at 2 29 51 PM

### Tests

No tests added.

### Documentation

No docs added.

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/NavBar/styles.scss
+++ b/amundsen_application/static/js/components/NavBar/styles.scss
@@ -80,7 +80,11 @@ $avatar-container-size: 40px;
   .search-bar {
     flex-grow: 1;
     margin: auto $spacer-2 auto auto;
+  }
+
+  #search-bar {
     max-width: 560px;
+    margin: auto 0px auto auto;
   }
 
   .logo-icon {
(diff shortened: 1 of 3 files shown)


amundsen_application/static/js/components/NavBar/styles.scss
```diff
@@ -82,6 +82,11 @@ $avatar-container-size: 40px;
     margin: auto $spacer-2 auto auto;
   }
 
+  #search-bar {
```
ttannis, line 85: nit: Can you next these styles. This matter because whether or not the styles for the element with id `#search-bar` will work is also dependent on the styles defined in its parent with class name `.search-bar`. The effect will be the same, but organizational wise it gives someone the context to understand that if they change something in `.search-bar`, the desired affect of `#search-bar` could also be affected.

nit: Also, we tend to use explicit class names for the styles instead of the id. So here you can add a new classname in the markup and use that instead. That also raises the concept that the classname `search-bar` should have a more specific className like `nav-search-bar`, then the actual search bar can have the `search-bar` class.  So overall:

```
.nav-search-bar {
    flex-grow: 1;
    margin: auto $spacer-3 auto auto;

    .search-bar {
      max-width: 560px;
      margin: auto 0px auto auto;
    }
}
```

PR #508, opened 2020-07-12
fix: Fixing EditableSection component

dikshathakur3119:
### Summary of Changes
Fixing Editable Section Component since following metadata elements are behaving strangely:
- Tags "Add +" button does not work
- Clicking on a blank area inside the editable section toggles edit mode (unexpected behavior). This also results in the tag editing being difficult – clicking in a blank area will enter mode and then trying to select the text box to type a new tag will then exit edit mode because of the state the component is in. 
When editing an editable description, clicking on the "cancel" button should exit edit mode does nothing

### Tests
No new test added.

### Documentation
NA

### CheckList


amundsen_application/static/js/components/common/EditableSection/index.tsx
```diff
@@ -118,15 +118,15 @@ export class EditableSection extends React.Component<
 
     return (
       <section className="editable-section">
-        <label className="editable-section-label">
+        <div className="editable-section-label">
```
ttannis, line 121: This element is intended to a be `label`. Previous discussion: https://github.com/lyft/amundsenfrontendlibrary/pull/480#discussion_r441140060

Can you explain this fix and why you think this works? When making fixes we are interested in debugging the source of the problem, and fixing the issue at the source -- even if other hacks or changes are successful in making the problem go away, we're always concerned about making the best fix possible. 

I would revisit our exploration of event propagation, and the fact that we say that clicking on a child in `EditableSection` was triggering `toggleEdit` when it wasn't supposed to.

dikshathakur3119: While I was debugging this and monitoring event propagation I observed following for two cases:
Case1 Editable : when table description is editable
    - when we hover anywhere within <label> tags, `edit` icon becomes visible and changes its color to `violet` , means it gets activated and if we click anywhere inside `<label>` tags it will run `toggleEdit function`.
Case2 Read Only: When table description is un-editable
     - when we hover anywhere within <label> tags,`edit` icon becomes visible in `gray` color, and when we hover over `edit` icon only then it changes its color to `violet`. That means `edit` icon in this case gets activated only when we hover over icon itself.
Later is the case which is expected.
In my understanding, difference arise due to rendering of different tags in both cases. We are using `<button>` for editable description and `overlayTrigger` for un-editable. 
But when I use `<div>` tag instead of `<label>`, `edit button` becomes activated only when we click on `edit` icon (not anywhere inside div tags). Hence when we click on any button inside child component like `update` and `cancel`, `edit` button will not get clicked.
Event Propagation takes place when event listeners for both parent and child component are active on same area, which was happening earlier. So I de-coupled it, now `onclick` event inside child component does not activate event listener of parent component.
I tried to stop event propagation by identifying event origin and checking `event.target.name` inside `toggelEdit` function but it was not able to distinguish between origin. 
I also found out that explicitly stopping event propagation by using functions should be avoided as it may affect other functionalities. This may not happen in our case but it is still not recommended.

ttannis: > I also found out that explicitly stopping event propagation by using functions should be avoided as it may affect other functionalities.

Can you link to the source for this. Overall, with web technologies there are many ways to achieve different things, and cases where some things may or may not be recommended -- but at the end of the day engineers have to weigh the pros and cons of certain recommendations within the context of their specific use cases and decide what is best for the application. For example, some questions to ask in this case would be what functionalities would stopping the event propagation affect, and do we care about those functionalities for this component?

As it stands, using a `label` here is also a best practice that we should maintain, so let's consider any other options (i.e. if there is some other fix you can identify without using event propagation), and discuss the pros and cons of each one.

dikshathakur3119: Here are sources of that give this recommendation [1](https://css-tricks.com/dangers-stopping-event-propagation/) , [2](https://javascript.info/bubbling-and-capturing). But yes you are right, we have to weigh pros and cons at the end.

dikshathakur3119: I saw [this](https://javascript.info/bubbling-and-capturing) article which suggests I can compare `event.target` and `event.currentTarget` to know from where event has bubbled., but this is also not working for in our case, both values are same for different cases.

amundsen_application/static/js/components/common/EditableText/index.tsx
```diff
@@ -86,15 +86,24 @@ class EditableText extends React.Component<
...
-  exitEditMode = () => {
+  exitEditMode = (event?: React.MouseEvent<HTMLElement>) => {
+    if (event) {
+      event.preventDefault();
+    }
     this.props.setEditMode(false);
   };
 
-  enterEditMode = () => {
+  enterEditMode = (event?: React.MouseEvent<HTMLElement>) => {
```
ttannis, line 96: nit: can specify `<HTMLAnchorElement>`

amundsen_application/static/js/components/common/EditableText/index.tsx
```diff
@@ -86,15 +86,24 @@ class EditableText extends React.Component<
     }
   }
 
-  exitEditMode = () => {
+  exitEditMode = (event?: React.MouseEvent<HTMLElement>) => {
```
ttannis, line 89: nit: can specify `<HTMLButtonElement>`

amundsen_application/static/js/components/Tags/TagInput/index.tsx
```diff
@@ -298,6 +307,7 @@ class TagInput extends React.Component<TagInputProps, TagInputState> {
           isValidNewOption={this.isValidNewOption}
           name="tags"
           noOptionsMessage={this.noOptionsMessage}
+          onClick={this.stopPropagation}
```
ttannis, line 310: The reason this doesn't work is that `CreateableSelect` is a 3rd party component, created just like all the custom components we create here, where it can only be interacted it via its properties. This component does not have an `onClick` method. See the link left in the comments: https://react-select.com/props

However you should be able to put an `onClick={this.doTheThing}` on the parent `<div>` in this file for it to get triggered.

danwom:
I suspect the root of this problem would be nesting the `{childrenWithProps}` under the `label` in the `EditableSection` component. The label element is special where it tries to connect click events to some input such as a checkbox or radio button. In this case, it looks like it activates the edit state on any click of anything inside the `label`. 

I'd recommend changing the label back to a div/span, or moving `{childrenWithProps}` outside of the label. That way we don't need to add a `stopPropagation` to all of the children components. LMK what you think.

https://github.com/lyft/amundsenfrontendlibrary/blob/c32080eaf8fb375976e6f374b3a5a961b3584541/amundsen_application/static/js/components/common/EditableSection/index.tsx#L124-L132

ttannis:
> I suspect the root of this problem would be nesting the `{childrenWithProps}` under the `label` in the `EditableSection` component. The label element is special where it tries to connect click events to some input such as a checkbox or radio button. In this case, it looks like it activates the edit state on any click of anything inside the `label`.
> 
> I'd recommend changing the label back to a div/span. That way we don't need to add a `stopPropagation` to all of the children components. Alternatively moving `{childrenWithProps}` outside of the `label` would still activate the edit state on clicking on the title, which is not intended.
> 
> LMK what you think.
> 
> https://github.com/lyft/amundsenfrontendlibrary/blob/c32080eaf8fb375976e6f374b3a5a961b3584541/amundsen_application/static/js/components/common/EditableSection/index.tsx#L124-L132

There's some previous context regarding changes that were done to adhere to semantic html & accessibility best practices, and so the explorations we were doing weren't only trying  to fix the issue but exploring if there is a way to maintain both those changes plus the desired behavior. A timeboxed effort on my end didn't yield any results, closest option I found will maintain semantic html & desired behavior at the expense of accessibility decision in https://github.com/lyft/amundsenfrontendlibrary/pull/480#discussion_r441140060

ttannis:
@dikshathakur3119  given the guidance from design to only have the edit button trigger the state, I'm proposing a fix that will prioritize the desired behavior at the expense of a piece of the decisions described inhttps://github.com/lyft/amundsenfrontendlibrary/pull/480#discussion_r441140060.
1. Seconding what @danwom said above, let's move the children outside of the label.
2. On top of that, if you add an `onClick` method to the label that calls `event.preventDefault()` will make it such that the editable state is only triggered when clicking on the button.

Let us know if you have any questions here. The iterations were worthwhile as this ticket helped you learn about some new topics, but we can wrap up with a best effort solution.

amundsen_application/static/js/components/common/EditableSection/index.tsx
```diff
@@ -47,6 +47,10 @@ export class EditableSection extends React.Component<
     this.setState({ isEditing: !this.state.isEditing });
   };
 
+  stopPropagation = (event: React.MouseEvent<HTMLDivElement>) => {
```
ttannis, line 50: Looks, great just a couple more things:
1. Let's call this `preventDeafult` since that's the name of the method it calls. 
2. Let's also add a unit test for this new functionality, to try to prevent the bug from happening again as this component gets updated. We will want to simulate a click event on the element, and verify that `prefentDefault` was called. Here's an example: https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/components/common/Bookmark/BookmarkIcon/index.spec.tsx#L39-L44

PR #511, opened 2020-07-15
fix: Getting tooltip on edit button of Column Description if un-editable

dikshathakur3119:
### Summary of Changes

We have provided feature for making table and column description un-editable, in this PR we are providing link to source file from column's edit button if someone wants to change column description at source file.
Edit link and source is same as table source link. 

### Tests
NA

### Documentation

NA

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/TableDetail/ColumnList/index.tsx
+++ b/amundsen_application/static/js/components/TableDetail/ColumnList/index.tsx
@@ -6,10 +6,14 @@ import './styles.scss';
 
 interface ColumnListProps {
   columns?: TableColumn[];
+  editText?: string;
+  editUrl?: string;
 }
 
 const ColumnList: React.SFC<ColumnListProps> = ({
   columns,
+  editText,
+  editUrl,
 }: ColumnListProps) => {
   if (columns.length < 1) {
     return <div />;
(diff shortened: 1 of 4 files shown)


dikshathakur3119:
@ttannis  @allisonsuarez  let me know if this is good way to pass `editText` and `editUrl` to `ColumnListItem`. I am planning to add a unit test case as well.

allisonsuarez:
> @ttannis @allisonsuarez let me know if this is good way to pass `editText` and `editUrl` to `ColumnListItem`. I am planning to add a unit test case as well.

I think this makes sense assuming that all of the column descriptions would be uneditable rather than just individual columns being un-editable. I think is how this is supposed to work so looks good to me.

allisonsuarez:
> @ttannis @allisonsuarez let me know if this is good way to pass `editText` and `editUrl` to `ColumnListItem`. I am planning to add a unit test case as well.

it also follows the same way this is being done for editable descriptions which is good for consistency.

ttannis:
> > @ttannis @allisonsuarez let me know if this is good way to pass `editText` and `editUrl` to `ColumnListItem`. I am planning to add a unit test case as well.
> 
> I think this makes sense assuming that all of the column descriptions would be uneditable rather than just individual columns being un-editable. I think is how this is supposed to work so looks good to me.

Agreed, you mentioned feeling as if you were making components messy, but passing properties from parents to children is a regular thing to do. This implementation is fine from my perspective as well.

PR #515, opened 2020-07-16
chore[fe]: apply license headers for fe (html, scss, tsx)

feng-tao:
### Summary of Changes

### Tests

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/static/css/_animations.scss
+++ b/amundsen_application/static/css/_animations.scss
@@ -1,3 +1,6 @@
+// Copyright Contributors to the Amundsen project.
+// SPDX-License-Identifier: Apache-2.0
+
 @import 'variables';
 
 $loading-duration: 1s;
--- a/amundsen_application/static/css/_avatars.scss
+++ b/amundsen_application/static/css/_avatars.scss
@@ -1,3 +1,6 @@
+// Copyright Contributors to the Amundsen project.
+// SPDX-License-Identifier: Apache-2.0
+
 .sb-avatar > img {
   margin: 0;
 }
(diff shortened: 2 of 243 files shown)


ttannis:
Do we really have to put this on every file? Is there some information you can provide about this and what it means?

feng-tao:
I don't have much context, but mostly follow what LFAI suggested. This is from one email Mark cc from LFAI.
```
 We did notice also that although the repos have top-level LICENSE files, most of the source code files do not have license headers or references in them. We encourage including a license statement in each file where reasonably possible; this provides many benefits, including making it easier for other projects to reuse the open source files without losing track of their license information. One easy way to do this is by adding a single line comment to the top of each file with its SPDX license ID, such as the following for Amundsen source code:

SPDX-License-Identifier: Apache-2.0

or the following for Amundsen documentation, assuming it adopts CC-BY-4.0 for documentation:

SPDX-License-Identifier: CC-BY-4.0

More information about these SPDX short-form identifiers can be found at https://spdx.dev/ids. Using the SPDX license ID will save you from putting a full licensing header and it is also discoverable by source code scanning tools. 
```

Then I look at an existing example of one of LFAI project from UBER (https://github.com/pyro-ppl/pyro/blob/dev/pyro/ops/arrowhead.py#L1-L2) and they did this. Hence that's how I come up with the pr. I think we only need to apply the headers to source code in a best effort.

ttannis:
Also have concerns about what this means legally. For example with some of the css and style choices (example: All color specifications in `_colors.scss`) are concepts that we are currently allowed take from Lyft's internal design specifications to use in this tool. 

But then if we're now applying some copyright to that, I wonder is that will change the decision we originally got from the people who work on Lyft Product Language. I can look into it, but first trying to understand the implications of this copyright text.

feng-tao:
good point @ttannis , unfortunately I didn't see any existing project within LFAI that have code using css or HTML. Let me cc you in the same email Mark cc me so that you could clarify with the LFAI folks

PR #516, opened 2020-07-18
fix: Footer consistently in loading state

dikshathakur3119:
### Summary of Changes

Footer shows when did Amundsen last indexed, it loads properly for first time but when we navigate to other pages it goes into loading state forever.

### Tests

No addition

### Documentation

No addition

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/Footer/index.tsx
+++ b/amundsen_application/static/js/components/Footer/index.tsx
@@ -33,6 +33,10 @@ export class Footer extends React.Component<FooterProps> {
     this.props.getLastIndexed();
   }
 
+  componentDidUpdate() {
+    this.props.getLastIndexed();
+  }
+
   generateDateTimeString = () => {
     return formatDateTimeLong({ epochTimestamp: this.props.lastIndexed });
   };
--- a/amundsen_application/static/js/components/Footer/index.tsx
+++ b/amundsen_application/static/js/components/Footer/index.tsx
@@ -36,10 +36,6 @@ export class Footer extends React.Component<FooterProps> {
     this.props.getLastIndexed();
   }
 
-  componentDidUpdate() {
-    this.props.getLastIndexed();
-  }
-
   generateDateTimeString = () => {
     return formatDateTimeLong({ epochTimestamp: this.props.lastIndexed });
   };
--- a/amundsen_application/static/js/ducks/tableMetadata/reducer.ts
+++ b/amundsen_application/static/js/ducks/tableMetadata/reducer.ts
@@ -309,7 +309,10 @@ export default function reducer(
         },
       };
     case GetTableData.REQUEST:
-      return initialState;
+      return {
+        ...initialState,
+        lastIndexed: state.lastIndexed,
+      };
     case GetTableData.FAILURE:
       return {
         ...state,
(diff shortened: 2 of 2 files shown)


amundsen_application/static/js/components/Footer/index.tsx
```diff
@@ -33,6 +33,10 @@ export class Footer extends React.Component<FooterProps> {
     this.props.getLastIndexed();
   }
 
+  componentDidUpdate() {
```
ttannis, line 36: My understanding of the bug is that `this.props.lastIndexed` is set when the component mounts however when the component goes through the update cycle when we navigate to a new page, `this.props.lastIndexed` become undefined. So while re-requesting the data will result in the desired outcome, it has two down-sides:
1. It doesn't answer the question of the source of the bug, which is "why is `this.props.lastIndexed` null?"
2. We try to avoid unnecessary network requests whenever possible. We should be able to request the data once and have the application correctly save it in the application state. This is especially valuable for a value like `lastIndexed` which in our internal use case only changes twice a day. 

Next, let's think about the source of the issue. If you look at where `lastIndexed` is stored, it's actually stored under the `globalState.tableMetadata` -- which gets reset to it's initial state when we go to a new table detail page. Also notice the bug doesn't occur when we navigate from the homepage to the dashboard page. So my conclusion for the source of the bug and potential fix is:
1. The `lastIndexed` state is stored on the `tableMetadata` state, which gets re-set to null when we go to the table detail page. It then becomes `null` in the `Footer`, and the user is stuck seeing the loading state.
2. Therefore the fix is to prevent `lastIndexed` from getting re-written when we request table metadata.

We can do this two ways:
1. Decouple the `lastIndexed` state from `tableMetadata`. We associated `lastIndexed` with `tableMetadata` Back back when we only had one resource, but now that no longer makes sense. This solution is the best for the longterm, but it will require creating a new reducer for all the single values that applying across the application and has a larger work effort.
2. We can also prevent `lastIndexed` from getting overridden by updating the logic in the tableMedata reducer as follows:
```
case GetTableData.REQUEST:
    return {
        ...initialState,
        lastIndexed: state.lastIndexed, // lastIndexed will not be overridden, because we set it back it to its current value
```

What do you think of this diagnosis of the bug, as well as the suggested fixes? Let me know if you have any questions, opinions, or preferences regarding the proposed solutions.

dikshathakur3119: Yes it is a valid point that we should not make unnecessary network requests. I was not sure why `lastIndexed` values was getting re-written. I think it will be good if I create new reducer for all single values so that we do not face same issue with any other value as well.

feng-tao: chimein here: I assume the lastIndexed state is https://github.com/lyft/amundsendatabuilder/blob/master/databuilder/models/neo4j_es_last_updated.py which shows when our ETL is indexed. I think the option 1 seems more reasonable and cleaner...

ttannis: I do favor option1 as well, but one thing to raise is that option2 will serve as a 1-line stop gap that will address the issue until option1 is implemented. 

@dikshathakur3119 option1 will require a bit of ramping up in Redux + you are on call next week, so what do you think of submitting a PR using option 2 just to fix the issue for our users, but then you can still take the time to implement option1 next sprint.

dikshathakur3119: Sounds good! I will put option2 fix here and create another PR to implement option1 for solving this bug in a better way.

PR #518, opened 2020-07-22
fix: Limit max char to display for preview column

feng-tao:
### Summary of Changes

Limit the max length of preview column for display

### Tests

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/TableDetail/DataPreviewButton/constants.ts
+++ b/amundsen_application/static/js/components/TableDetail/DataPreviewButton/constants.ts
@@ -1,0 +1,2 @@
+export const PREVIEW_COLUMN_MAX_LEN = 50;
+export const PREVIEW_COLUMN_MSG = 'Data Exceeds Render Limit';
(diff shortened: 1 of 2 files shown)


amundsen_application/static/js/components/TableDetail/DataPreviewButton/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const PREVIEW_COLUMN_MAX_LEN = 50;
```
ttannis, line 1: We would need to be a lot more generous here, a lot of data is more than 50 characters but still doesn't pose a problem (example: some of our hostnames are > 50 characters if the service name is long). For now, I'd suggest targeting only really really long fields like the polylines, and set this to at least 250. What do you think?

feng-tao: +1, i choose 50 mostly based on the initial ticket(https://jira.lyft.net/browse/AMD-1545) that the user suggested.

PR #519, opened 2020-07-24
feat: making clickable badges

allisonsuarez:
### Summary of Changes

Added a component that leverages the style fo the Flag component to make badges on Amundsen clickable. On click a search should be performed with that badge value.

### Documentation

N/A

### CheckList


allisonsuarez:
👀  @ttannis @dikshathakur3119

allisonsuarez:
🔨

amundsen_application/static/js/components/Badges/styles.scss
```diff
@@ -0,0 +1,15 @@
+// Copyright Contributors to the Amundsen project.
+// SPDX-License-Identifier: Apache-2.0
+
+@import 'variables';
+
+.clickable-badge-primary {
```
ttannis, line 6: For adding the hover styles, those are new specifications within our designs which our components don't have proper support for. It was an unexpected aspect of this ticket that requires some more dedication and what we might refer to as "scope creep". So what I want to propose is that we split your task into two PRs. This one will provide the functionality and the second will be purely UI and CSS focused. What do you think? 

My concern here is just that I didn't intend for your initial tickets to introduce too many concepts at once. I saw this ticket as an opportunity to build a new component, basic css concepts,  and unit tests -- but implementing the overlay as desired may involve specific CSS concepts + additional component restructuring.

amundsen_application/static/js/components/common/BadgeList/index.spec.tsx
```diff
@@ -38,9 +38,9 @@ describe('BadgeList', () => {
       expect(container.exists()).toBe(true);
     });
 
-    it('renders a <Flag> for each badge in the input', () => {
-      expect(badgeList.find(Flag).length).toEqual(badges.length);
-    });
+    // it('renders a <ClickableBadge> for each badge in the input', () => {
```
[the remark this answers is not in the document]
allisonsuarez: Yes, I am not sure if I refactored the test correctly or not so I commented it out in the meanwhile

amundsen_application/static/js/components/Badges/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  render() {
+    return (
+      <span
+        className={`clickable-badge-${this.props.labelStyle}`}
+        onClick={this.onClick}
+      >
+        <Flag
+          caseType={this.props.caseType}
+          text={this.props.text.replace(/[-_]/g, ' ')}
```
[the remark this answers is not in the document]
allisonsuarez: I agree with this, I also found it confusing at first

amundsen_application/static/js/components/Badges/styles.scss
allisonsuarez: I will create a new PR to introduce the style changes. I was playing around with this and got something close to what we want so I think it should't be too bad.

amundsen_application/static/js/config/config-utils.ts
```diff
@@ -84,7 +84,7 @@ export function getBadgeConfig(badgeName: string): BadgeStyleConfig {
 
   return {
     style: BadgeStyle.DEFAULT,
-    displayName: badgeName.replace(/[-_]/g, ' '),
+    displayName: badgeName,
```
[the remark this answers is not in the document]
feng-tao: @ttannis do you know why we did `badgeName.replace(/[-_]/g, ' '),` before?

amundsen_application/static/js/components/common/BadgeList/index.spec.tsx
```diff
@@ -4,7 +4,7 @@
 import * as React from 'react';
 import { shallow } from 'enzyme';
 
-import Flag from 'components/common/Flag';
+import { ClickableBadge } from 'components/common/Badges';
```
ttannis, line 7: To get the tests to pass update to `import ClickableBadge from 'components/common/Badges';`

amundsen_application/static/js/components/common/Badges/styles.scss
```diff
@@ -0,0 +1,12 @@
+// Copyright Contributors to the Amundsen project.
+// SPDX-License-Identifier: Apache-2.0
+
+@import 'variables';
+
+.clickable-badge-primary {
```
ttannis, line 6: This only applies the style to badges with the primary style, but we will want it applied to all badges all of the time. So you can either:
1. Use the suggestion from my previous comment that will apply this to any `.clickable-badge-xyz` classes.
2. Update the component to split the classNames such that every badge has the `clickable-badge` class + a class that defines the labelStyle, and update this style here to `.clickable-badge`. 
e.g. ` className={'clickable-badge ${this.props.labelStyle}'}`

amundsen_application/static/js/components/common/BadgeList/index.spec.tsx
feng-tao: is it running unit test just by `npm run test` locally which has the same effect of running it on CI?

ttannis: @feng-tao yes.

amundsen_application/static/js/config/config-utils.ts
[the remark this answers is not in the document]
ttannis: I think product/design wanted the text in the UI to appear differently than the badge name stored in metadata. But we don't see value in that anymore since badge search requires knowing the exact name. If we modify the badge name in the UI users aren't going to know what they need to type in when they want to use the filter UI.

Alternatively though, if we also create the API to get all the badges -- we could also go ahead and update the badge filter to UI so users can select instead of having to know what name to type in.

amundsen_application/static/js/components/common/Badges/styles.scss
allisonsuarez: Yeah until I implement the style changes there is no reason to have a distinction here so I will just change it to clickable-badge, I can use your suggestion for the distinction on my next PR

amundsen_application/static/js/components/common/Badges/styles.scss
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+@import 'variables';
+
+.clickable-badge {
+  .flag {
+    &:hover {
+      cursor: pointer;
+    }
+  }
```
ttannis, line 11: nit: On second thought I should have suggested this instead. 
```
clickable-badge {
    &:hover {
      cursor: pointer;
    }
}
```
The reason is that the. pointer cursor should appear when hovering over the element that triggers an interaction, and based on your markup, the element that the click event is attached to is the `clickable-badge` itself.

amundsen_application/static/js/components/common/Badges/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { UpdateSearchStateRequest } from 'ducks/search/types';
+import { logClick } from 'ducks/utilMethods';
+
+import './styles.scss';
+
+export interface DispatchFromProps {
+  searchBadge: (badgeText: string) => UpdateSearchStateRequest;
+}
+
+export type ClickableBadgeProps = FlagProps & DispatchFromProps;
```
ttannis, line 20: Let's add a unit test for this file before wrapping up. When writing a unit test, I first like to think about what could be important to verify for the new component and/or updated behavior. For this component the following three cases come to mind:
1. Verify the component renders the `Flag` as expected, I think you've already seen examples of this in the `BadgeList` test.
2. We want to simulate a click event and verify the correct thing happens. [Example](https://github.com/lyft/amundsenfrontendlibrary/blob/ccfd2d6b82957fef347e956b243e4048c191fc0d/amundsen_application/static/js/components/common/Bookmark/BookmarkIcon/index.spec.tsx#L34-L44).
3. Verify the `mapDispatchToProps` functionality (as that is what makes the feature work!). We admittedly don't have good examples of this in the repo, but let me push an example with `TagInfo` for reference.

amundsen_application/static/js/components/common/BadgeList/index.spec.tsx
```diff
@@ -38,10 +38,11 @@ describe('BadgeList', () => {
       expect(container.exists()).toBe(true);
     });
 
-    it('renders a <Flag> for each badge in the input', () => {
-      expect(badgeList.find(Flag).length).toEqual(badges.length);
+    it('renders a <ClickableBadge> for each badge in the input', () => {
+      expect(badgeList.find(ClickableBadge).length).toEqual(badges.length);
     });
 
+    // TODO refactor to be clickable badge isntead of flag
```
ttannis, line 45: What is intended by this todo?

amundsen_application/static/js/components/common/Badges/index.spec.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    it('renders a <Flag> for the ClickableBadge', () => {
+      const flagPerBadge = 1;
+      expect(wrapper.find(Flag).length).toEqual(flagPerBadge);
+    });
+
+    it('renders with correct text', () => {
+      expect(wrapper.find(Flag).props().text).toEqual(props.text);
+    });
+    //   TODO look into why no className defined for Flag in this case
```
ttannis, line 74: Yeah let's remove this test. I think the existing tests cover the most useful ideas to test.

amundsen_application/static/js/components/common/BadgeList/index.spec.tsx
allisonsuarez: The test right below it refers to Flag but the component is now Clickable badge so I need to update the test to reflect this

PR #521, opened 2020-07-24
fix: Fix layout

ttannis:
### Summary of Changes

This PR fixes some of the layout issues we have observed recently, such as https://github.com/lyft/amundsen/issues/575

There were some updates made to layout a few weeks ago as part of https://github.com/lyft/amundsenfrontendlibrary/commit/a87354bda52f5151d8785f4a0d7023a51e215df2, in order to try to enforce a better horizontal alignment of metadata sections that appear in columns. However not only did those styles cause layout issues on skinnier screens, but our new design guidance for the layout in the left panel no longer includes the need to force the minimum height or horizontal alignment of metadata section headers.

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/css/_layouts.scss
+++ b/amundsen_application/static/css/_layouts.scss
@@ -117,7 +117,6 @@ $aside-separation-space: 40px;
       .metadata-section {
         margin-top: $aside-separation-space;
         display: inline-table;
-        min-height: 70px;
       }
 
       .editable-text {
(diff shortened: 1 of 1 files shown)


ttannis:
> oh, so it is not related to recent chrome upgraded for flexbox support? but great that we root cause the issue.
> cc @jonhehir who experienced the UI issue for FYI.

@feng-tao  yeah it wasn't that. The reason this fix works is a bit difficult to explain, but When I saw this bug I guessed that there must have been some conflicting style specifications resulting in the issues. The reason I guessed this is that in my experience it's unlikely for a well developed tool like Chrome to add support in their updates that breaks any valid css.  

So under that assumption I started looking at what styles we could have added to cause this. This one was a bit easier on my end, since I was able to remember recent changes I made and was able to start looking at older commits to figure it out.

feng-tao:
nice, also the CLA check hangs, close and reopen pr could fix it.

PR #524, opened 2020-07-28
fix: Correctly set column editable state

ttannis:
### Summary of Changes

The for loop where the column editable state got set was done written incorrectly. It was nested inside a check for `app.config['COLUMN_STAT_ORDER']`, so it would not be set if the configuration did not exist. This PR updates the logic to work as intended. 

### Tests

N/A

### Documentation

Will be called out in release. 

### CheckList

### DIFF
--- a/amundsen_application/api/utils/metadata_utils.py
+++ b/amundsen_application/api/utils/metadata_utils.py
@@ -96,14 +96,15 @@ def marshall_table_full(table_dict: Dict) -> Dict:
     for reader_object in readers:
         reader_object['user'] = _map_user_object_to_schema(reader_object['user'])
 
-    # If order is provided, we sort the column based on the pre-defined order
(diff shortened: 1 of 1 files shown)


ttannis:
> lgtm, so the bug seems to exist since the beginning that users can't change col description if no COLUMN_STAT_ORDER is provided?
> 
> A bit surprised that the issue is brought up recently. But great that we fix it.


> lgtm, so the bug seems to exist since the beginning that users can't change col description if no COLUMN_STAT_ORDER is provided?
> 
> A bit surprised that the issue is brought up recently. But great that we fix it.

Yes, seems so. We haven't caught it because we always had that configuration set. The nature of the bug makes it such that someone might only encounter this before they fully set up their configuration, or if they don't use column stats.

PR #525, opened 2020-07-29
chore: Update a test

ttannis:
####  Context
This is a test of `mapDispatchToProps` that provides slightly more value than many current examples. The purpose of this method is to connect a presentational component to Redux in order to fire actions that ultimately trigger API calls. 

#### Improvement
Many of our current tests do a bare minimum check, to just verify that the expected property exists and is a function. This example goes further because it verifies the expected property is a function by calling it (test would fail if it wasn't) + verifies that calling it triggers the expected action. 

In the event that `mapDispatchToProps` has extra logic written into it that is worth testing (e.g. boolean of if/elese logic) we could test that too. 

### Tests

Updating a test to provide example.

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/Tags/TagInfo/index.spec.tsx
+++ b/amundsen_application/static/js/components/Tags/TagInfo/index.spec.tsx
@@ -5,10 +5,16 @@ import * as React from 'react';
 import { shallow } from 'enzyme';
 
 import * as UtilMethods from 'ducks/utilMethods';
+import { updateSearchState } from 'ducks/search/reducer';
+
 import { mapDispatchToProps, TagInfo, TagInfoProps } from '.';
 
 const logClickSpy = jest.spyOn(UtilMethods, 'logClick');
 logClickSpy.mockImplementation(() => null);
+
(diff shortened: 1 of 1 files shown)


amundsen_application/static/js/components/Tags/TagInfo/index.spec.tsx
```diff
@@ -5,11 +5,17 @@ import * as React from 'react';
...
 
 import * as UtilMethods from 'ducks/utilMethods';
+import { updateSearchState } from 'ducks/search/reducer';
+
 import { mapDispatchToProps, TagInfo, TagInfoProps } from '.';
 
 const logClickSpy = jest.spyOn(UtilMethods, 'logClick');
 logClickSpy.mockImplementation(() => null);
 
+jest.mock('ducks/search/reducer', () => ({
```
ttannis, line 15: Our testing framework `Jest` requires that any methods we will run an assertion on needs to be be mocked. This syntax is one way to mock imported methods, and you only have to mock the ones that will be used in the test file.

PR #527, opened 2020-07-30
feat: Update prog_desc configuration + UI

ttannis:
#### Context
We want to update how programmatic descriptions appear in the UI, by allowing them to be inserted more naturally into our layout. Programmatic descriptions are still meant to appear below supported structured metadata, but this work allows them to be configured such that they either can fall in with the existing column layout, or just be rendered below the columns using the full width of that panel.

#### Implementation
This PR extends the existing `PROGRAMMATIC_DISPLAY` configuration to include a `LEFT` and `RIGHT` configuration, and updates the UI rendering logic. Alternatives are up for discussion, though we want to ensure backwards compatibility for this version.

### Tests

TODO after finalizing implementation

### Documentation

TODO after finalizing implementation

### CheckList


ttannis:
cc @samshuster

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -150,16 +150,75 @@ def marshall_dashboard_full(dashboard_dict: Dict) -> Dict:
...
+         "test2" : { "display_order": 2 }
+       },
+       'test5': {"display_order": 1},
+       'test4': {"display_order": 0},
+    }
+    """
+    prog_descriptions = []
+    prog_descriptions.append({'source': 'test7', 'text': 'Test'})
+    prog_descriptions.append({'source': 'test8', 'text': 'Test'})
+    prog_descriptions.append({'source': 'test6', 'text': 'Test'})
```
ttannis, line 182: TODO: `test6` isn't showing up need to update default order logic.

feng-tao:
cc @samshuster 
It would be good if Sam could take a look as well.

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -150,16 +150,75 @@ def marshall_dashboard_full(dashboard_dict: Dict) -> Dict:
...
+         "test1" : { "display_order": 1 },
+         "test0" : { "display_order": 0 },
+         "test2" : { "display_order": 2 }
+       },
+       'test5': {"display_order": 1},
+       'test4': {"display_order": 0},
+    }
+    """
+    prog_descriptions = []
+    prog_descriptions.append({'source': 'test7', 'text': 'Test'})
```
feng-tao, line 180: I assume this is for testing?

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -150,16 +150,75 @@ def marshall_dashboard_full(dashboard_dict: Dict) -> Dict:
...
+        if left_config:
+            left = [x for x in prog_descriptions if x.get('source') in left_config]
+            left.sort(key=lambda x: _sort_prog_descriptions(left_config, x))
+
+        right_config = prog_display_config.get('RIGHT')
+        if right_config:
+            right = [x for x in prog_descriptions if x.get('source') in right_config]
+            right.sort(key=lambda x: _sort_prog_descriptions(right_config, x))
+
+    updated_descriptions['left'] = left
```
feng-tao, line 218: need to take a closer look once have time, but how / which programmatic descriptions we put on LEFT/RIGHT/OTHER?

amundsen_application/api/utils/metadata_utils.py
ttannis: Yes, and setting up different edge cases.

amundsen_application/static/js/components/TableDetail/index.tsx
```diff
@@ -284,6 +284,16 @@ export class TableDetail extends React.Component<
                     <div className="section-title title-3">Frequent Users</div>
                     <FrequentUsers readers={data.table_readers} />
                   </section>
+                  {data.programmatic_descriptions.left.map((d) => (
+                    <EditableSection title={d.source} readOnly>
```
Golodhros, line 288: Curious if React doesn't complain about this component not having a key attribute.

amundsen_application/api/utils/metadata_utils.py
ttannis: That we be up to the engineer to set the FE configuration based on what they want.

amundsen_application/static/js/components/TableDetail/index.tsx
ttannis: It surprisingly did not -- at least in the build script -- but I will add them

ttannis: Will also pull this out into a helper method since it's the same logic 3 times.

samshuster:
hey @ttannis if I am seeing this right it looks like there should be no problems for backwards compatibility as we can always just opt to configure the prog_descriptions without left/right if users prefer the full width. Thanks for providing that option!

Were there any issues with rendering when markdown text was supplied in the left and right sections?

ttannis:
> hey @ttannis if I am seeing this right it looks like there should be no problems for backwards compatibility as we can always just opt to configure the prog_descriptions without left/right if users prefer the full width. Thanks for providing that option!
> 
> Were there any issues with rendering when markdown text was supplied in the left and right sections?

No issues with rendering, though I'd say it's the responsibility of the developer to leverage the full width option if they know the data will be better displayed that way (e.g. a chart, large amount of markdown, etc).

Thanks for the input. I'll be cleaning up this code now, but still giving some time for others to comment as well.

amundsen_application/api/utils/metadata_utils.py
ttannis: Fixed in recent upload.

ttannis:
@feng-tao @Golodhros final review. 

I will also post in our internal tickets what exact markup needs to be ingested for our Lyft specific programmatic description usage to match the desired UI.

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -150,16 +149,43 @@ def marshall_dashboard_full(dashboard_dict: Dict) -> Dict:
     return dashboard_dict
 
 
-def _update_prog_descriptions(prog_descriptions: List) -> None:
+def _convert_prog_descriptions(prog_descriptions: List = []) -> Dict:
```
feng-tao, line 152: For good practice, we don' use [] for default argument (https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) but instead just use None for default.

ttannis: I see, updating this before merging because I see how it can cause a bug.

docs/flask_config.md
```diff
@@ -81,53 +80,54 @@ report to provide context.
...
-       }
+    'RIGHT': {
+      "test3" : {},
+      "test2" : { "display_order": 0 }
+    },
+    'LEFT': {
+      "test1" : { "display_order": 1 },
+      "test0" : { "display_order": 0 },
+    },
+    'test4': {"display_order": 0},
```
feng-tao, line 97: so we don't need to specify other in config but only left, right?

docs/flask_config.md
```diff
@@ -81,53 +80,54 @@ report to provide context.
...
-           "quality_service": {
-               "display_order": 1
-           },
-           "doesnt_exist": {
-               "display_order": 2
-           }
-       }
+    'RIGHT': {
+      "test3" : {},
+      "test2" : { "display_order": 0 }
```
feng-tao, line 91: so test2 should match the source field in programmatic description in backend?

docs/flask_config.md
ttannis: Yes, it's meant to be backwards compatible so I didn't want to change the existing configuration shape, only add new keys for `left` and `right`.

docs/flask_config.md
ttannis: Yes, the original implementation used the `source` as a unique id. We could rethink this but the idea was that since engineers control what to ingest as the value of `source` to begin with, they can ensure they are unique.

PR #529, opened 2020-07-30
feat: clickable badge style behavior

allisonsuarez:
### Summary of Changes

Adding behavior for clickable badges, clickable badge functionality PR [here](https://github.com/lyft/amundsenfrontendlibrary/pull/519).

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/css/_variables-default.scss
+++ b/amundsen_application/static/css/_variables-default.scss
@@ -56,12 +56,20 @@ $line-height-large: 1.5 !default;
 
 // Badges
 $badge-text-color: $text-primary;
-$badge-overlay: $gray100;
 $badge-danger-color: $sunset20;
 $badge-default-color: $gray20;
 $badge-primary-color: $cyan10;
 $badge-success-color: $mint20;
 $badge-warning-color: $amber30;
+
+$badge-negative-color: $sunset20;
+$badge-neutral-color: $gray20;
+$badge-positive-color: $mint20;
+$badge-warning-color: $amber30;
+
+$badge-overlay: $gray100;
+$opacity-negative-neutral-positive-primary: .14;
+$opacity-warning: .16;
 
 // Buttons
 $btn-border-radius-base: 4px;
--- a/amundsen_application/static/css/_variables-default.scss
+++ b/amundsen_application/static/css/_variables-default.scss
@@ -70,6 +70,8 @@ $badge-warning-color: $amber30;
 $badge-overlay: $gray100;
 $opacity-negative-neutral-positive-primary: .14;
 $opacity-warning: .16;
+$opacity-pressed-negative-neutral-positive-primary: .21;
+$opacity-pressed-warning: .22;
 
 $badge-height: 20px;
 
--- a/amundsen_application/static/css/_variables-default.scss
+++ b/amundsen_application/static/css/_variables-default.scss
@@ -56,14 +56,9 @@ $line-height-large: 1.5 !default;
 
 // Badges
 $badge-text-color: $text-primary;
-$badge-danger-color: $sunset20;
-$badge-default-color: $gray20;
-$badge-primary-color: $cyan10;
-$badge-success-color: $mint20;
-$badge-warning-color: $amber30;
-
 $badge-negative-color: $sunset20;
 $badge-neutral-color: $gray20;
+$badge-primary-color: $cyan10;
 $badge-positive-color: $mint20;
 $badge-warning-color: $amber30;
 
--- a/amundsen_application/static/css/_variables-default.scss
+++ b/amundsen_application/static/css/_variables-default.scss
@@ -56,6 +56,7 @@ $line-height-large: 1.5 !default;
 
 // Badges
 $badge-text-color: $text-primary;
+$badge-overlay: $gray100;
 $badge-danger-color: $sunset20;
 $badge-default-color: $gray20;
 $badge-primary-color: $cyan10;
--- a/amundsen_application/static/css/_variables-default.scss
+++ b/amundsen_application/static/css/_variables-default.scss
@@ -63,10 +63,10 @@ $badge-positive-color: $mint20;
 $badge-warning-color: $amber30;
 
 $badge-overlay: $gray100;
-$badge-opacity-light: .14;
-$badge-opacity-dark: .16;
-$badge-pressed-light: .21;
-$badge-pressed-dark: .22;
+$badge-opacity-light: 0.14;
+$badge-opacity-dark: 0.16;
+$badge-pressed-light: 0.21;
+$badge-pressed-dark: 0.22;
 
 $badge-height: 20px;
 
--- a/amundsen_application/static/css/_variables-default.scss
+++ b/amundsen_application/static/css/_variables-default.scss
@@ -68,10 +68,10 @@ $badge-positive-color: $mint20;
 $badge-warning-color: $amber30;
 
 $badge-overlay: $gray100;
-$opacity-negative-neutral-positive-primary: .14;
-$opacity-warning: .16;
-$opacity-pressed-negative-neutral-positive-primary: .21;
-$opacity-pressed-warning: .22;
+$badge-opacity-light: .14;
+$badge-opacity-dark: .16;
+$badge-pressed-light: .21;
+$badge-pressed-dark: .22;
 
 $badge-height: 20px;
 
--- a/amundsen_application/static/css/_variables-default.scss
+++ b/amundsen_application/static/css/_variables-default.scss
@@ -71,6 +71,8 @@ $badge-overlay: $gray100;
 $opacity-negative-neutral-positive-primary: .14;
 $opacity-warning: .16;
 
+$badge-height: 20px;
+
 // Buttons
 $btn-border-radius-base: 4px;
 
--- a/amundsen_application/static/js/components/common/Badges/styles.scss
+++ b/amundsen_application/static/js/components/common/Badges/styles.scss
@@ -9,12 +9,13 @@
     padding: 0 0;
   }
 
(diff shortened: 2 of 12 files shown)


amundsen_application/static/js/components/common/Badges/styles.scss
```diff
@@ -1,10 +1,25 @@
...
   &:hover {
     cursor: pointer;
   }
-}
+  .label {
+    overflow: hidden;
+    padding: 0 0;
+  }
+  .badge-overlay-primary {
+    font-size: 14px;
```
Golodhros, line 13: Could you use $font-size-base instead?

amundsen_application/static/js/components/common/Badges/styles.scss
```diff
@@ -1,10 +1,25 @@
...
+    text-align: center;
+    display: inline-block;
+    padding: .2em .6em .3em;
```
Golodhros, line 16: We normally use pixel sizes. They are not great for accessibility, but still what we have right now.

If you could try to make them relate to the $spacer-size, as $spacer-1, $spacer-2 and so on, much better!

amundsen_application/static/js/components/common/Badges/styles.scss
```diff
@@ -1,10 +1,25 @@
...
+  }
+  .badge-overlay-primary:focus, .badge-overlay-primary:hover {
+    background-color: rgba($color: $badge-overlay, $alpha: .14);
```
Golodhros, line 19: I would store the .14 in a local variable, something like $badge-interaction-opacity or similar

amundsen_application/static/js/components/common/Badges/styles.scss
```diff
@@ -1,10 +1,25 @@
...
+    color: $text-primary;
+    height: 20px;
```
Golodhros, line 21: I would store this into a local variable, something like $badge-height or similar

amundsen_application/static/js/components/common/Flag/index.tsx
```diff
@@ -41,7 +41,9 @@ const Flag: React.SFC<FlagProps> = ({
   // https://getbootstrap.com/docs/4.1/components/badge/
   return (
     <span className={`flag label label-${labelStyle}`}>
+      <div className={`badge-overlay-${labelStyle}`}>
       {convertText(text, caseType)}
```
Golodhros, line 45: Nit: I would expect an extra indentation here

amundsen_application/static/js/components/common/Flag/index.tsx
```diff
@@ -41,7 +41,9 @@ const Flag: React.SFC<FlagProps> = ({
   // https://getbootstrap.com/docs/4.1/components/badge/
   return (
     <span className={`flag label label-${labelStyle}`}>
+      <div className={`badge-overlay-${labelStyle}`}>
```
Golodhros, line 44: Curious, if you use a <span> instead, would you be able to avoid setting `display: inline-block` in the styles above?

amundsen_application/static/js/components/common/Badges/styles.scss
allisonsuarez: The em units were used on .label so I was trying to keep that consistency

amundsen_application/static/js/components/common/Badges/styles.scss
allisonsuarez: Yeah, was planning on making all of them variables since different colors require different opacity, this is just the implementation from primary

amundsen_application/static/js/components/common/Badges/styles.scss
allisonsuarez: Sounds good, will make sure this is consistent in Flag as well

amundsen_application/static/js/components/common/Flag/index.tsx
allisonsuarez: You mean `a` as in a hyperlink tag?

allisonsuarez: I actually don't think it is needed either way, testing it out when I remove it nothing changes so might just have been leftover from my experimenting with different styles

amundsen_application/static/css/_variables-default.scss
```diff
@@ -62,6 +62,19 @@ $badge-primary-color: $cyan10;
...
+$badge-negative-color: $sunset20;
+$badge-neutral-color: $gray20;
+$badge-positive-color: $mint20;
+$badge-warning-color: $amber30;
+
+$badge-overlay: $gray100;
+$opacity-negative-neutral-positive-primary: .14;
+$opacity-warning: .16;
+$opacity-pressed-negative-neutral-positive-primary: .21;
+$opacity-pressed-warning: .22;
```
Golodhros, line 74: I would suggest something like:
$badge-opacity-light
$badge-opacity-medium
$badge-opacity-dark

I would try to get those 0.21 and 0.22 into the same one

amundsen_application/static/js/components/common/Badges/styles.scss
Golodhros: Feel free to leave then as ems then!

amundsen_application/static/js/components/common/Badges/styles.scss
```diff
@@ -1,10 +1,54 @@
...
+  [class^="badge-overlay-"] { 
+    font-size: $font-size-base;
+    text-align: center;
+    padding: .2em .6em .3em;
+
+    &:hover, &:focus {
+      color: $badge-text-color;
+      height: $badge-height;
+      width: 100%;
+      border-radius: 20px;
```
Golodhros, line 21: I would use the $badge-height variable

amundsen_application/static/js/components/common/Flag/index.tsx
Golodhros: Oh, something weird happened with the markdown, I meant a '\<span\>'

amundsen_application/static/css/_variables-default.scss
allisonsuarez: yeah I think that naming is way more scalable ahahaha

amundsen_application/static/js/components/common/Badges/styles.scss
```diff
@@ -1,10 +1,51 @@
...
+    }
+
+    &:focus {
```
ttannis, line 24: tldr; I think this is an issue we can come back to

In practice the focus effect applies cases when components are keyboard accessible, i.e. being able to use the `Tab` key to navigate to this component and the `Enter` key when it's focused to trigger the action. However the way this component is written it is currently not keyboard accessible. If you try it out and you'll see you won't be able to reach this component hitting the `Tab` key.  And let's say it were accessible as written, I think the focus will actually end up on the parent element as opposed to the overlay, so then this css wouldn't apply regardless. 

More recently Eng has been trying to do a better job with writing accessible application code (but internally, this hasn't been a priority for the product), so what we can do is to have a TODO to update this component to be accessible. And then we would fix the focus state. Any thoughts or questions?

amundsen_application/static/css/_variables-default.scss
```diff
@@ -62,6 +62,19 @@ $badge-primary-color: $cyan10;
 $badge-success-color: $mint20;
 $badge-warning-color: $amber30;
 
+$badge-negative-color: $sunset20;
+$badge-neutral-color: $gray20;
+$badge-positive-color: $mint20;
+$badge-warning-color: $amber30;
```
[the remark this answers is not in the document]
allisonsuarez: The reason why I had not changed it is because this is the change that would break any custom configuration an implementation has for badges. I can go ahead with the change but I think we might need to let people know in case they are using it.

amundsen_application/static/js/components/common/Badges/styles.scss
allisonsuarez: I understand, I was using the focused option on inspect so I didn't try reaching it but I was wondering if the way the components were nester would be an issue. I found a potential solution just now that leverages children selection so when the .clickable-badge is focus we can change the style of the .badge-overlay-. I will give it a quick try see how that goes.

amundsen_application/static/css/_variables-default.scss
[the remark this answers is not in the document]
ttannis: You should be able to change them because the style users are supposed to set is a enumerable value, so `BadgeStyle.DANGER`. So although we might not be able to remove that enum itself, it's string value defined [here](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/config/config-types.ts#L136-L143) can be updated to `'negative'` for all usages and that wouldn't break any existing configurations, but allow us to clean up our style variables.

amundsen_application/static/js/components/common/Badges/styles.scss
allisonsuarez: Will create a separate PR to decouple Flag from clickable badge and make the component keyboard accessible.

amundsen_application/static/css/_variables-default.scss
```diff
@@ -56,12 +56,20 @@ $line-height-large: 1.5 !default;
 
 // Badges
 $badge-text-color: $text-primary;
-$badge-danger-color: $sunset20;
-$badge-default-color: $gray20;
+$badge-negative-color: $sunset20;
```
ttannis, line 59: Last thing, we do have some places where we have hardcoded `labelStyle: 'danger'`, vs using `BadgeStyle.DANGER`.  If you search the `components` folder repo for `'danger'`, `'success'`, and `'default'` you will see a few places where we have to use the proper method of using `BadgeStyle`.

PR #533, opened 2020-08-04
fix: Uneditable descriptions with no value or possible actions should not show header

dikshathakur3119:
### Summary of Changes

For un-editable column descriptions that have no value and have no edit url, our current UI shows the `Description` header when the column is expanded. However there is nothing to show here and nothing the user will ever be able to interact with (e.g. no click to edit in url and no `Add Description`). As a result, we shouldn't show the header.

### Tests
Will be adding tests

### Documentation

NA

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/TableDetail/ColumnListItem/index.spec.tsx
+++ b/amundsen_application/static/js/components/TableDetail/ColumnListItem/index.spec.tsx
@@ -124,133 +124,137 @@ describe('ColumnListItem', () => {
       expect(wrapper.find('.column-dropdown').exists()).toBe(false);
     });
 
-    it('renders column stats and editable text when expanded', () => {
-      instance.setState({ isExpanded: true });
-      const newWrapper = shallow(instance.render());
-      expect(newWrapper.find('.expanded-content').exists()).toBe(true);
-      expect(newWrapper.find(ColumnDescEditableText).exists()).toBe(true);
-      expect(newWrapper.find(ColumnStats).exists()).toBe(true);
-    });
-
-    it('renders EditableSection with correct description, edit text and Url when expanded', () => {
-      instance.setState({ isExpanded: true });
-      const newWrapper = shallow(instance.render());
-      expect(newWrapper.find(EditableSection).exists()).toBe(true);
-      const editableSection = newWrapper.find(EditableSection);
-      expect(editableSection.props()).toMatchObject({
-        title: 'Description',
-        readOnly: !props.data.is_editable,
-        editText: props.editText,
-        editUrl: props.editUrl,
+    describe('when expanded', () => {
+      it('renders column stats and editable text when expanded', () => {
+        instance.setState({ isExpanded: true });
+        const newWrapper = shallow(instance.render());
+        expect(newWrapper.find('.expanded-content').exists()).toBe(true);
+        expect(newWrapper.find(EditableSection).exists()).toBe(true);
+        expect(newWrapper.find(ColumnDescEditableText).exists()).toBe(true);
+        expect(newWrapper.find(ColumnStats).exists()).toBe(true);
       });
-    });
 
-    it('renders EditableSection with non-empty description, empty edit text and url for editable column description when expanded', () => {
-      const { props, wrapper } = setup({
-        editText: '',
-        editUrl: '',
+      it('renders EditableSection with non-empty description, edit text and url when expanded', () => {
+        instance.setState({ isExpanded: true });
+        const newWrapper = shallow(instance.render());
+        const editableSection = newWrapper.find(EditableSection);
+        expect(editableSection.props()).toMatchObject({
+          title: 'Description',
+          readOnly: !props.data.is_editable,
+          editText: props.editText,
+          editUrl: props.editUrl,
+        });
       });
-      const instance = wrapper.instance();
-      instance.setState({ isExpanded: true });
-      const newWrapper = shallow(instance.render());
-      expect(newWrapper.find(EditableSection).exists()).toBe(true);
-      const editableSection = newWrapper.find(EditableSection);
-      expect(editableSection.props()).toMatchObject({
-        title: 'Description',
-        readOnly: !props.data.is_editable,
-        editText: props.editText,
-        editUrl: props.editUrl,
-      });
-    });
 
-    it('renders EditableSection with empty description, edit text and url for editable column description when expanded', () => {
-      const { props, wrapper } = setup({
-        data: {
-          name: 'test_column_name',
-          description: '',
-          is_editable: true,
-          col_type: 'varchar(32)',
-          stats: [
-            {
-              end_epoch: 1571616000,
-              start_epoch: 1571616000,
-              stat_type: 'count',
-              stat_val: '12345',
+      it('renders EditableSection with non-empty description, empty edit text and url when expanded', () => {
+        const { props, wrapper } = setup({
+          editText: '',
+          editUrl: '',
+        });
+        const instance = wrapper.instance();
+        instance.setState({ isExpanded: true });
+        const newWrapper = shallow(instance.render());
+        expect(newWrapper.find(EditableSection).exists()).toBe(true);
+        const editableSection = newWrapper.find(EditableSection);
+        expect(editableSection.props()).toMatchObject({
+          title: 'Description',
+          readOnly: !props.data.is_editable,
+          editText: props.editText,
+          editUrl: props.editUrl,
+        });
+      });
+
+      describe('when empty description', () => {
+        it('renders EditableSection with empty edit text and url for editable column description when expanded', () => {
+          const { props, wrapper } = setup({
+            data: {
+              name: 'test_column_name',
+              description: '',
+              is_editable: true,
+              col_type: 'varchar(32)',
+              stats: [
+                {
+                  end_epoch: 1571616000,
+                  start_epoch: 1571616000,
+                  stat_type: 'count',
+                  stat_val: '12345',
+                },
+              ],
             },
-          ],
-        },
-        editText: '',
-        editUrl: '',
-      });
-      const instance = wrapper.instance();
-      instance.setState({ isExpanded: true });
-      const newWrapper = shallow(instance.render());
-      expect(newWrapper.find(EditableSection).exists()).toBe(true);
-      const editableSection = newWrapper.find(EditableSection);
-      expect(editableSection.props()).toMatchObject({
-        title: 'Description',
-        readOnly: !props.data.is_editable,
-        editText: props.editText,
-        editUrl: props.editUrl,
-      });
-    });
+            editText: '',
+            editUrl: '',
+          });
+          const instance = wrapper.instance();
+          instance.setState({ isExpanded: true });
+          const newWrapper = shallow(instance.render());
+          expect(newWrapper.find(EditableSection).exists()).toBe(true);
+          const editableSection = newWrapper.find(EditableSection);
+          expect(editableSection.props()).toMatchObject({
+            title: 'Description',
+            readOnly: !props.data.is_editable,
+            editText: props.editText,
+            editUrl: props.editUrl,
+          });
+        });
 
-    it('does not render EditableSection with empty description, edit text and url for non-editable column description when expanded', () => {
-      const { wrapper } = setup({
-        data: {
-          name: 'test_column_name',
-          description: '',
-          is_editable: false,
-          col_type: 'varchar(32)',
-          stats: [
-            {
-              end_epoch: 1571616000,
-              start_epoch: 1571616000,
-              stat_type: 'count',
-              stat_val: '12345',
+        it('does not render EditableSection with empty edit text and url for non-editable column description when expanded', () => {
+          const { wrapper } = setup({
+            data: {
+              name: 'test_column_name',
+              description: '',
+              is_editable: false,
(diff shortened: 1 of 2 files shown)


amundsen_application/static/js/components/TableDetail/ColumnListItem/index.tsx
```diff
@@ -73,6 +73,19 @@ export class ColumnListItem extends React.Component<
     e.stopPropagation();
   };
 
+  renderDescription = () => {
```
ttannis, line 76: nit: mind changing the name and return type to `shouldRenderDescription(): boolean`. This is to make it more intuitive what this does.

amundsen_application/static/js/components/TableDetail/ColumnListItem/index.tsx
```diff
@@ -73,6 +73,19 @@ export class ColumnListItem extends React.Component<
...
+    let expand: boolean = true;
+    if (
```
ttannis, line 78: I assume the logic you are aiming for is:
1. Return `true` if there is a description
2. Return `false` if no `description`, no `editText`, or no `editUrl`
3. Return `true` is neither apply

If so, this can be simplified as:
```
const { data, editText, editUrl } = this.props; // just for code cleanliness help
if (data.description) { // in JS this checks if a string is "truthy", which means it exists AND is not empty-string
    return true;
}
if (!editText && !editUrl) { // in JS this checks if a string is "falsy", which means it does not exist OR is empty-string
    return false;
}
return true;
```

dikshathakur3119: Should I call this logic from toggleEdit, so that toggleEdit does not work on click? Currently it toggles and size component changes.

amundsen_application/static/js/components/TableDetail/ColumnListItem/index.tsx
dikshathakur3119: yeah sure, will do

amundsen_application/static/js/components/TableDetail/ColumnListItem/index.tsx
dikshathakur3119: I got my answer, we cannot do that since a column may have stats even if it does not have description so toggle expand is required to run

ttannis: Oh I see what you mean. But here are my suggestions after taking a second look:
1. Can you instead heck for whether or not we should render something shouldn't just wrap the `EditableSection`, rather the parent <div> as well...no use rendering a parent if it won't have a child. 
2. The size changing is acceptable. We are also technically supposed to have `ColumnStats` component show up. However currently at Lyft we aren't showing column stats, and the current `ColumnStats` UI should nothing if there are no stats. So it's the `ColumnStats` rendering that would have to be improved I think if we didn't want any extra space.

dikshathakur3119: I did not understand your first point completely . But on the same lines , to stop rendering of parent if there is no child component I have put check inside toggleExpand function. Let me know if that is okay. `ColumnStats` rendering returns `null` if there are no stats.

amundsen_application/static/js/components/TableDetail/ColumnListItem/index.tsx
```diff
@@ -172,19 +185,21 @@ export class ColumnListItem extends React.Component<
...
-                  <ColumnDescEditableText
-                    columnIndex={this.props.index}
-                    editable={metadata.is_editable}
-                    maxLength={getMaxLength('columnDescLength')}
-                    value={metadata.description}
-                  />
-                </EditableSection>
+                {this.shouldRenderDescription() && (
+                  <EditableSection
+                    title="Description"
```
Golodhros, line 190: Could you move this string into a constant? Something like EDITABLE_SECTION_TITLE or similar.

amundsen_application/static/js/components/TableDetail/ColumnListItem/index.tsx
```diff
@@ -59,7 +59,9 @@ export class ColumnListItem extends React.Component<
         label: `${metadata.name} ${metadata.col_type}`,
       });
     }
-    this.setState({ isExpanded: !this.state.isExpanded });
+    if (this.shouldRenderDescription() || this.props.data.stats.length !== 0) {
```
Golodhros, line 62: Nit: I would move the line 55 up one position and use here `|| metadata.stats.length`

amundsen_application/static/js/components/TableDetail/ColumnListItem/index.spec.tsx
```diff
@@ -132,7 +132,113 @@ describe('ColumnListItem', () => {
       expect(newWrapper.find(ColumnStats).exists()).toBe(true);
     });
 
-    it('renders EditableSection with correct edit text and Url when expanded', () => {
+    it('renders EditableSection with correct description, edit text and Url when expanded', () => {
```
Golodhros, line 135: Nit: I would create a describe clause `describe('when expanded', () => ...` and nest your test and the previous one inside of it.

amundsen_application/static/js/components/TableDetail/ColumnListItem/index.spec.tsx
```diff
@@ -132,7 +132,113 @@ describe('ColumnListItem', () => {
...
+      const newWrapper = shallow(instance.render());
+      expect(newWrapper.find(EditableSection).exists()).toBe(true);
+      const editableSection = newWrapper.find(EditableSection);
+      expect(editableSection.props()).toMatchObject({
+        title: 'Description',
+        readOnly: !props.data.is_editable,
+        editText: props.editText,
+        editUrl: props.editUrl,
+      });
+    });
```
Golodhros, line 146: I would probably split this into two tests

amundsen_application/static/js/components/TableDetail/ColumnListItem/index.spec.tsx
```diff
@@ -132,7 +132,113 @@ describe('ColumnListItem', () => {
...
+
+    it('renders EditableSection with empty description, edit text and url for editable column description when expanded', () => {
```
Golodhros, line 166: Nit: I would nest this test and the next within a describe clause that says `describe('when empty description', ...`

dikshathakur3119:
> Not a fan of the way the tests were originally setup for that page, but your code is consistent so looks good to me.
> 
> Only left some minor comments.

Thanks Marcos, I will make suggested changes. I was not sure what is a good way to write unit test cases when I want to test different values of `props` in each test case, so I just followed from some references I could find.

ttannis:
> > Not a fan of the way the tests were originally setup for that page, but your code is consistent so looks good to me.
> > Only left some minor comments.
> 
> Thanks Marcos, I will make suggested changes. I was not sure what is a good way to write unit test cases when I want to test different values of `props` in each test case, so I just followed from some references I could find.

@Golodhros could you comment on how you think we could better structure a unit test for what is added here, regardless of what already exists? 

I think this context would be good to share to help inform on different ways to do things (this goes for all things frontend), so that everyone has more understanding and context if/when we shift towards one practice vs. another.

Golodhros:
> > > Not a fan of the way the tests were originally setup for that page, but your code is consistent so looks good to me.
> > > Only left some minor comments.
> > 
> > 
> > Thanks Marcos, I will make suggested changes. I was not sure what is a good way to write unit test cases when I want to test different values of `props` in each test case, so I just followed from some references I could find.
> 
> @Golodhros could you comment on how you think we could better structure a unit test for what is added here, regardless of what already exists?
> 
> I think this context would be good to share to help inform on different ways to do things (this goes for all things frontend), so that everyone has more understanding and context if/when we shift towards one practice vs. another.

Sure!
In a nutshell, I would:
- Use enzyme's "mount" instead of "shallow", so that we render the components all the way down
- Done that ^, I would avoid checking the props we pass down, and instead I would check that the proper elements are rendered on the "DOM". 
- I would avoid setting the state directly, and instead trigger a click on the element that expands the element.

It is not that these tests are completely wrong. They do their work, but they do it on a really specific way (Imagine we change the property names of the chilld component, these tests would break even if everything kept on working fine). 

In general, any setting of state on enzyme wrappers should raise red flags. But also, I am aware that the way we have most of our tests in the application is not aligned with what I think is best, so I approve this PR!

ttannis:
> > > > Not a fan of the way the tests were originally setup for that page, but your code is consistent so looks good to me.
> > > > Only left some minor comments.
> > > 
> > > 
> > > Thanks Marcos, I will make suggested changes. I was not sure what is a good way to write unit test cases when I want to test different values of `props` in each test case, so I just followed from some references I could find.
> > 
> > 
> > @Golodhros could you comment on how you think we could better structure a unit test for what is added here, regardless of what already exists?
> > I think this context would be good to share to help inform on different ways to do things (this goes for all things frontend), so that everyone has more understanding and context if/when we shift towards one practice vs. another.
> 
> Sure!
> In a nutshell, I would:
> 
> * Use enzyme's "mount" instead of "shallow", so that we render the components all the way down
> * Done that ^, I would avoid checking the props we pass down, and instead I would check that the proper elements are rendered on the "DOM".
> * I would avoid setting the state directly, and instead trigger a click on the element that expands the element.
> 
> It is not that these tests are completely wrong. They do their work, but they do it on a really specific way (Imagine we change the property names of the child component, these tests would break even if everything kept on working fine).
> 
> In general, any setting of state on enzyme wrappers should raise red flags. But also, I am aware that the way we have most of our tests in the application is not aligned with what I think is best, so I approve this PR!

Thanks. @dikshathakur3119 I was asking this out of my own curiosity no action needed for you PR. Marcos we haven't put much work into unit tests since a previous push for improved coverage, so I may experiment with writing new unit tests in that manner as well. We can find a balance between consistency and trending towards improvements.

amundsen_application/static/js/components/TableDetail/ColumnListItem/index.tsx
ttannis: +1

dikshathakur3119:
👀 @ttannis @Golodhros @allisonsuarez

amundsen_application/static/js/components/TableDetail/ColumnListItem/index.spec.tsx
```diff
@@ -124,27 +124,137 @@ describe('ColumnListItem', () => {
...
 
-    it('renders column stats and editable text when expanded', () => {
-      instance.setState({ isExpanded: true });
-      const newWrapper = shallow(instance.render());
-      expect(newWrapper.find('.expanded-content').exists()).toBe(true);
-      expect(newWrapper.find(ColumnDescEditableText).exists()).toBe(true);
-      expect(newWrapper.find(ColumnStats).exists()).toBe(true);
-    });
+    describe('when expanded', () => {
+      it('renders column stats and editable text when expanded', () => {
```
Golodhros, line 128: Nit: you could remove the "when expanded" part at the end of this text

amundsen_application/static/js/components/TableDetail/ColumnListItem/index.spec.tsx
```diff
@@ -124,27 +124,137 @@ describe('ColumnListItem', () => {
...
+            editText: '',
+            editUrl: '',
+          });
+          const instance = wrapper.instance();
+          instance.setState({ isExpanded: true });
+          const newWrapper = shallow(instance.render());
+          expect(newWrapper.find(EditableSection).exists()).toBe(false);
+        });
+
+        it('renders EditableSection with non-empty edit text and url for non-editable column description when expanded', () => {
```
Golodhros, line 225: I guess this one shouldn't be nested inside the "when empty description" describe

amundsen_application/static/js/components/TableDetail/ColumnListItem/index.tsx
```diff
@@ -51,15 +52,17 @@ export class ColumnListItem extends React.Component<
...
-      const metadata = this.props.data;
       logClick(e, {
         target_id: `column::${metadata.name}`,
         target_type: 'column stats',
         label: `${metadata.name} ${metadata.col_type}`,
       });
     }
-    this.setState({ isExpanded: !this.state.isExpanded });
+    if (this.shouldRenderDescription() || metadata.stats.length !== 0) {
+      this.setState({ isExpanded: !this.state.isExpanded });
```
Golodhros, line 64: Oh, sorry, I missed this one before.

When we set a state that depends on the previous state, we should always use the functional way:
```
this.setState(prevState => ({
  isExpanded: !prevState.isExpanded
}));
```
This is because `setState` is asynchronous, and it could result on unexpected state being saved.
Check the docs for more info: https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

amundsen_application/static/js/components/TableDetail/ColumnListItem/index.spec.tsx
dikshathakur3119: In this case also we have empty column description, but user have option to edit and add some description, hence I have put it inside "when empty description" describe.

amundsen_application/static/js/components/TableDetail/ColumnListItem/index.tsx
dikshathakur3119: Okay, thank you!

PR #534, opened 2020-08-04
build: use static CSS and JS filenames in dev

dorianj:
In production, it's helpful to use content addressing to guarantee cache busting
when publishing changes.

However, in development, because the server serves a static copy of the
index.html, that means that changing the Javascript bundle or CSS requires an
otherwise-unnecessary restart of the server.

This change makes the CSS and JS assets use fixed filenames so that a server
restart isn't needed.

Downside of this approach is there's now a dev/prod disparity and the config is
more spread out between files. The positive is we don't need to modify the
semantics of the server at all.

### Tests

AFAICT, there are no selenium-style integration tests, so I'm not aware of an established pattern to test this behavior. Happy to add tests if there's a pathway, though!

### Documentation

The docs don't mention this stuff right now, so this should mean we don't need to change the docs.

### CheckList

### DIFF
--- a/amundsen_application/static/webpack.common.ts
+++ b/amundsen_application/static/webpack.common.ts
@@ -100,11 +100,7 @@ const config: webpack.Configuration = {
   plugins: [
     new CleanWebpackPlugin(),
     new MomentLocalesPlugin(), // To strip all locales except “en”
-    new MiniCssExtractPlugin({
-      filename: '[name].[contenthash].css',
-    }),
     ...htmlWebpackPluginConfig,
-    // new BundleAnalyzerPlugin(),   // Uncomment to analyze the production bundle on local
   ],
   optimization: {
     moduleIds: 'hashed',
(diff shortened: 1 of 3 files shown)


dorianj:
Lint failure is real, will fix once get feedback that this approach is sane

ttannis:
> Lint failure is real, will fix once get feedback that this approach is sane

Thanks, will approve when fixed.

ttannis:
@Golodhros these changes look good to me but wanted your input in case there were specific cases that you considered when last updating these files.

Golodhros:
Looks good to me!

Thanks for the contribution @dorianj !!

dorianj:
@ttannis great! I force pushed to fix the lint problems.

jornh:
@dorianj nice stuff! 

Am I right in assuming this almost or fully resolves the feature request in https://github.com/lyft/amundsen/issues/392 if you’re doing a  `npm run watch`?

dorianj:
@jornh yes, it does fulfill that ticket. It's not full hot-reloading: I'm not sure what would be needed to make that work, I've never worked on a project where that actually worked end-to-end. But it does remove requirement to restart server, so will close that ticket.

PR #538, opened 2020-08-05
chore: Fixes issues with no-unused-expressions ESLint rule

Golodhros:
### Summary of Changes
Fixes issues related to the [no-unused-expressions rule](https://eslint.org/docs/2.0.0/rules/no-unused-expressions) of ESLint
Removes the rule warning (now it will trigger an error)

### Tests
Updated one test

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/Tags/TagInput/index.tsx
+++ b/amundsen_application/static/js/components/Tags/TagInput/index.tsx
@@ -179,9 +179,12 @@ class TagInput extends React.Component<TagInputProps, TagInputState> {
 
   toggleTag = (event, tagName) => {
     const element = event.currentTarget;
-    element.classList.contains('selected')
-      ? element.classList.remove('selected')
-      : element.classList.add('selected');
+
+    if (element.classList.contains('selected')) {
+      element.classList.remove('selected');
+    } else {
+      element.classList.add('selected');
+    }
 
     if (!this.batchEditSet.hasOwnProperty(tagName)) {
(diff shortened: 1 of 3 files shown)


amundsen_application/static/js/components/Tags/TagInput/index.tsx
```diff
@@ -179,9 +179,12 @@ class TagInput extends React.Component<TagInputProps, TagInputState> {
 
   toggleTag = (event, tagName) => {
     const element = event.currentTarget;
-    element.classList.contains('selected')
-      ? element.classList.remove('selected')
-      : element.classList.add('selected');
+
```
ttannis, line 182: wait, which rule doesn't prefer the syntax in this files? I'm a bit confused there should also be similar binary and ternary statements elsewhere in the app as well.

Golodhros: Yeah, I got a bit confused at first too.

The main difference is that usually we use ternaries to set the value of a variable or even to return a value.
However, in this case, we don't do any of that, we just call a function.

In the rule definition (https://eslint.org/docs/2.0.0/rules/no-unused-expressions) there is a mention about one option that would allow this case, it is the 'allowTernary' option. If you feel strongly about keeping this ternary, we could switch that option on and keep it.

ttannis: I don't feel strongly, only wanted to understand. Given how JS is interpreted I see how these statements result in what would be considered an unused expression.

PR #543, opened 2020-08-07
fix: Create separate reducer for last indexed value

dikshathakur3119:
### Summary of Changes

Currently single reducer is handling states related to tableMetadata and lastIndexed values. To optimize, we decoupled tableMetadata and lastIndexed values and created different reducer to handle lastIndexed state.

### Tests
Added unit tests

### Documentation

NA

### CheckList

### DIFF
--- a/amundsen_application/static/js/ducks/lastIndexed/api/v0.ts
+++ b/amundsen_application/static/js/ducks/lastIndexed/api/v0.ts
@@ -10,6 +10,14 @@ export function getLastIndexed() {
   return axios
     .get(`${API_PATH}/get_last_indexed`)
     .then((response: AxiosResponse<LastIndexedAPI>) => {
-      return response.data.timestamp;
+      const { data } = response;
+
+      return data.timestamp;
+    })
+    .catch((e) => {
+      const timestamp = null;
+      return Promise.reject({
+        timestamp,
+      });
     });
 }
--- a/amundsen_application/static/js/ducks/lastIndexed/api/v0.ts
+++ b/amundsen_application/static/js/ducks/lastIndexed/api/v0.ts
@@ -1,0 +1,15 @@
+import axios, { AxiosResponse, AxiosError } from 'axios';
+
+export const API_PATH = '/api/metadata/v0';
+
+type MessageAPI = { msg: string };
+
+export type LastIndexedAPI = { timestamp: string } & MessageAPI;
+
+export function getLastIndexed() {
+  return axios
+    .get(`${API_PATH}/get_last_indexed`)
+    .then((response: AxiosResponse<LastIndexedAPI>) => {
+      return response.data.timestamp;
+    });
+}
--- a/amundsen_application/static/js/ducks/lastIndexed/reducer.ts
+++ b/amundsen_application/static/js/ducks/lastIndexed/reducer.ts
@@ -1,0 +1,53 @@
+import {
+  GetLastIndexed,
+  GetLastIndexedRequest,
+  GetLastIndexedResponse,
+} from './types';
+
+export interface LastIndexedReducerState {
+  lastIndexed: number;
+}
+export const initialState: LastIndexedReducerState = {
+  lastIndexed: null,
+};
+
+/* ACTIONS */
+export function getLastIndexed(): GetLastIndexedRequest {
+  return { type: GetLastIndexed.REQUEST };
+}
+
+export function getLastIndexedFailure(): GetLastIndexedResponse {
+  return { type: GetLastIndexed.FAILURE };
+}
+
+export function getLastIndexedSuccess(
+  lastIndexedEpoch: number
+): GetLastIndexedResponse {
+  return {
+    type: GetLastIndexed.SUCCESS,
+    payload: {
+      lastIndexedEpoch,
+    },
+  };
+}
+
+/* REDUCER */
+export default function reducer(
+  state: LastIndexedReducerState = initialState,
+  action
+): LastIndexedReducerState {
+  switch (action.type) {
+    case GetLastIndexed.REQUEST:
+      return initialState;
+    case GetLastIndexed.SUCCESS:
+      return {
+        lastIndexed: (<GetLastIndexedResponse>action).payload.lastIndexedEpoch,
+      };
+    case GetLastIndexed.FAILURE:
+      return {
+        lastIndexed: null,
+      };
+    default:
+      return state;
+  }
+}
--- a/amundsen_application/static/js/ducks/lastIndexed/tests/index.spec.ts
+++ b/amundsen_application/static/js/ducks/lastIndexed/tests/index.spec.ts
@@ -1,0 +1,97 @@
+import { testSaga } from 'redux-saga-test-plan';
+
+import * as API from '../api/v0';
+
+import reducer, {
+  getLastIndexed,
+  getLastIndexedFailure,
+  getLastIndexedSuccess,
+  initialState,
+  LastIndexedReducerState,
+} from '../reducer';
+
+import { getLastIndexedWatcher, getLastIndexedWorker } from '../sagas';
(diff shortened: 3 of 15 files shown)


dikshathakur3119:
I will be removing all the commented code in next push (code cleanup).

amundsen_application/static/js/ducks/lastIndexed/api/v0.ts
```diff
@@ -0,0 +1,15 @@
...
+type MessageAPI = { msg: string };
+
+export type LastIndexedAPI = { timestamp: string } & MessageAPI;
+
+export function getLastIndexed() {
+  return axios
+    .get(`${API_PATH}/get_last_indexed`)
+    .then((response: AxiosResponse<LastIndexedAPI>) => {
+      return response.data.timestamp;
+    });
```
Golodhros, line 14: I would add a catch clause here and pass down the statusCode and maybe the errorMessage>

You can see an example in the dashboard code: https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/ducks/dashboard/api/v0.ts#L25

amundsen_application/static/js/ducks/lastIndexed/reducer.ts
```diff
@@ -0,0 +1,53 @@
...
+/* ACTIONS */
+export function getLastIndexed(): GetLastIndexedRequest {
+  return { type: GetLastIndexed.REQUEST };
+}
+
+export function getLastIndexedFailure(): GetLastIndexedResponse {
+  return { type: GetLastIndexed.FAILURE };
+}
+
+export function getLastIndexedSuccess(
```
Golodhros, line 23: I would create an export an interface for the payload.
Check here an example: https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/ducks/dashboard/types.ts#L23

amundsen_application/static/js/ducks/lastIndexed/tests/index.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+  describe('reducer', () => {
+    let testState: LastIndexedReducerState;
+    beforeAll(() => {
+      testState = initialState;
+    });
+
+    it('should return the existing state if action is not handled', () => {
+      expect(reducer(testState, { type: 'INVALID.ACTION' })).toEqual(testState);
+    });
+
```
Golodhros, line 51: We are missing a test for the REQUEST, check here for an example: https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/ducks/dashboard/tests/reducer.spec.ts#L24

amundsen_application/static/js/ducks/rootSaga.ts
```diff
@@ -44,13 +44,16 @@ import { updateTableOwnerWatcher } from './tableMetadata/owners/sagas';
 import {
   getTableDataWatcher,
   getColumnDescriptionWatcher,
-  getLastIndexedWatcher,
+  // getLastIndexedWatcher,
```
Golodhros, line 47: Let's clean this up

amundsen_application/static/js/ducks/rootSaga.ts
```diff
@@ -100,12 +103,13 @@ export default function* rootSaga() {
     // TableDetail
     getTableDataWatcher(),
     getColumnDescriptionWatcher(),
-    getLastIndexedWatcher(),
     getPreviewDataWatcher(),
     getTableDescriptionWatcher(),
     updateColumnDescriptionWatcher(),
     updateTableDescriptionWatcher(),
     updateTableOwnerWatcher(),
+    // LastIndexed
```
Golodhros, line 111: Nit: I would say these comments are not necessary

amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
```diff
@@ -173,13 +173,13 @@ export function updateColumnDescription(
...
-      return response.data.timestamp;
-    });
-}
+// export function getLastIndexed() {
+//   return axios
+//     .get(`${API_PATH}/get_last_indexed`)
+//     .then((response: AxiosResponse<LastIndexedAPI>) => {
+//       return response.data.timestamp;
+//     });
+// }
```
Golodhros, line 182: Let's remove all commented code, thanks!

dikshathakur3119: yes I will be removing all the comments since those are related to old code.

amundsen_application/static/js/ducks/lastIndexed/api/v0.ts
```diff
@@ -0,0 +1,15 @@
+import axios, { AxiosResponse, AxiosError } from 'axios';
+
+export const API_PATH = '/api/metadata/v0';
```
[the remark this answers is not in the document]
dikshathakur3119: Yes I can do for all the reducers using same API_PATH

amundsen_application/static/js/ducks/rootSaga.ts
dikshathakur3119: Done just for consistency purposes. I can remove them though. Should I remove all the comments in this file or just new ones added in this PR?

dikshathakur3119:
> I'll have to spend some time to look into this but wanted to comment on the reducer structure. One thing that might become more obvious here is that this stuff does have a lot of boiler plate.
> 
> You've followed existing precedents in that so far we've only had entire objects to store in the global state, but we don't particularly need that here to store a single value, e.g. have `lastIndexed: { lastIndexed: ... }`. I'll see what else I can suggest.

Yes you are right, I decided to go with same structure because I thought it will keep this object more flexible and consistent. In case we need to add new properties to it in future e.g error code, error message, etc, it would be easier to modify it.

ttannis:
> > I'll have to spend some time to look into this but wanted to comment on the reducer structure. One thing that might become more obvious here is that this stuff does have a lot of boiler plate.
> > You've followed existing precedents in that so far we've only had entire objects to store in the global state, but we don't particularly need that here to store a single value, e.g. have `lastIndexed: { lastIndexed: ... }`. I'll see what else I can suggest.
> 
> Yes you are right, I decided to go with same structure because I thought it will keep this object more flexible and consistent. In case we need to add new properties to it in future e.g error code, error message, etc, it would be easier to modify it. But if it is too much of code for updating a single value, I can totally make it more straight forward.

I wasn't able to think of something that doesn't require rethinking a lot of concepts outside of the scope of this PR. The issue of it being a lot of boilerplate code is something that we'd have to rethink this entire layer to handle, but I hope to see if we can eventually make this kind of work less cumbersome.

amundsen_application/static/js/ducks/rootSaga.ts
ttannis: I'd prefer to leave them in.  I believe these comments aren't necessary for me, but I put them in because since we happened to organize this list in order of section vs. alphabetically I figured the comments helps give some context.

dikshathakur3119: Okay, left them as it is.

amundsen_application/static/js/ducks/lastIndexed/api/v0.ts
dikshathakur3119: Added catch phrase. But not included error code and message since currently we are not using error code and message if we do not put error message in Footer if we do not get last_indexed value.

amundsen_application/static/js/ducks/lastIndexed/api/v0.ts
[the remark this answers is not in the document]
dikshathakur3119: There are several files included so I will make these changes as part of another PR.

PR #546, opened 2020-08-07
feat: Basic unstyled Card skeleton with shimmer loader

Golodhros:
### Summary of Changes
Adds a Card component
Not styled yet
Includes shimmering loader

### Tests
Added tests

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/common/Card/index.spec.tsx
+++ b/amundsen_application/static/js/components/common/Card/index.spec.tsx
@@ -1,0 +1,122 @@
+import * as React from 'react';
+import { mount } from 'enzyme';
+
+import Card, { CardProps } from '.';
+
+const setup = (propOverrides?: Partial<CardProps>) => {
+  const props = {
+    title: 'test title',
+    ...propOverrides,
+  };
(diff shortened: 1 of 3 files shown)


amundsen_application/static/js/components/common/Card/index.spec.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { mount } from 'enzyme';
+
+import Card, { CardProps } from '.';
+
+const setup = (propOverrides?: Partial<CardProps>) => {
+  const props = {
+    title: 'test title',
+    ...propOverrides,
+  };
+  const wrapper = mount<typeof Card>(<Card {...props} />);
```
ttannis, line 11: Is there some TypeScript benefit for  `<typeof Card>` vs `<Card>`?

Golodhros: Not really sure. In this case if I take out the 'typeof', I get an error: "Card' refers to a value, but is being used as a type here. Did you mean 'typeof Card'?"

I saw that in other cases, we don't actually add this type here at all. 🤷

PR #547, opened 2020-08-07
fix: Update EditableSection interaction

ttannis:
### Summary of Changes

For our desired interaction, we only need to call `preventDefault` when the rendering edit button (`readOnly=false`), and not when rendering the edit link (`readOnly=true`). 

### Tests

N/A

### Documentation

N/A

### CheckList


dikshathakur3119:
I see only buttons inside `label` tags behave strangely and responds to all events from parent component while `OverlayTrigger` is working normally (in case of un-editable tables). So we need not to stop propagation when click is not on a button.

PR #548, opened 2020-08-07
chore: Re-useable OwnerEditor

ttannis:
### Summary of Changes

This PR further updates `OwnerEditor` to be reusable for different resources, then introduces it for use on `DashboardPage` to render a readOnly `OwnerEditor`. 

1. Moved `TableDetail/OwnerEditor` to `common/OwnerEditor`. Added `resourceType` a new property, and disconnected the generic component from Redux.
2. Introduced `TableOwnerEditor` which is an `OwnerEditor` connected to the table metadata state.
3. Introduced `DashboardOwnerEditor` which is a readOnly `OwnerEditor` connected to the dashboard metadata state.
4. Removed code `DashboardPage` that copied the owner editor UI.

To the end user, the functionality and the UI remains the same from this PR.

### Tests

Updates/added tests + some mock data for users

### Documentation

N/A

### CheckList


amundsen_application/static/js/components/DashboardPage/DashboardOwnerEditor/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+      isExternalLink = false;
+      profileLink = `/user/${user_id}?source=${DASHBOARD_OWNER_SOURCE}`;
+    }
+    obj[user_id] = {
+      label: display_name,
+      link: profileLink,
+      isExternal: isExternalLink,
+    };
+    return obj;
+  }, {});
```
Golodhros, line 32: This logic looks like a "selector". It is usually recommended to extract into an independent function these types of functions which we use to take info from the state.

From the [Redux Styleguide](https://redux.js.org/style-guide/style-guide/#use-selector-functions-to-read-from-store-state):
""Selector functions" are a powerful tool for encapsulating reading values from the Redux store state and deriving further data from those values. In addition, libraries like Reselect enable creating memoized selector functions that only recalculate results when the inputs have changed, which is an important aspect of optimizing performance."

Creating a different function makes it easy to test and reuse. Here is a [short video](https://www.youtube.com/watch?v=frT3to2ACCw) about selectors and [an article](https://medium.com/@matthew.holman/what-is-a-redux-selector-a517acee1fe8#:~:text=Why%20should%20you%20use%20a,store%20the%20minimal%20possible%20state.) about them

ttannis: I can create a separate method for this PR, while adding selectors can be a separate initiative. What do you think? (On the side I also want to identify a set of 2-3 concrete fe improvements initiatives based on the current challenges we shared, and see who in #frontend might want to help or own an initiative so this can be one)  

The reason we haven't invested much here yet is that initial Amundsen APIs were designed to basically serve the exact needs of the frontend, and entire chunks of state represented all the information that would be needed by the main components so there was no need for selectors yet. But now would be an appropriate time to trend towards this as our use cases increase in complexity.

Golodhros: Sounds good! Thanks!

PR #550, opened 2020-08-10
feat: Card Styling for announcements

Golodhros:
### Summary of Changes
Adds styling to the cards we'll use on announcements

Amundsen_-_Data_Discovery_Portal

### Tests
Adds new tests

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/HomePage/index.tsx
+++ b/amundsen_application/static/js/components/HomePage/index.tsx
@@ -16,6 +16,7 @@ import { resetSearchState } from 'ducks/search/reducer';
 import { UpdateSearchStateReset } from 'ducks/search/types';
 import SearchBar from 'components/common/SearchBar';
 import TagsList from 'components/common/TagsList';
+import Card from 'components/common/Card';
 import {
   SEARCH_BREADCRUMB_TEXT,
   HOMEPAGE_TITLE,
@@ -39,6 +40,7 @@ export class HomePage extends React.Component<HomePageProps> {
         <div className="row">
           <div className="col-xs-12 col-md-offset-1 col-md-10">
             <h1 className="sr-only">{HOMEPAGE_TITLE}</h1>
+            <Card isLoading />
             <SearchBar />
             <div className="filter-breadcrumb pull-right">
               <Breadcrumb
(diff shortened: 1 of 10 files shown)


allisonsuarez:
We have simmer loading components in a few places, granted all of them look a bit different depending on what component they are used for, but I wonder if we can hav them be together in some way? Other than that thought looks good to me.

Golodhros:
> We have simmer loading components in a few places, granted all of them look a bit different depending on what component they are used for, but I wonder if we can have them be together in some way? Other than that thought looks good to me.

Good call! Up to now, we have only reused the animation style, as the loaders have been pretty different. I think we are now in the position to abstract it a bit more, as Knowl wants us to use only two sizes of lines.

amundsen_application/static/js/components/HomePage/index.tsx
```diff
@@ -16,6 +16,7 @@ import { resetSearchState } from 'ducks/search/reducer';
 import { UpdateSearchStateReset } from 'ducks/search/types';
 import SearchBar from 'components/common/SearchBar';
 import TagsList from 'components/common/TagsList';
+import Card from 'components/common/Card';
```
ttannis, line 19: assume this is just for testing.

amundsen_application/static/js/components/common/Card/index.tsx
```diff
@@ -1,35 +1,41 @@
 import * as React from 'react';
+import { Link } from 'react-router-dom';
+
+import './styles.scss';
 
 export interface CardProps {
-  title: string;
+  title?: string;
```
ttannis, line 7: nit: Just wondering why all properties are optional, to be more generic/reuseable?

amundsen_application/static/js/components/HomePage/index.tsx
Golodhros: Yeah, will clean when done!

amundsen_application/static/js/components/common/Card/index.tsx
Golodhros: It is a byproduct of integrating the loading state inside the card, as we won't have any of the properties available when loading them.

PR #582, opened 2020-08-15
feat: Added GraphQL source to gschema views

dikshathakur3119:
### Summary of Changes

If table has Adding GraphQL as Table source, show link on top left corner

### Tests

Added Unit test cases

### Documentation

### CheckList


amundsen_application/static/js/components/TableDetail/SourceLink/index.tsx
```diff
@@ -16,8 +16,22 @@ const SourceLink: React.FC<SourceLinkProps> = ({
...
+      image = '/static/images/github.png';
+      break;
+    }
+    case 'gql': {
+      image = '/static/images/gql.png';
+      break;
+    }
+    default:
+      image = '';
+  }
```
Golodhros, line 34: Most of the times, switch statements are not recommended. Better off if we use a dictionary:
```
const sourceToImageURL = {
  'github': ...,
  'gql': ...,
};
const imageSRC = sourceToImageURL[tableSource.source_type] ? sourceToImageURL[tableSource.source_type] : '';
```

Golodhros: Also, those source names should probably be part of an Enum, and do:
```
const sourceToImageURL = {
  [SOURCES.gitHub]: ...,
  [SOURCES.graphQL]: ...,
};

File: amundsen_application/static/js/components/TableDetail/SourceLink/index.tsx [hunk 2]
```diff
@@ -16,8 +16,22 @@ const SourceLink: React.FC<SourceLinkProps> = ({
...
+  // const image =
+  //   tableSource.source_type === 'github' ? '/static/images/github.png' : '';
+
+  let image;
+  switch (tableSource.source_type) {
+    case 'github': {
+      image = '/static/images/github.png';
+      break;
+    }
+    case 'gql': {
```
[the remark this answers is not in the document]
dikshathakur3119: yes I can put this in private config. But I do not think this is specific to Lyft as Table lineage because we are not creating any url/ pattern specific to Lyft. It is more similar to adding new database which can be used by any company. Image repo provided here is path to images folder which is inside public repo.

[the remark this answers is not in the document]
dikshathakur3119: Do you think I should configure it similar to Mode and redash icon as done for dashboards? e.g https://github.com/amundsen-io/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/config/config-default.ts#L57

amundsen_application/static/js/components/TableDetail/SourceLink/index.tsx
```diff
@@ -16,22 +17,16 @@ const SourceLink: React.FC<SourceLinkProps> = ({
...
-  }
+  const config = AppConfig.tableSource;
+
+  const image = config[tableSource.source_type]
+    ? config[tableSource.source_type].iconPath
+    : '';
+
+  const displayName = config[tableSource.source_type]
+    ? config[tableSource.source_type].displayName
+    : tableSource.source_type;
```
Golodhros, line 28: Could we cover this logic with a test?

amundsen_application/static/js/components/TableDetail/SourceLink/index.tsx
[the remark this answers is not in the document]
feng-tao: if we think graphql generates table could be generic, do you think we could have a small doc about which case we could leverage this new source link?

This could help to adopt this new source icon.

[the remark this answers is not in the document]
feng-tao: I am totally ok to put this in the repo, but would like have a doc to document this feature and how users could use it.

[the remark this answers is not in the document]
dikshathakur3119: yeah I will add documentation for this feature. We can keep github as default table source in public configurations and new table sources can be added to private configurations.

ttannis:
@dikshathakur3119 is this officially ready for review? I have comments that would change the direction in terms of the configuration, but I'm concerned we're introducing churn to your work by posting reviews before you've marked this as ready.

dikshathakur3119:
> @dikshathakur3119 is this officially ready for review? I have comments that would change the direction in terms of the configuration, but I'm concerned we're introducing churn to your work by posting reviews before you've marked this as ready.

From logic part I am done, I will be adding unit test cases and documentation. Now it would be good to have your comments related to configuration.

amundsen_application/static/js/components/TableDetail/SourceLink/index.tsx
dikshathakur3119: Added tests

amundsen_application/static/js/config/config-default.ts
```diff
@@ -186,6 +186,12 @@ const configDefault: AppConfig = {
       return `https://DEFAULT_EXPLORE_URL?schema=${schema}&cluster=${cluster}&db=${database}&table=${table}`;
     },
   },
+  tableSource: {
```
ttannis, line 189: What do you think about adding an `upstreamSources` configuration to the existing table resource configuration. Let me try to use the types to convey what I mean.

Existing code in config.types:
```
type SourcesConfig = {
  [id: string]: {
    displayName?: string;
    iconClass?: string;
  };
};

interface BaseResourceConfig {
  displayName: string;
  filterCategories?: FilterConfig;
  supportedSources?: SourcesConfig;
}
...
interface ResourceConfig {
  [ResourceType.dashboard]: BaseResourceConfig;
  [ResourceType.table]: BaseResourceConfig;
  [ResourceType.user]: BaseResourceConfig;
}
```

We could extend to:
```
type SourcesConfig = {
  [id: string]: {
    displayName?: string;
    iconClass?: string;
    iconPath?: string; // add new parameter
  };
};

// add new extended interface
interface TableResourceConfig extends BaseResourceConfig {
  upstreamSources?: SourcesConfig;
}

interface ResourceConfig {
  [ResourceType.dashboard]: BaseResourceConfig;
  [ResourceType.table]: TableResourceConfig; // use new interface here
  [ResourceType.user]: BaseResourceConfig;
}
```
Then follow similar patterns in terms of setting up the icons and configuration helper methods (e.g. not importing and accessing `AppConfig` directly).

I think this helps because:
1. It fits nicely into existing precedents by extending them vs. having to create something new that is almost exactly the same.
2. Makes its such this new table related configuration is part of the existing table resource configuration object, and developers won't have to jump around to different spots to configure everything for tables because it will all be part of the same nested object.

What do you think about this?

amundsen_application/static/js/components/TableDetail/SourceLink/index.spec.tsx
```diff
@@ -0,0 +1,106 @@
...
+    it('renders AvatarLabel', () => {
+      const { wrapper } = setup();
+      const expected = 1;
+      const actual = wrapper.find(AvatarLabel).length;
+
+      expect(actual).toEqual(expected);
+    });
+
+    describe('renders correct display name and icon path', () => {
+      it('when source type is present in configurations', () => {
```
Golodhros, line 84: Nit: "when" clauses usually go on "describe" blocks, something like:
```
describe('when source type is present in configuration', () => {
   it('should pass the proper info to AvatarLabel',
```

amundsen_application/static/js/components/TableDetail/SourceLink/index.tsx
```diff
@@ -3,9 +3,14 @@
 
 import * as React from 'react';
 
+import {
+  getDescriptionSourceDisplayName,
+  getDescriptionSourceIconPath,
+} from 'config/config-utils';
 import { logClick } from 'ducks/utilMethods';
 import AvatarLabel from 'components/common/AvatarLabel';
 import { TableSource } from 'interfaces';
```
Golodhros, line 12: if you do `import { TableSource, ResourceType } from 'interfaces'; `,  would that work?

amundsen_application/static/js/components/TableDetail/SourceLink/index.tsx
```diff
@@ -16,8 +21,10 @@ const SourceLink: React.FC<SourceLinkProps> = ({
 }: SourceLinkProps) => {
   if (tableSource === null || tableSource.source === null) return null;
 
-  const image =
-    tableSource.source_type === 'github' ? '/static/images/github.png' : '';
+  const image = getDescriptionSourceIconPath(tableSource.source_type);
+
+  const displayName = getDescriptionSourceDisplayName(tableSource.source_type);
```
Golodhros, line 26: Nit: I would just inline these calls on the line 37 and avoid the temporal variables

amundsen_application/static/js/config/config-types.ts
```diff
@@ -119,6 +119,7 @@ type SourcesConfig = {
   [id: string]: {
     displayName?: string;
     iconClass?: string;
+    iconPath?: string;
```
Golodhros, line 122: To get rid of the two new TypeScript errors you got you would need to remove the optionality of displayName and iconPath.

Maybe you don't want to share this type? What if you create a `CodeSource` type that requires iconPath and displayName?

amundsen_application/static/js/config/config-utils.ts
```diff
@@ -205,3 +205,41 @@ export function generateExploreUrl(tableData: TableMetadata): string {
...
+ * If a configuration or display name does not exist for the given description source id, the id
+ * is returned.
+ */
+export function getDescriptionSourceDisplayName(sourceId: string): string {
+  const config = AppConfig.resourceConfig[ResourceType.table];
+  if (
+    !config ||
+    !config.supportedDescriptionSources ||
+    !config.supportedDescriptionSources[sourceId] ||
+    !config.supportedDescriptionSources[sourceId].displayName
```
Golodhros, line 220: Nit: usually it is easier to understand a non-negated condition, so it would be easier to do:
 ``` if (
    config ||
    config.supportedDescriptionSources ||
    config.supportedDescriptionSources[sourceId] ||
    config.supportedDescriptionSources[sourceId].displayName
){
 return config.supported....
}
return sourceId;```

amundsen_application/static/js/components/TableDetail/SourceLink/index.spec.tsx
dikshathakur3119: Okay. I did not know that. Thanks Marcos!

amundsen_application/static/js/components/TableDetail/SourceLink/index.tsx
dikshathakur3119: yeah it would work. I would change it. Thanks

amundsen_application/static/js/components/TableDetail/SourceLink/index.tsx
dikshathakur3119: 👍

amundsen_application/static/js/config/config-types.ts
dikshathakur3119: I am not sure if we want to do that. from @ttannis review I am re-using existing components. 
From your suggestion, I can do following following so that I do not get type error. 

create new object `DescriptionSourceConfig` similar to `SourceConfig`: 
````
DescriptionSourceConfig =  [id: string]: {
displayName: string;
iconPath: string;
};`

interface TableResourceConfig extends BaseResourceConfig {
  supportedDescriptionSources?: DescriptionSourceConfig;
}
```

@ttannis let me know what do you think about it?

amundsen_application/static/js/config/config-utils.ts
dikshathakur3119: 👍

amundsen_application/static/js/config/config-types.ts
Golodhros: That would work too!

dikshathakur3119: will make the changes. Thanks!

PR #595, opened 2020-08-18
feat: Update 'table view', 'alumni' and SLA status badges

allisonsuarez:
### Summary of Changes

The "Table View" badge on tables and "Alumni" badge on users are not real badges since they are part of the structured metadata. Since you can't click on these and filter. We should update the design to not look like a badge. Similarly the SLA Hit/Missed status should be updated to no longer use badges.

### Tests

Removed tests that specifically required the rendering of a Flag component on these pages and added ones to test the new design.

### Documentation

N/A

### CheckList


amundsen_application/static/css/_icons.scss
```diff
@@ -196,6 +196,11 @@ img.icon {
     -webkit-mask-image: url('/static/images/icons/More.svg');
     mask-image: url('/static/images/icons/More.svg');
   }
+
+  &.icon-check {
+    -webkit-mask-image: url('/static/images/icons/check.svg');
+    mask-image: url('/static/images/icons/check.svg');
+  }
```
Golodhros, line 203: I don't recommend to use this approach for new icons, as it is not supported on Firefox and other browsers and frankly, is less than ideal.

I would try to follow the pattern at the top, using the iconBackgrounds mixin

amundsen_application/static/js/components/common/ResourceListItem/UserListItem/index.tsx
```diff
@@ -55,12 +55,6 @@ class UserListItem extends React.Component<UserListItemProps, {}> {
             </div>
           </div>
           <div className="resource-type">User</div>
-          <div className="resource-badges">
```
ttannis, line 58: We still need this `<div>` and the ``, just remove the `Flag`.

amundsen_application/static/js/components/common/ResourceListItem/UserListItem/index.spec.tsx
```diff
@@ -171,56 +171,6 @@ describe('UserListItem', () => {
...
-      let resourceBadges;
-      beforeAll(() => {
-        resourceBadges = wrapper.find('.resource-badges');
-      });
-
-      it('renders resource badges section', () => {
-        expect(resourceBadges.exists()).toBe(true);
-      });
-
-      it('does not render Alumni flag if user is active', () => {
```
ttannis, line 185: We still need this `<div>` and the ``, just remove the two tests related to the `Flag`

amundsen_application/static/js/components/DashboardPage/RunState/index.spec.tsx
```diff
@@ -0,0 +1,48 @@
...
+  };
+  const wrapper = mount<typeof RunStateContainer>(
+    <RunStateContainer {...props} />
+  );
+  return { props, wrapper };
+};
+
+describe('RunStateContainer', () => {
+  describe('Hit', () => {
+    const { wrapper } = setup({ lastRunState: LAST_RUN_SUCCEEDED });
```
Golodhros, line 25: I would include the setup inside the 'it' clause

amundsen_application/static/js/components/DashboardPage/RunState/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+const MissedState: React.FC = () => {
+  return (
+    <div className="missed">
+      <div className="missed-icon">
+        <div className="exclamation-top" />
+        <div className="exclamation-bottom" />
+      </div>
+      <span className="status-text body-2 text-primary">{MISSED_RUN_TEXT}</span>
```
Golodhros, line 21: I would avoid to use text classes, and instead add the text styles to the 'status-text' class with:
```
@import 'typography-default';
...

.missed .status-text {
  @extend %text-body-w2;
}

File: amundsen_application/static/js/components/DashboardPage/RunState/styles.scss [hunk 6]
```diff
@@ -0,0 +1,47 @@
...
+.missed-icon,
+.hit-icon {
+    width: 14px;
+    height: 14px;
+    border-radius: 100%;
+    display: inline-block;
+    vertical-align: middle;
+}
+.missed-icon {
+    background-color: #DB3615;
```
Golodhros, line 15: Let's use a variable, or a given color from our _colors file if possible

amundsen_application/static/js/components/DashboardPage/RunState/styles.scss
```diff
@@ -0,0 +1,47 @@
+// Copyright Contributors to the Amundsen project.
+// SPDX-License-Identifier: Apache-2.0
+
+@import 'variables';
+
+.missed-icon,
+.hit-icon {
+    width: 14px;
```
Golodhros, line 8: I would create a $icon-size variable

amundsen_application/static/js/components/DashboardPage/RunState/styles.scss
```diff
@@ -0,0 +1,47 @@
...
+    border-radius: 100%;
+    display: inline-block;
+    vertical-align: middle;
+}
+.missed-icon {
+    background-color: #DB3615;
+    .exclamation-top {
+        width: 2px;
+        height: 4px;
+        background-color: #F2F2F2;
```
Golodhros, line 19: Variable please

amundsen_application/static/js/components/ProfilePage/index.spec.tsx
```diff
@@ -366,6 +349,17 @@ describe('ProfilePage', () => {
...
 
+    it('renders alumni bullet is user not active', () => {
+      const userCopy = {
+        ...globalState.user.profile.user,
+        is_active: false,
+      };
+      const { wrapper } = setup({
+        user: userCopy,
+      });
+      expect(wrapper.find('#alumni').text()).toEqual('Alumni');
```
Golodhros, line 360: I would avoid to check for text on the UI, this makes this test fragile

amundsen_application/static/js/components/common/ResourceListItem/UserListItem/index.tsx
allisonsuarez: My understanding was that in the future Users would not have badges anymore, are there any other badges that are not Alumni for users?

allisonsuarez: Also I don't see the `icon-right` img in the user page at all in prod.

ttannis: 1. The decision is not that Users will never have badges, but that the `Alumni` UI would change. Users could still have badges if there is a product requirement for it (not at Lyft but maybe elsewhere). Additionally, there's a case for `Alumni` returning to official badge status in the future, so that we can lower the search ranking for old employees. (Example: I'm not the first person with my name that shows up in user search, but I'm the only active employee 🙃 )
2. The icon shows up on hover.

amundsen_application/static/js/components/ProfilePage/index.spec.tsx
ttannis: Could also check against a constant value.

amundsen_application/static/css/_icons.scss
```diff
@@ -43,6 +49,7 @@ span.icon {
   @include iconBackgrounds($data-stores);
   @include iconBackgrounds($dashboards);
   @include iconBackgrounds($users);
+  @include iconBackgrounds($check);
```
Golodhros, line 53: Could you please add a comment saying that the code below is deprecated and we shouldn't add more icons this way? 

I should have done it before! Thanks!

amundsen_application/static/js/components/DashboardPage/constants.ts
```diff
@@ -6,5 +6,8 @@ export const OWNER_HEADER_TEXT = 'Owners';
 export const ADD_DESC_TEXT = 'Add Description in';
 export const EDIT_DESC_TEXT = 'Click to edit description in';
 
+export const MISSED_RUN_TEXT = 'Missed';
+export const HIT_RUN_TEXT = 'Hit';
+
 export const LAST_RUN_SUCCEEDED = 'succeeded';
-export const LAST_RUN_FAILED = 'failed';
+
```
Golodhros, line 13: Are your pre-commit hooks working at the moment? They should have fixed this automatically.

If they don't work, try to do `rm -rf node_modules && npm i`

amundsen_application/static/css/_variables-default.scss
```diff
@@ -92,6 +92,10 @@ $icon-bg-brand: $brand-color-3 !default;
 $icon-bg-dark: $gray60 !default;
 $icon-bg-disabled: $gray20 !default;
 $red-triangle-warning: $sunset60;
+$icon-status-color: #f2f2f2;
+$icon-status-missed: $sunset60;
+$icon-status-hit: #00824C;
+$icon-size: 14px;
```
Golodhros, line 98: I wouldn't bring this variables up here, it could be deceiving.
I would let them on the specific file.

amundsen_application/static/js/components/common/ResourceStatusMarker/styles.scss
```diff
@@ -0,0 +1,49 @@
...
+        margin-left: 2px;
+        min-width: 0px;
+    }
+}
+
+.status-text {
+    display: inline-block;
+    margin-left: $spacer-1;
+    vertical-align: middle;
+    @extend %text-body-w2;
```
Golodhros, line 48: I would expect this to give you a warning, as they usually need to go at the top with a space below.

amundsen_application/static/css/_icons.scss
```diff
@@ -55,6 +60,8 @@ span.icon {
 }
 
 img.icon {
+  /*DEPRECATED: follow behavior above to generate
```
ttannis, line 63: nit: In favor of filing issues on the open source issues vs. leaving comments that end up being vague for people without context. So I'd suggest we file an issue for this concept, and then if we do feel strongly about leaving a comment, we can leave the link to the issue.

amundsen_application/static/js/components/DashboardPage/constants.ts
```diff
@@ -6,5 +6,7 @@ export const OWNER_HEADER_TEXT = 'Owners';
 export const ADD_DESC_TEXT = 'Add Description in';
 export const EDIT_DESC_TEXT = 'Click to edit description in';
 
+export const MISSED_RUN_TEXT = 'Missed';
+export const HIT_RUN_TEXT = 'Hit';
+
 export const LAST_RUN_SUCCEEDED = 'succeeded';
```
ttannis, line 12: I think this needs to be capitalized now? Before I think the `Flag` component converted it. We also aren't changing what text this component shows, it should still use these constants...unless Design has changed the Dashboard UI text to `Hit` and `Missed`?

amundsen_application/static/js/components/DashboardPage/index.tsx
```diff
@@ -297,10 +299,13 @@ export class DashboardPage extends React.Component<
                         : NO_TIMESTAMP_TEXT}
                     </time>
                     <div className="last-run-state">
-                      <Flag
-                        caseType="sentenceCase"
-                        text={dashboard.last_run_state}
-                        labelStyle={this.mapStatusToStyle(
+                      <ResourceStatusMarker
```
ttannis, line 302: We still want to make sure this shows the same text. `Succeeded` or `Failed`, whereas it's the upcoming SLA use case on tabledeatil page would pass in `Hit` & `Missed`.

amundsen_application/static/js/components/DashboardPage/constants.ts
allisonsuarez: Design was changed to Hit and Missed according to the mocks.

amundsen_application/static/css/_icons.scss
allisonsuarez: I created an issue to have these all be refactored, in the meanwhile I think the comment might be necessary so that of anyone else is adding an icon before the issue is fixed they don't make the same mistake.

Golodhros (dismissed):
LGTM

amundsen_application/static/js/components/DashboardPage/constants.ts
ttannis: Can you paste in slack? I see the mock, but I'm only seeing Hit and Missed on the table page.

PR #611, opened 2020-08-25
feat: Parse column types

ttannis:
Signed-off-by: Tamika Tannis <ttannis@lyft.com>

### Summary of Changes

This PR:
1. Passes `database` property from `TableDetail` through the column list component to `ColumnType`.
2. Creates `parser.ts` which owns the logic for parsing complex column types for supported databases. It creates an object of the `NestedType` interface.
3. Adds logic to `ColumnType` to render the information in `NestedType`. 

### Tests

Added some tests for `parser.ts`.

TODO on feature branch before merge:
1. Tests for `parseNestedTypeHelper` after collecting use cases.
2. Still need tests for `ColumnType` interactions

### Documentation

N/A for this feature

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/TableDetail/ColumnListItem/ColumnType/index.tsx
+++ b/amundsen_application/static/js/components/TableDetail/ColumnListItem/ColumnType/index.tsx
@@ -6,11 +6,15 @@ import { Modal, OverlayTrigger, Popover } from 'react-bootstrap';
 
 import './styles.scss';
 
+import { getTruncatedText, parseComplexType, NestedType, ParsedType } from './parser';
+
 const CTA_TEXT = 'Click to see nested fields';
 const MODAL_TITLE = 'Nested Type';
+const TEXT_INDENT= 8;
 
 export interface ColumnTypeProps {
   columnName: string;
+  database: string;
   type: string;
 }
 
@@ -22,12 +26,16 @@ export class ColumnType extends React.Component<
   ColumnTypeProps,
   ColumnTypeState
 > {
+  parsedType: ParsedType;
+
   constructor(props) {
     super(props);
 
     this.state = {
       showModal: false,
     };
+
+    this.parsedType = parseComplexType(this.props.type, this.props.database);
   }
 
   hideModal = (e) => {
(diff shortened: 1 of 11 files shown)


ttannis:
@Golodhros I'm not understanding why the betterer check doesn't pass here, as I expected it to only block on errors with new code not existing code. Am I missing something?

amundsen_application/static/js/components/TableDetail/ColumnListItem/ColumnType/index.tsx
```diff
@@ -44,74 +57,86 @@ export class ColumnType extends React.Component<
...
+          overlay={popoverHover}
+          rootClose
+        >
+          <a
+            className="column-type"
+            href="JavaScript:void(0)"
+            onClick={this.showModal}
+          >
+            {getTruncatedText(this.nestedType)}
+          </a>
```
Golodhros, line 119: Button instead of link?

amundsen_application/static/js/components/TableDetail/ColumnListItem/ColumnType/index.tsx
```diff
@@ -44,74 +57,86 @@ export class ColumnType extends React.Component<
...
+        {this.createLineItem(tail, textIndent)}
+      </div>
+    );
+  };
+
+  render = () => {
+    const { columnName, database, type } = this.props;
+
+    if (this.nestedType === null) {
+      return <div className="column-type">{type}</div>;
```
Golodhros, line 94: I would expect a different tag to wrap a text, maybe a low h or a p?

amundsen_application/static/js/components/TableDetail/ColumnListItem/ColumnType/parser.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  return {
+    nextStartIndex: currentIndex + 1,
+    results: children,
+  };
+}
+
+/*
+ * Returns whether or not a columnType string represents a complexc type for the given database
```
Golodhros, line 94: Nit: complex

amundsen_application/static/js/components/TableDetail/ColumnListItem/ColumnType/parser.ts
```diff
@@ -0,0 +1,126 @@
...
+  return null;
+}
+
+/*
+ * Returns the truncated string representation for a NestedType
+ */
+export function getTruncatedText(nestedType: NestedType): string {
+  const { head, tail } = nestedType;
+  return `${head}...${tail.replace(SEPARATOR_DELIMETER, '')}`;
+}
```
Golodhros, line 126: It would be really useful to have tests for this kind of logic, as it is not trivial.

Golodhros:
> @Golodhros I'm not understanding why the betterer check doesn't pass here, as I expected it to only block on errors with new code not existing code. Am I missing something?

Sometimes it happens that the line of the error changes, so the new error entry is different. You will need to update the results with `npm run betterer:update` if that's the case.

amundsen_application/static/js/components/TableDetail/ColumnListItem/ColumnType/parser.ts
ttannis: I plan on it but I'm mostly a "write tests when I'm sure of the implementation" person, but I'll add tests for these smaller functions because its really the `parseNestedTypeHelper` that I want manually test on a greater sample size. (i.e. trying to find a very complex case of each type for each database to check any assumptions I might have made, and then use those kinds of cases for unit tests)

amundsen_application/static/js/components/TableDetail/ColumnListItem/ColumnType/index.tsx
ttannis: `p` seems appropriate.

amundsen_application/static/js/components/TableDetail/ColumnListItem/ColumnType/index.tsx
ttannis: Thanks for reminding

ttannis:
> > @Golodhros I'm not understanding why the betterer check doesn't pass here, as I expected it to only block on errors with new code not existing code. Am I missing something?
> 
> Sometimes it happens that the line of the error changes, so the new error entry is different. You will need to update the results with `npm run betterer:update` if that's the case.

Oh that makes sense. I'll see how many I can fix in a chore PR, especially for files I think we are about to touch quite often for the new project`.

amundsen_application/static/js/components/TableDetail/ColumnListItem/ColumnType/index.tsx
```diff
@@ -110,13 +110,9 @@ export class ColumnType extends React.Component<
           overlay={popoverHover}
           rootClose
         >
-          <a
-            className="column-type"
-            href="JavaScript:void(0)"
-            onClick={this.showModal}
-          >
+          <button className="column-type" onClick={this.showModal}>
```
Golodhros, line 113: You would need to add `type="button"` as well.

amundsen_application/static/js/components/TableDetail/ColumnListItem/ColumnType/styles.scss
```diff
@@ -11,6 +11,18 @@ $modal-header-height: 94px;
 // TODO (design qa): Also using fixed width from Figma
 $modal-dialog-width: 418px;
 
+button.column-type {
```
Golodhros, line 14: I am mostly against styling using tags, as it makes harder to solve specificity issues.
I would recommend instead to create a new class, something like `.column-type-btn` and style that one.

feng-tao:
would like to take a look as well once the pr is merged back to master.

PR #621, opened 2020-08-28
feat: Alerts component

Golodhros:
### Summary of Changes
Adds the alerts component
Includes tests and a SVG icon for the alert

### Documentation
Added Storybook story for the component

Components___Alert_-_Alert_⋅_Storybook

Attributes___Icons_-_SVG_Icons_⋅_Storybook

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/common/Alert/AlertIcon.tsx
+++ b/amundsen_application/static/js/components/common/Alert/AlertIcon.tsx
@@ -1,0 +1,24 @@
+import * as React from 'react';
+
+const STROKE_COLOR = '#b8072c'; // $red70
+const SIZE = 16;
+
+const AlertIcon: React.FC = () => {
+  return (
+    <svg
+      width={SIZE}
+      height={SIZE}
+      viewBox="0 0 24 24"
+      fill="none"
+      stroke={STROKE_COLOR}
+      strokeWidth={2}
+      strokeLinecap="round"
+      strokeLinejoin="round"
+      className="alert-triangle-svg-icon"
+    >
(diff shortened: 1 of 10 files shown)


amundsen_application/static/js/components/common/Alert/AlertIcon.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+
+const STROKE_COLOR = '#b8072c'; // $red70
+const SIZE = 16;
+
+const AlertIcon: React.FC = () => {
+  return (
+    <svg
```
ttannis, line 8: Would it make sense to have a `common/icons` where we start putting new amundsen icons that use inline-svgs?

amundsen_application/static/js/components/common/Alert/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+  message,
+  onAction,
+  actionText,
+}: AlertProps) => {
+  return (
+    <div className="alert">
+      <AlertIcon />
+      <p className="alert-message">{message}</p>
+      {actionText && onAction && (
+        <button type="button" className="btn btn-link" onClick={onAction}>
```
ttannis, line 26: For our internal use case, the text will end up being a link to another Amundsen page. Given our desire to use the correct components for specific actions, do you think we should still support a button with onClick vs and anchor with an href...or both?

amundsen_application/static/js/components/common/Alert/AlertIcon.tsx
Golodhros: Makes sense! I will move it there!

amundsen_application/static/js/components/common/Alert/index.tsx
Golodhros: Oh, I see.
I could make it support both easily. I will add an optional actionHref prop, if that has a value, I will render a router Link instead

PR #624, opened 2020-08-29
refactor: Moves pages into its own folder

Golodhros:
### Summary of Changes
Moves the Pages currently living on the components/ folder into their own pages/ folder
Updates all paths

amundsenfrontend-private

### Tests
Ran tests
Tested manually

### CheckList


allisonsuarez (dismissed):
LGTM, thank you for doing this!

Golodhros:
> Since we're creating a new "top-level" directory, I would expect to see some changes in:
> 
> 1. `jest.config.js` where we set coverage thresholds to ensure our coverage does not get worse. We decided a long time ago that we would aim for at least 75% across all testable code, and every now and then I actually bump the threshold depending on how much we've improved to try to keep in that way. Could you set some thresholds for `pages` in that file based on what our current coverage happens to be, as well as update the thresholds for `components`? I usually leave at least 5% buffer, at most 10%.
> 
> 2.`webpack.common` where we define paths.

Good catch! Updating those!

PR #628, opened 2020-08-31
refactor: Refactors connector out of ColumnLIstItems

Golodhros:
### Summary of Changes
Refactors the Redux connector out of the ColumnListItems
Fixes issue with RequestMetadataForm
Adds optional usage field to ColumnData

### Tests
Updated tests

### CheckList


Golodhros:
> LGTM! I wanted to know why we want to remove redux connector from ColumnListIndex?

Sure!
There are several reasons:
1.- You don't want to have a lot of connected components. The way we have it now, it creates a connected component for each of the rows on the list. That means more complexity, more memory consumption and it is in general, wasteful.
2.- We are connecting the ColumnListIndex not to get information from the state to render the item, as we already have that info. We are connecting it to get a handler function, and this handler function is the same for all the rows, so we can just get that function once and pass it down a couple of levels. This PR implements that.
3.- By removing the connection, ColumnListItem becomes a "dumb" component, also called "presentational" component. This makes this component more reusable and easy to test. We want to have as many presentational components as we can.

PR #636, opened 2020-09-01
feat: Adds a basic Table to our component library

Golodhros:
### Summary of Changes
Adds a basic table to our component library

### Tests
Added tests

### Documentation
Added story
Components___Table_-_Table_⋅_Storybook

### CheckList


amundsen_application/static/js/components/common/Table/testDataBuilder.ts
```diff
@@ -0,0 +1,74 @@
+const defaultData = [
```
ttannis, line 1: Something I don't runderstand about this yet is that it appears to be a helper method for the test `setup()` where propoverrides + jest methods like `beforeAll()` can be used to set up different scenarios. 

Any examples you want to merge in will help with discussion though, which would be centered around understanding the value of creating another file/layer of code and why its abstracted out in this way (over let's say, being built into the `setup()` method support.

Golodhros: Yep, it can be used inside the setup method, but also within each test case to pass the values into the setup method.

The utility of this extra layer is the flexibility it gives us, as we can use it on unit tests as we do here, and also on stories for our Storybook and for End to End tests with Cypress in the future.

amundsen_application/static/js/components/common/Table/index.spec.tsx
```diff
@@ -0,0 +1,122 @@
...
+          const expected = 9;
+          const actual = wrapper.find('.ams-table-row .ams-table-cell').length;
+
+          expect(actual).toEqual(expected);
+        });
+      });
+    });
+
+    describe('when more data than columns', () => {
+      const { columns, data } = dataBuilder.withMoreDataThanColumns().build();
```
Golodhros, line 87: See here @ttannis how we can use it within specific test suites too

PR #668, opened 2020-09-15
fix: disable announcement component by default

feng-tao:
Signed-off-by: feng-tao <fengtao04@gmail.com>

### Summary of Changes

Fix https://github.com/amundsen-io/amundsen/issues/690, currently we only provide an example but not e2e integration to enable announcement. We should disable the component by default and enable it through config-custom file.

### Tests

### Documentation

### CheckList


feng-tao:
test it locally by applying the custom config and verify the announcement page shows up in homepage in local env.
cc @Golodhros @ttannis

amundsen_application/static/js/config/config-custom.ts
```diff
@@ -16,6 +16,9 @@ const configCustom: AppConfigCustom = {
     feedbackEnabled: false,
     notificationsEnabled: false,
   },
+  announcements: {
```
ttannis, line 19: We don't need to update this file with the defaults, in the public repo it exists as more of an example. So no need to remove it because it does not harm, just clarifying that its not absolutely necessary.

amundsen_application/static/js/config/config-default.ts
```diff
@@ -38,7 +38,7 @@ const configDefault: AppConfig = {
     notificationsEnabled: false,
   },
   announcements: {
-    enabled: true,
+    enabled: false,
```
ttannis, line 41: fyi our internal version will need to be updated to `true` when we update the submodule.

feng-tao:
thanks @ttannis , PR is updated

amundsen_application/static/js/config/index.spec.ts
```diff
@@ -132,6 +132,7 @@ describe('getNavLinks', () => {
 
   describe('when announcements is active', () => {
     it('returns all the navLinks', () => {
+      AppConfig.announcements.enabled = true;
```
ttannis, line 135: I notice here you are just following other examples this test file is using by directly setting the value needed for the test, however this habit can cause side effects in other tests because the modifications will apply across all unit tests in this file.

It's been on my todo list to update these tests to show how to properly mock the object, so I will submit a PR this week for that and it can be used as an example moving forward.

ttannis (approved):

amundsen_application/static/js/config/index.spec.ts
feng-tao: oh cool, are all the JS tests executed in parallel or in sequential order?

PR #669, opened 2020-09-15
feat: accept new badges

allisonsuarez:
### Summary of Changes

Temporary change to accept both types of badges in badge list

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/common/BadgeList/index.tsx
+++ b/amundsen_application/static/js/components/common/BadgeList/index.tsx
@@ -4,7 +4,7 @@
 import * as React from 'react';
 
 import ClickableBadge from 'components/common/Badges';
-import { getBadgeConfig } from 'config/config-utils';
+import { getBadgeConfig, mapBadgeStyle } from 'config/config-utils';
 import { Badge } from 'interfaces/Tags';
 
 export interface BadgeListProps {
@@ -29,7 +29,7 @@ const BadgeList: React.FC<BadgeListProps> = ({ badges }: BadgeListProps) => {
           return (
             <ClickableBadge
               text={badge.badge_name}
-              labelStyle={badge.badge_type}
+              labelStyle={mapBadgeStyle(badge.badge_type)}
               key={`badge-${index}`}
             />
           );
--- a/amundsen_application/static/js/components/common/BadgeList/index.tsx
+++ b/amundsen_application/static/js/components/common/BadgeList/index.tsx
@@ -15,7 +15,7 @@ export interface BadgeListProps {
 /*
  * maps badge type to a badge style
  */
-export function mapBadgeStyle(badgeType: string): BadgeStyle {
+function mapBadgeStyle(badgeType: string): BadgeStyle {
   if (badgeType === 'negative') return BadgeStyle.DANGER;
   if (badgeType === 'positive') return BadgeStyle.SUCCESS;
   if (badgeType === 'warning') return BadgeStyle.WARNING;
--- a/amundsen_application/static/js/components/common/BadgeList/index.tsx
+++ b/amundsen_application/static/js/components/common/BadgeList/index.tsx
@@ -11,13 +11,10 @@ export interface BadgeListProps {
   badges: any[]; // TODO replace with new badges later
 }
 
-
-
 const BadgeList: React.FC<BadgeListProps> = ({ badges }: BadgeListProps) => {
   return (
     <span className="badge-list">
       {badges.map((badge, index) => {
-        var name;
         if (badge.tag_name) {
           const badgeConfig = getBadgeConfig(badge.tag_name);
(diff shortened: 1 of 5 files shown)


amundsen_application/static/js/components/common/BadgeList/index.tsx
```diff
@@ -8,14 +8,17 @@ import { getBadgeConfig } from 'config/config-utils';
 import { Badge } from 'interfaces/Tags';
 
 export interface BadgeListProps {
-  badges: Badge[];
+  badges: any[]; // TODO replace with new badges later
```
ttannis, line 11: nit: This is personal preference but can you add your github handle to this, e.g `TODO (ttannis): ...`. Same with all temporary code and files that you are adding. Since temporary code merged into master can be a bit confusing -- and in the worst case de-stabilizing if anyone else happens to be working on those files -- I think this habit helps signal who to reach out to if it causes any problems.

Additionally...it's even easier to fix the temp code later, or even have someone else without context fix it if the comment has a good explanation. Because you/they can easily search your handle to quickly find all the spots that need changing, and then just follow any directions.

amundsen_application/static/js/components/common/BadgeList/index.tsx
```diff
@@ -8,14 +8,17 @@ import { getBadgeConfig } from 'config/config-utils';
...
 }
 
 const BadgeList: React.FC<BadgeListProps> = ({ badges }: BadgeListProps) => {
   return (
     <span className="badge-list">
       {badges.map((badge, index) => {
-        const badgeConfig = getBadgeConfig(badge.tag_name);
+        var name;
```
ttannis, line 20: `let name;`. 

`var` is an old way of declaring all variables in JavaScript. Several years ago a new standard was introduced to use `let` for variables that would be reassignable, and `const` for ones that weren't. Long story short this allowed web engines to optimize certain processes.t.

ttannis (commented):
Initial comments.

amundsen_application/static/js/components/common/Flag/index.tsx
```diff
@@ -16,7 +16,7 @@ export enum CaseType {
 export interface FlagProps {
   caseType?: string | null;
   text: string;
-  labelStyle?: BadgeStyle;
+  labelStyle?: string;
```
ttannis, line 19: This shouldn't have to change.

amundsen_application/static/js/components/common/BadgeList/index.tsx
```diff
@@ -8,22 +8,32 @@ import { getBadgeConfig } from 'config/config-utils';
...
+              labelStyle={badgeConfig.style}
+              key={`badge-${index}`}
+            />
+          );
+        }
+        if (badge.badge_name) {
+          return (
+            <ClickableBadge
+              text={badge.badge_name}
+              labelStyle={badge.badge_type}
```
ttannis, line 32: I would convert the `badge_type` to a `BadgeStyle`. Yes, the string values do happen to be the same but our implementation shouldn't be so tightly coupled with the fact that we know exactly what metadata is sending back. `BadgeList` should have intentional logic that coverts all supported `badge_type` to a `BadgeStyle`, accounting for edge cases like mapping any unsupported or unexpected `badge_type` to the neutral style, etc.

ttannis (commented):
What do you think of the following comments?

amundsen_application/static/js/config/config-utils.ts
```diff
@@ -90,6 +90,16 @@ export function getBadgeConfig(badgeName: string): BadgeStyleConfig {
   };
 }
 
+/*
+ * maps badge type to a badge style
+ */
+export function mapBadgeStyle(badgeType: string): BadgeStyle {
```
ttannis, line 96: nit: This isn't particularly related to a configuration so I'd suggest `BadgeList` own this logic.

ttannis (commented):
Unit tests are failing because we have coverage thresholds, and the untested method added on `config-utils` lowered the coverage:
```
Jest: "./js/config" coverage threshold for branches (90%) not met: 85%
```

amundsen_application/static/js/components/common/BadgeList/index.tsx
```diff
@@ -5,25 +5,46 @@ import * as React from 'react';
...
 
 export interface BadgeListProps {
-  badges: Badge[];
+  badges: any[]; // TODO replace with new badges later @allisonsuarez
+}
+
+/*
+ * maps badge type to a badge style
+ */
+export function mapBadgeStyle(badgeType: string): BadgeStyle {
```
ttannis, line 18: nit: We don't need to export the function if it is only used in this file.

ttannis (approved):

PR #691, opened 2020-09-28
feat: Adds sortable table detail page

Golodhros:
### Summary of Changes
Adds ability to sort the table columns on the Table detail page

core_fact_rides_-_Amundsen_Table_Details

### Tests
Added tests

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/ColumnList/index.spec.tsx
+++ b/amundsen_application/static/js/components/ColumnList/index.spec.tsx
@@ -96,6 +96,23 @@ describe('ColumnList', () => {
 
         expect(actual).toEqual(expected);
       });
+
+      describe('when sorting is passed', () => {
+        it('should sort the data by that value', () => {
+          const { wrapper } = setup({
+            columns,
+            sortBy: { key: 'usage', direction: 'desc' },
+          });
+          const expected = 'simple_column_name_timestamp';
+          const actual = wrapper
+            .find('.table-detail-table .ams-table-row')
+            .at(0)
+            .find('.column-name')
+            .text();
+
+          expect(actual).toEqual(expected);
+        });
+      });
     });
 
     describe('when complex type columns are passed', () => {
--- a/amundsen_application/static/js/components/ColumnList/index.spec.tsx
+++ b/amundsen_application/static/js/components/ColumnList/index.spec.tsx
@@ -114,17 +114,59 @@ describe('ColumnList', () => {
         expect(actual).toEqual(expected);
       });
 
-      describe('when sorting is passed', () => {
+      describe('when usage sorting is passed', () => {
         it('should sort the data by that value', () => {
           const { wrapper } = setup({
             columns,
             sortBy: {
+              name: 'Usage',
               key: 'usage',
               direction: SortDirection.descending,
-              name: 'Usage',
             },
           });
           const expected = 'simple_column_name_timestamp';
+          const actual = wrapper
+            .find('.table-detail-table .ams-table-row')
+            .at(0)
+            .find('.column-name')
+            .text();
+
+          expect(actual).toEqual(expected);
+        });
+      });
+
+      describe('when default sorting is passed', () => {
+        it('should sort the data by that value', () => {
+          const { wrapper } = setup({
+            columns,
+            sortBy: {
+              name: 'Default',
+              key: 'sort_order',
+              direction: SortDirection.ascending,
+            },
+          });
+          const expected = 'simple_column_name_string';
+          const actual = wrapper
+            .find('.table-detail-table .ams-table-row')
+            .at(0)
+            .find('.column-name')
+            .text();
+
+          expect(actual).toEqual(expected);
+        });
+      });
+
+      describe('when name sorting is passed', () => {
+        it('should sort the data by name', () => {
+          const { wrapper } = setup({
+            columns,
+            sortBy: {
+              name: 'Name',
+              key: 'name',
+              direction: SortDirection.descending,
+            },
+          });
+          const expected = 'simple_column_name_bigint';
           const actual = wrapper
             .find('.table-detail-table .ams-table-row')
             .at(0)
--- a/amundsen_application/static/js/components/ColumnList/index.spec.tsx
+++ b/amundsen_application/static/js/components/ColumnList/index.spec.tsx
@@ -7,7 +7,11 @@ import { Provider } from 'react-redux';
 import configureStore from 'redux-mock-store';
 import { mocked } from 'ts-jest/utils';
 
-import { notificationsEnabled } from 'config/config-utils';
+import { SortDirection } from 'interfaces';
+import {
+  notificationsEnabled,
+  getTableSortCriterias,
+} from 'config/config-utils';
 
 import globalState from 'fixtures/globalState';
 import ColumnList, { ColumnListProps } from '.';
@@ -19,6 +23,7 @@ import TestDataBuilder from './testDataBuilder';
 jest.mock('config/config-utils');
 
 const mockedNotificationsEnabled = mocked(notificationsEnabled, true);
+const mockedGetTableSortCriterias = mocked(getTableSortCriterias, true);
 const dataBuilder = new TestDataBuilder();
 const middlewares = [];
 const mockStore = configureStore(middlewares);
@@ -46,6 +51,18 @@ const setup = (propOverrides?: Partial<ColumnListProps>) => {
 };
 
 describe('ColumnList', () => {
+  mockedGetTableSortCriterias.mockReturnValue({
+    sort_order: {
+      name: 'Table Default',
+      key: 'sort_order',
+      direction: SortDirection.ascending,
+    },
+    usage: {
+      name: 'Usage Count',
+      key: 'usage',
+      direction: SortDirection.descending,
+    },
+  });
   mockedNotificationsEnabled.mockReturnValue(true);
 
   describe('render', () => {
@@ -101,7 +118,11 @@ describe('ColumnList', () => {
         it('should sort the data by that value', () => {
           const { wrapper } = setup({
             columns,
-            sortBy: { key: 'usage', direction: 'desc' },
+            sortBy: {
+              key: 'usage',
+              direction: SortDirection.descending,
+              name: 'Usage',
+            },
           });
           const expected = 'simple_column_name_timestamp';
           const actual = wrapper
--- a/amundsen_application/static/js/components/ColumnList/index.spec.tsx
+++ b/amundsen_application/static/js/components/ColumnList/index.spec.tsx
@@ -232,6 +232,39 @@ describe('ColumnList', () => {
       });
     });
 
+    describe('when columns with serveral stats including usage are passed', () => {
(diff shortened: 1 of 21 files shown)


amundsen_application/static/js/components/ColumnList/index.tsx
```diff
@@ -29,6 +29,9 @@ import {
 
 import './styles.scss';
 
+export type SortDirection = 'asc' | 'desc';
```
ttannis, line 32: nit: I have a preference for the full words over abbreviations so that nobody misinterprets what things are meant to do. For example, here I have to use the context clue of `SortDirection` to understand that `desc` means `descending` as opposed to maybe `description`. So in favor of `ascending` and `descending`

amundsen_application/static/js/pages/TableDetailPage/index.tsx
```diff
@@ -105,13 +126,26 @@ const ErrorMessage = () => {
...
-  TableDetailProps & RouteComponentProps<any>
+  TableDetailProps & RouteComponentProps<any>,
+  StateProps
 > {
   private key: string;
 
   private didComponentMount: boolean = false;
 
+  // @ts-ignore
+  state = {
```
ttannis, line 145: What issue does this cause that we have to ignore it? Does the fact that we're not initializing in a `constructor` method have any effect?

amundsen_application/static/js/pages/TableDetailPage/index.tsx
```diff
@@ -65,8 +66,28 @@ import './styles.scss';
...
+  },
+};
+const SORT_OPTIONS = [
+  {
+    name: 'Table Default',
+    type: 'sort_order',
+  },
+  {
+    name: 'Usage Count',
+    type: 'usage',
```
ttannis, line 86: Does this work right now as is? This looks like it works by needing a `usage` attribute on the column metadata, but usage is on the column stats.

Given this implementation -- unless I've missed some recent changes -- we will have to add logic in https://github.com/amundsen-io/amundsenfrontendlibrary/blob/master/amundsen_application/api/utils/metadata_utils.py#L68 to pull the column usage object from the stats list (so that it doesnt show up on the stats UI), and add its value directly on the column object under the name `usage`.

amundsen_application/static/js/components/ColumnList/index.tsx
Golodhros: Oh, actually I was working on exactly that!
Using an enum, check the updated PR!

amundsen_application/static/js/pages/TableDetailPage/index.tsx
Golodhros: Not sure!
I don't think so, usually this method is preferred.

amundsen_application/static/js/pages/TableDetailPage/index.tsx
Golodhros: Oh, I do some rework of the data inside Column list o add that info into the data itself.

amundsen_application/static/js/pages/TableDetailPage/index.tsx
Golodhros: Actually, it doesn't fail anymore. Updating it!

amundsen_application/static/js/components/common/Table/index.tsx
```diff
@@ -8,8 +8,14 @@ import { UpIcon, DownIcon } from '../SVGIcons';
 
 import './styles.scss';
 
-type TextAlignmentValues = 'left' | 'right' | 'center';
+// export type SortDirection = 'asc' | 'desc';
```
allisonsuarez, line 11: Why are these commented out?

Golodhros: I will implement it shortly!

amundsen_application/static/js/components/ColumnList/index.spec.tsx
```diff
@@ -96,6 +113,27 @@ describe('ColumnList', () => {
...
+        it('should sort the data by that value', () => {
+          const { wrapper } = setup({
+            columns,
+            sortBy: {
+              key: 'usage',
+              direction: SortDirection.descending,
+              name: 'Usage',
+            },
+          });
+          const expected = 'simple_column_name_timestamp';
```
allisonsuarez, line 127: So here are you testing if it was sorted correctly by checking just the first item to see if its the one that is supposed to be first?

amundsen_application/static/js/components/ColumnList/index.tsx
```diff
@@ -153,17 +197,26 @@ const ColumnList: React.FC<ColumnListProps> = ({
         name: item.name,
         database,
       },
-      usage: hasItemStats ? item.stats[0].stat_val : '',
+      sort_order: item.sort_order,
+      usage: hasItemStats ? +item.stats[0].stat_val : null,
```
ttannis, line 201: I don't think the api guarantees the column usage will be the first statistic, so perhaps we should have a helper method to intentionally search for the `column_usage` statistic and extract the value.

amundsen_application/static/js/components/ColumnList/index.tsx
```diff
@@ -153,17 +197,26 @@ const ColumnList: React.FC<ColumnListProps> = ({
...
+      name: item.name,
       isEditable: item.is_editable,
       editText,
       editUrl,
       index,
     };
   });
   const statsCount = formattedData.filter((item) => !!item.stats).length;
-  const hasStats = statsCount >= SHOW_STATS_THRESHOLD;
+  const hasStats =
```
ttannis, line 212: nit: Is this more accurately `hasUsage`?

amundsen_application/static/js/pages/TableDetailPage/index.tsx
ttannis: Just noticed it. Left a comment about it below.

ttannis:
Looks like tests are failing because of amundsencommon? cc @allisonsuarez since I think you were working on it. On my end only concern is my comment about reliable logic to extract the column usage from the stats.

feng-tao:
> Looks like tests are failing because of amundsencommon? cc @allisonsuarez since I think you were working on it. On my end only concern is my comment about reliable logic to extract the column usage from the stats.

let me take a look at the code as well. I was raising a question: could we sort based on a metric which defined by usage? We name column usage but other OSS could name it differently.

Also is @ttannis 's concern extracting column usage in stats requiring certain hack? do you think if we put query / column usage as separate node helps?

amundsen_application/static/js/components/ColumnList/index.tsx
Golodhros: Good call!
Right now this will work, but as we add more stats it will definitely break.
I will add something now.

amundsen_application/static/js/components/ColumnList/index.tsx
Golodhros: Yep, will change

amundsen_application/static/js/components/ColumnList/index.tsx
feng-tao: +1, currently we have single stat at Lyft, but we will ingest more stat later. I think we should either define the query / metric name in config for sorting or redefine the usage not part of stat. I wonder whether we could choose the former?

amundsen_application/static/js/components/ColumnList/index.tsx
```diff
@@ -76,6 +90,50 @@ type ExpandedRowProps = {
 };
 
 const SHOW_STATS_THRESHOLD = 1;
+const USAGE_STAT_TYPE = 'column_usage';
```
Golodhros, line 93: This would probably be the item to move to the config, however I am wary of creating that abstraction now, without knowing more about what it is to come.

ttannis: I can understand that. I am alright with it as it since the community will be able to follow existing examples to make this more flexible if/when needed.

ttannis:
> > Looks like tests are failing because of amundsencommon? cc @allisonsuarez since I think you were working on it. On my end only concern is my comment about reliable logic to extract the column usage from the stats.
> 
> let me take a look at the code as well. I was raising a question: could we sort based on a metric which defined by usage? We name column usage but other OSS could name it differently.
> 
> Also is @ttannis 's concern extracting column usage in stats requiring certain hack? do you think if we put query / column usage as separate node helps?

I wouldn't call it a hack, rather extra logic to explicitly search the stats array for `column_usage`. Having it as a separate node would make it such that we don't need that extra logic, but I still believe your original perspective that column usage is a statistic in the metadata is fair. The case here is that the UI design has chosen to display usage differently from other column statistics, so I think it is appropriate for the frontend to have to have go parse the data because of how our designs have chosen to display them. 

In short, no strong opinion on changing it!

ttannis (approved):
Pending passing CI checks.

amundsen_application/static/js/components/ColumnList/index.tsx
```diff
@@ -153,17 +212,26 @@ const ColumnList: React.FC<ColumnListProps> = ({
...
       isEditable: item.is_editable,
       editText,
       editUrl,
       index,
     };
   });
   const statsCount = formattedData.filter((item) => !!item.stats).length;
-  const hasStats = statsCount >= SHOW_STATS_THRESHOLD;
+  const hasUsageStat =
+    getTableSortCriterias().usage && statsCount >= SHOW_STATS_THRESHOLD;
```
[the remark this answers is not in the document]
Golodhros: If no element on the data has usage, we won't show the whole column, right.

amundsen_application/static/js/config/config-default.ts
```diff
@@ -162,6 +162,18 @@ const configDefault: AppConfig = {
...
+        sort_order: {
+          name: 'Table Default',
+          key: 'sort_order',
+          direction: SortDirection.ascending,
+        },
+        name: {
+          name: 'Alphabetical',
+          key: 'name',
+          direction: SortDirection.descending,
+        },
```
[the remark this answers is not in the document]
Golodhros: Yep, that change is on the private repo, only for us right now.

Golodhros:
> > > Looks like tests are failing because of amundsencommon? cc @allisonsuarez since I think you were working on it. On my end only concern is my comment about reliable logic to extract the column usage from the stats.
> > 
> > 
> > let me take a look at the code as well. I was raising a question: could we sort based on a metric which defined by usage? We name column usage but other OSS could name it differently.
> > Also is @ttannis 's concern extracting column usage in stats requiring certain hack? do you think if we put query / column usage as separate node helps?
> 
> I wouldn't call it a hack, rather extra logic to explicitly search the stats array for `column_usage`. Having it as a separate node would make it such that we don't need that extra logic, but I still believe your original perspective that column usage is a statistic in the metadata is fair. The case here is that the UI design has chosen to display usage differently from other column statistics, so I think it is appropriate for the frontend to have to have go parse the data because of how our designs have chosen to display them.
> 
> In short, no strong opinion on changing it!

Done!

PR #699, opened 2020-09-30
chore: Testing types for table input

Golodhros:
Signed-off-by: Marcos Iglesias <miglesiasvalle@lyft.com>

### Summary of Changes
Trying to type the table entry properly, relating the data with the columns

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/common/Table/index.spec.tsx
+++ b/amundsen_application/static/js/components/common/Table/index.spec.tsx
@@ -3,6 +3,7 @@
 
 import * as React from 'react';
 import { mount } from 'enzyme';
+import { mocked } from 'ts-jest/utils';
 
 import Table, { TableProps } from '.';
 
@@ -406,11 +407,11 @@ describe('Table', () => {
 
         beforeEach(() => {
           jest.spyOn(console, 'error');
-          console.error.mockImplementation(() => {});
+          mocked(console.error).mockImplementation(() => {});
         });
 
         afterEach(() => {
-          console.error.mockRestore();
+          mocked(console.error).mockRestore();
         });
 
         it('throws an error', () => {
--- a/amundsen_application/static/js/components/common/Table/index.spec.tsx
+++ b/amundsen_application/static/js/components/common/Table/index.spec.tsx
@@ -407,7 +407,7 @@ describe('Table', () => {
 
         beforeEach(() => {
           jest.spyOn(console, 'error');
-          mocked(console.error).mockImplementation(() => {});
+          mocked(console.error).mockImplementation(jest.fn);
         });
 
         afterEach(() => {
--- a/amundsen_application/static/js/components/common/Table/index.spec.tsx
+++ b/amundsen_application/static/js/components/common/Table/index.spec.tsx
@@ -404,6 +404,15 @@ describe('Table', () => {
       describe('when columns specify fields not in the data', () => {
         const { columns, data } = dataBuilder.withWrongData().build();
 
+        beforeEach(() => {
+          jest.spyOn(console, 'error');
+          console.error.mockImplementation(() => {});
+        });
+
+        afterEach(() => {
+          console.error.mockRestore();
+        });
+
         it('throws an error', () => {
           expect(() => {
             setup({
--- a/amundsen_application/static/js/components/common/Table/index.spec.tsx
+++ b/amundsen_application/static/js/components/common/Table/index.spec.tsx
@@ -400,6 +400,19 @@ describe('Table', () => {
           });
         });
       });
+
+      describe('when columns specify fields not in the data', () => {
+        const { columns, data } = dataBuilder.withWrongData().build();
+
+        it('throws an error', () => {
+          expect(() => {
+            setup({
+              data,
+              columns,
+            });
+          }).toThrow();
+        });
+      });
     });
 
     describe('options', () => {
--- a/amundsen_application/static/js/components/common/Table/index.tsx
+++ b/amundsen_application/static/js/components/common/Table/index.tsx
@@ -16,13 +16,17 @@ export enum TextAlignmentValues {
   right = 'right',
   center = 'center',
 }
-export interface TableColumn {
+export interface TableColumn<T> {
   title: string;
-  field: string;
+  field: T;
   horAlign?: TextAlignmentValues;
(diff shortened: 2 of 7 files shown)


ttannis:
Might need some more context on this one since I'm unsure what current use cases we are trying to support/make stronger typing for.

Golodhros:
> Might need some more context on this one since I'm unsure what current use cases we are trying to support/make stronger typing for.

Basically, I am aiming for the example on the storybook to break, so if a user passes a column which key is not on the data, it should break!

amundsen_application/static/js/components/common/Table/table.story.tsx
```diff
@@ -62,6 +66,9 @@ export const TableStates = () => (
         options={object('options', { isLoading: true })}
       />
     </StorySection>
+    <StorySection title="Loading Table">
+      <Table columns={columnsWithWrongData} data={dataWithWrongData} />
```
Golodhros, line 70: Ideally, this should break!

amundsen_application/static/js/components/common/Table/index.tsx
```diff
@@ -37,13 +41,15 @@ export interface TableOptions {
...
 export interface TableProps {
+  data: RowData[];
   columns: TableColumn[];
-  data: any[];
   options?: TableOptions;
 }
 
 const DEFAULT_EMPTY_MESSAGE = 'No Results';
 const EXPAND_ROW_TEXT = 'Expand Row';
+const INVALID_DATA_ERROR_MESSAGE =
```
gagoar, line 51: is the `Error` word needed here?

if the Error is serialized ti will add the `Error` string by itself.

Golodhros: Makes sense, thanks!

amundsen_application/static/js/components/common/Table/index.tsx
```diff
@@ -68,6 +74,15 @@ type EmptyRowProps = {
...
   ALIGNEMENT_TO_CLASS_MAP[alignment];
 
+const fieldIsDefined = (field, row) => row[field] !== undefined;
+const checkValidData = (data, fields) => {
+  fields.forEach((field) => {
+    if (data.filter(fieldIsDefined.bind(null, field)).length === 0) {
+      throw new Error(INVALID_DATA_ERROR_MESSAGE);
+    }
+  });
+};
```
gagoar, line 84: I would suggest something along the lines of:
```suggestion
type Some = string | number | boolean | symbol | bigint | object; 
interface ValidData =   Record<string, Some> // to sass out the undefined | null;
// guard
const checkValidData = (data: unknown[], fields): is ValidData[] => {
   return data.filter(fieldIsDefined.bind(null, field)).length === 0
}
```

amundsen_application/static/js/components/common/Table/index.tsx
```diff
@@ -182,6 +197,8 @@ const Table: React.FC<TableProps> = ({
   );
 
   if (data.length) {
+    checkValidData(data, fields);
```
gagoar, line 200:
```suggestion
   if (!checkValidData(data, fields)) {
   throw new Error(INVALID_DATA_ERROR_MESSAGE);
 }
```

gagoar: what you get at the end is, that you can make ValidData type as interesting as you want so, if the type is validated you would have a type already casted and you will be able to do more things with it.

amundsen_application/static/js/components/common/Table/testDataBuilder.tsx
```diff
@@ -44,6 +44,36 @@ function TestDataBuilder(config = {}) {
...
+          field: 'type',
+        },
+        {
+          title: 'Value',
+          field: 'value',
+        },
+        {
+          title: 'Bad Field',
+          field: 'badfield',
+        },
```
gagoar, line 70: Just for me to fully understand the context. why is this a badFiled? is the string inside the `field` key what needs to be validated?

amundsen_application/static/js/components/common/Table/index.tsx
gagoar: to read more about guards, you can go [here](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)

Golodhros: Thanks man!

amundsen_application/static/js/components/common/Table/testDataBuilder.tsx
Golodhros: because is not present on any of the objects of the data array

amundsen_application/static/js/components/common/Table/index.tsx
```diff
@@ -55,19 +73,26 @@ const ALIGNEMENT_TO_CLASS_MAP = {
...
+  data: unknown[],
+  fields: string[]
+): data is ValidData[] => {
+  let isValid = true;
+
+  fields.forEach((field) => {
+    if (data.filter(fieldIsDefined.bind(null, field)).length === 0) {
+      isValid = false;
+    }
+  });
```
gagoar, line 91: just a nit for speed's sake. 

You might want to use a normal loop (I don't like them but in this case might be useful) to break when the false happens instead of keep iterating.  meaning, you want to break if anything turns false. 

for that. maybe 2 options. the aforementioned for loop so you can use break;

or in a functional way, you would have to flip the logic and use some instead of forEach.

amundsen_application/static/js/components/common/Table/index.spec.tsx
```diff
@@ -400,6 +401,28 @@ describe('Table', () => {
           });
         });
       });
+
+      describe('when columns specify fields not in the data', () => {
+        const { columns, data } = dataBuilder.withWrongData().build();
+
+        beforeEach(() => {
+          jest.spyOn(console, 'error');
+          mocked(console.error).mockImplementation(() => {});
```
gagoar, line 410:
```suggestion
          mocked(console.error).mockImplementation(js.fn);
```

allisonsuarez (approved):
LGTM, I like that we are getting more informative errors

ttannis:
Thanks for this PR, learned some helpful patterns.

PR #702, opened 2020-10-01
fix: Align and format column stats

dikshathakur3119:
### Summary of Changes

Each column may have different column statistics like mean, median, etc. In this PR:
1. Added Numerical separator in the values using [Intl.NumberFormat().format(number)], this format can be configured, default is US format (e.g 1,234,567). 

2. Right aligning numbers

3. Monospace font for the numbers 

See attached screenshots for different formats.
### Tests

Added test to see if configured number format is used to format column stats.

### Documentation

Nothing but can add if these changes looks fine.

image
image
image

### CheckList


amundsen_application/static/js/components/ColumnList/ColumnStats/index.tsx
```diff
@@ -23,10 +24,12 @@ const ColumnStatRow: React.FC<ColumnStatRowProps> = ({
   stat_type,
   stat_val,
 }: ColumnStatRowProps) => {
+  const numberVal = +stat_val;
+
   return (
     <div className="column-stat-row">
       <div className="stat-name body-3">{stat_type.toUpperCase()}</div>
```
Golodhros, line 31: Could you change this to use the new typography extends?
You can use this as an example: https://github.com/amundsen-io/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/components/ColumnList/styles.scss#L84

amundsen_application/static/js/components/ColumnList/ColumnStats/styles.scss
```diff
@@ -45,6 +45,8 @@
         .stat-value {
           flex: 100px 1 1;
           padding-right: 4px;
```
Golodhros, line 47: Nit: could you move this to use $spacer-1/2 ?

amundsen_application/static/js/utils/numberUtils.ts
```diff
@@ -0,0 +1,9 @@
+import { getNumberFormat } from 'config/config-utils';
+
+export function formatNumber(value) {
+  const numberformat = getNumberFormat();
+  if (numberformat) {
+    return Intl.NumberFormat(numberformat).format(value);
+  }
+  return Intl.NumberFormat().format(value);
+}
```
Golodhros, line 9: I think you could do:
```suggestion
export function formatNumber(value) {
  const numberformat = getNumberFormat() || undefined;

    return Intl.NumberFormat().format(value);
}
```

Golodhros (commented):
Looks good!

Just some minor comments and clean ups.

amundsen_application/static/js/config/config-types.ts
```diff
@@ -17,6 +17,7 @@ export interface AppConfig {
   userIdLabel?: string /* Temporary configuration due to lacking string customization/translation support */;
   issueTracking: IssueTrackingConfig;
   logoPath: string | null;
+  numberFormat: string | null;
```
ttannis, line 20: `Intl.NumberFormat` takes more options than just a string and also has multiple parameters. Our configuration needs to match that in order to allow folks to use the full capabilities: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat

In addition, I would suggest making this configuration more flexible to support the various types of number formatting. You'll see in the documentation above the number formatting has different styles, like for showing `decimal` (default), `percent`, `currency`, and `unit` for different units of measurement. As a tool meant to show data, I imagine we will have a use case in the future. So perhaps we could make this configuration a dictionary that supports being able to configure all of the styles:

Use Case: Imagine we have a future use case with percentages, which will need different formatting options. 
```
// Used in code would be: Intl.NumberFormat('en-US', { ...some optional options... })
numberFormatting: {
   decimal: {
        locale: 'en-US',
        options: {...some optional options...}
    },
   percentage: {
        locale: 'en-US',
        options: {...some optional options...}
    }
}
```

What do you think of structuring the configuration as something like above? We can refine the exact configuration the main point here about creating something that will be able to be more easily extended in a backwards compatible manner to cover number formatting use cases that are feasible for us to encounter in the future.

Lastly, all new configurations should have doc added: https://github.com/amundsen-io/amundsenfrontendlibrary/blob/master/docs/application_config.md

dikshathakur3119: yes you are right we can modify configurations so that it provides flexibility for future use cases if we may want to show different units in future. Only reason I did not use full capabilities of `Intl.NumberFormat` is that in current scenario column statistics only includes absolute values, like mean , median, stddev, etc. Data in any column with column statistics will always have numbers only even if it is storing a percentage or currency. But I can add the flexibility to make it easily extendable.

Golodhros (approved):
LGTM

dikshathakur3119:
🔒 will be creating a new branch to make these changes. Lot of chaos created because I accidentally removed package.json file while merging master and then forgot to sign a commit as well.

PR #706, opened 2020-10-05
feat: surface column level badges

allisonsuarez:
### Summary of Changes

Surface column level badges from metadata and display them as not clickable badges in the FE column list table within the table detail page.

### Tests

Added tests for BadgeList, ColumnList, and TextUtils.

### Documentation

N/A

### CheckList


amundsen_application/static/js/components/ColumnList/index.tsx
```diff
@@ -216,6 +229,7 @@ const ColumnList: React.FC<ColumnListProps> = ({
       sort_order: item.sort_order,
       usage: getUsageStat(item),
       stats: hasItemStats ? item.stats[0] : null,
+      badges: hasColumnBadges ? item.badges : [],
```
Golodhros, line 232: Yeah, I was thinking yesterday how we should do a local check for hasItemBadges if we wanted to default it to null

amundsen_application/static/js/components/ColumnList/index.tsx
```diff
@@ -274,6 +288,18 @@ const ColumnList: React.FC<ColumnListProps> = ({
...
   }
 
+  if (hasColumnBadges) {
+    formattedColumns = [
+      ...formattedColumns,
+      {
+        title: 'Badges',
+        field: 'badges',
+        horAlign: TextAlignmentValues.left,
+        component: (values) => (values ? <BadgeList badges={values} /> : null),
```
Golodhros, line 298: If we default to [], then this is going to be called always, so we won't need to use the ternary.
In this ocassion the BadgeList is handling the empty list case.

amundsen_application/static/js/components/common/BadgeList/index.tsx
```diff
@@ -23,6 +26,14 @@ const BadgeList: React.FC<BadgeListProps> = ({ badges }: BadgeListProps) => {
...
+            return (
+              <Flag
+                text={badgeConfig.displayName}
+                labelStyle={badgeConfig.style}
+              />
+            );
+          }
         }
         return (
           <ClickableBadge
```
Golodhros, line 39: Thinking twice about this, maybe it would make sense to separate the concept of a Badge (interactive) vs a Flag, so creating a FlagList to handle all this case.

WDYT?

amundsen_application/static/js/components/common/Flag/index.tsx
```diff
@@ -27,6 +28,13 @@ export function convertText(str = '', caseType: string): string {
       return `${str.charAt(0).toUpperCase()}${str.slice(1).toLowerCase()}`;
     case CaseType.UPPER_CASE:
       return str.toUpperCase();
+    case CaseType.TITLE_CASE:
+      const splitStr = str.toLowerCase().split(' ');
+      for (let i = 0; i < splitStr.length; i++) {
+        splitStr[i] =
+          splitStr[i].charAt(0).toUpperCase() + splitStr[i].substring(1);
+      }
+      return splitStr.join(' ');
```
Golodhros, line 37: I think this should be extracted into an independent, well tested utils function

amundsen_application/static/js/components/common/Table/styles.scss
```diff
@@ -82,20 +84,22 @@ $row-bottom-border-color: $gray10;
   overflow: visible;
 
   &:first-child {
-    padding-left: $spacer-3;
+    padding-left: $spacer-3 !important;
```
Golodhros, line 87: I wonder if there would be a way of avoiding this !important, maybe adding an extra class would be better.

Golodhros (commented):
Looks good!

Some minor comments.

amundsen_application/static/js/components/common/BadgeList/index.tsx
allisonsuarez: I feel like all of the functionality is so similar (and domain wise we refer to column level and table badges as exactly the same thing), that it might be worth keeping the list the same and them differentiating between clickable and not clickable badges in this point, I do believe that we could define those components differently in this file. Maybe we could do a refactor where we pass the on click action for triggering search as an optional parameter, and if it is present then we proceed to make the badge clickable since if you look in the ClickableBadge component you'll notice there is a Flag in there.

Golodhros: That sounds good too!

allisonsuarez: Wait now looking at it I don't believe this code would work with multiple badges in one badgeList??

allisonsuarez: Actually no because it is within that map function I forgot about

amundsen_application/static/js/components/common/BadgeList/index.tsx
```diff
@@ -4,36 +4,122 @@
...
+    <span className={`static-badge flag label label-${style}`}>
+      <div className={`badge-overlay-${style}`}>
+        {displayName}
+      </div>
+    </span>);
+};
+
+const ActionableBadge: React.FC<ActionableBadgeProps> = ({ style, displayName, action }: ActionableBadgeProps) => {
+  return (
+    <span className="actionable-badge" onClick={action}>
```
Golodhros, line 51: I think this should be a button instead of a span

amundsen_application/static/js/components/common/BadgeList/index.tsx
```diff
@@ -4,36 +4,122 @@
...
+            if (badge.category === COLUMN_BADGE_CATEGORY) {
+              return (<StaticBadge
+                style={badgeConfig.style}
+                displayName={badgeConfig.displayName}
+                key={`badge-${index}`}
+              />);
+            }
+          }
+          if (badge.category !== COLUMN_BADGE_CATEGORY) {
+            let onClick = (e) => {
```
Golodhros, line 85: Let's rename this to handleClick, following our new naming convention.

Also, this handler should be out of the map, so we don't create a new handler for each badge of the list

amundsen_application/static/js/components/common/BadgeList/index.tsx
```diff
@@ -4,36 +4,122 @@
...
+              const badgeText = originalBadgeName;
+              logClick(e, {
+                target_type: 'badge',
+                label: badgeText,
+              });
+              this.props.searchBadge(convertText(badgeText, CaseType.LOWER_CASE));
+            };
+            return (<ActionableBadge
```
Golodhros, line 93: Could be this inside the "else" for the check on line 76?

amundsen_application/static/js/components/common/BadgeList/index.tsx
```diff
@@ -4,36 +4,122 @@
...
-          />
-        );
-      })}
-    </span>
+        )}
+      </span>);
+  }
+}
+
+export const mapDispatchToProps = (dispatch: any) => {
```
Golodhros, line 107: I would love us to push this still upwards one more level, so BadgeList and Badges are presentational components.

Not urgent to do, but this is definitely a step on the right direction!

amundsen_application/static/js/config/config-utils.ts
```diff
@@ -74,6 +74,34 @@ export function getFilterConfigByResource(
...
+      for (let i = 0; i < splitStr.length; i++) {
+        splitStr[i] =
+          splitStr[i].charAt(0).toUpperCase() + splitStr[i].substring(1);
+      }
+      return splitStr.join(' ');
+    default:
+      return str;
+  }
+}
+
```
Golodhros, line 104: I think this should go into a utils file like 'utils/textUtils.ts`

amundsen_application/static/js/components/common/BadgeList/index.tsx
allisonsuarez: I agree, I'll refactor it in a separate PR when i finalize the badges work

amundsen_application/static/js/components/common/BadgeList/index.tsx
allisonsuarez: no because then badges that came in with tag_name from search would be excluded

amundsen_application/static/js/components/common/BadgeList/index.spec.tsx
```diff
@@ -2,13 +2,58 @@
...
+  };
+
+  const testState = globalState;
+  testState.tableMetadata.tableData.badges = badges;
+  const wrapper = mount<BadgeListProps>(
+    <Provider store={mockStore(testState)}>
+      <BadgeList {...props} />
+    </Provider>
+  );
+  // const wrapper = shallow(<BadgeList {...props} />);
```
Golodhros, line 53: Let's remove this before merging

amundsen_application/static/js/components/common/BadgeList/index.tsx
```diff
@@ -3,37 +3,132 @@
...
+    <span className="actionable-badge" onClick={action}>
+      <StaticBadge style={style} displayName={displayName} />
+    </span>
+  );
+};
+
+export class BadgeList extends React.Component<BadgeListProps> {
+  idx = 0;
+
+  onClick = (e) => {
```
Golodhros, line 63: handleClick instead

amundsen_application/static/js/components/common/BadgeList/index.tsx
```diff
@@ -3,37 +3,132 @@
...
+    });
+    this.props.searchBadge(convertText(badgeText, CaseType.LOWER_CASE));
+  };
+
+  render() {
+    return (
+      <span className="badge-list">
+        {this.props.badges.map((badge, index) => {
+          let badgeConfig;
+          // let originalBadgeName;
```
Golodhros, line 79: to clean?

amundsen_application/static/js/utils/textUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+export enum CaseType {
+  LOWER_CASE = 'lowerCase',
+  SENTENCE_CASE = 'sentenceCase',
+  UPPER_CASE = 'upperCase',
+  TITLE_CASE = 'titleCase',
+}
+
+export function convertText(str = '', caseType: string): string {
```
Golodhros, line 11:
```suggestion
export function convertText(str = '', caseType: CaseType): string {
```

Golodhros (approved):
LGTM

Thanks for your effort here!!

PR #739, opened 2020-10-29
feat: Analytics Tracking middleware

dorianj:
### Summary of Changes
This PR extends our current analytics approach, allowing for multiple and pluggable analytics services to be used and establishing a preferred way of tracking events by reusing our Redux actions.

We propose using the popular [Analytics project](https://github.com/DavidWells/analytics) to enable the pluggable architecture. 

We also propose using a Redux Middleware to manage the analytics events as side effects of user actions. In this PR we only include a test event, but the plan will be to extend it and eventually make it the only way of tracking analytics events.

### Tests
No tests added for analytics, as they are side effects that don't interfere with the behavior of the application.

### Documentation
Added documentation about how to set up the analytics for your app.

### CheckList


amundsen_application/static/js/ducks/bookmark/types.ts
```diff
@@ -11,6 +11,7 @@ export interface AddBookmarkRequest {
     resourceKey: string;
     resourceType: ResourceType;
   };
+  meta: any;
```
Golodhros, line 14: We would need to create an interface for meta objects

amundsen_application/static/js/utils/analytics.ts
```diff
@@ -0,0 +1,29 @@
+import Analytics, { AnalyticsInstance } from 'analytics';
+
+import * as ConfigUtils from 'config/config-utils';
+
+let _analyticsInstance;
```
Golodhros, line 5: Let's just remove the "_"

amundsen_application/static/templates/index.html
```diff
@@ -18,18 +18,13 @@
...
     {% endif %}
 
-    {% if <%= htmlWebpackPlugin.options.config.google.enabled%> %}
-      {% include 'fragments/google-analytics-loader.html' %}
-    {% endif %}
-
     <%= htmlWebpackPlugin.tags.headTags %>
+
+    <!-- Do not remove: some analytics scripts attach to this. -->
+    <script></script>
```
Golodhros, line 24: Could we add an id to this script to note why do we have it?

docs/application_config.md
```diff
@@ -37,9 +37,28 @@ This config allows you to specify various date formats across the app. There are
...
+  plugins: [
+    googleAnalytics({
+      trackingId: 'UA-121991291',
+      sampleRate: 100
+    }),
+  ],
+}
+```
+
+All [`@analytics/` plugins](https://github.com/DavidWells/analytics#analytic-plugins) are supported.
```
Golodhros, line 61: Thanks for the docs!

amundsen_application/static/js/config/config-custom.ts
```diff
@@ -2,15 +2,19 @@
...
     showAllTags: true,
   },
-  google: {
-    enabled: false,
-    key: 'default-key',
-    sampleRate: 100,
+  analytics: {
+    plugins: [
+      segment({
+        writeKey: 'MhibS57o1RGYs3ncYfeALSTbawRs1F25',
```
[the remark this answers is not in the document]
dorianj: oops, didn't mean to commit this

amundsen_application/static/js/ducks/bookmark/reducer.ts
```diff
@@ -135,15 +143,21 @@ export default function reducer(
         },
       };
     case GetBookmarksForUser.SUCCESS:
+      const bookmarksForUserResponse = <GetBookmarksForUserResponse>action;
```
dorianj, line 146: @Golodhros this PR pokes the beehive by revealing some latent non-null checks, and I'm trying to fix here.

I'm pretty uncertain of how clean this approach. I think it would be ideal to determine success higher up (perhaps using status codes), then you either have a fully-complete response with payload, or you have a failure. Having the presence of certain keys be dependent on other keys seems like a parsing mess, and while this certainly pleases the type-checker, it feels pretty fraught with errors. Definitely too big to fix in this PR, but is this the right temporary fix?

dorianj:
@Golodhros I addressed your feedback, removed cruft, and added typings. I also added some tricky fixes for some latent type errors -- asked a question about it.

One other question I have: this is a breaking change vs how google analytics was included before. I think that doing automatic compatibility bridging would not be worth the effort, because of the new requirement to install packages (it would not be great to just install GA package for everyone, even if not used, but it is certainly an option). What is the right way to communicate this change to the community?

amundsen_application/static/js/ducks/bookmark/reducer.ts
Golodhros: I think it looks good for now. 
We have plenty of these being tracked in our betterer, hopefully we'll get to them soon!

amundsen_application/static/js/interfaces/Analytics.ts
```diff
@@ -0,0 +1,4 @@
+export interface AnalyticsEvent {
+  name: string;
+  properties: { [prop: string]: any };
```
Golodhros, line 3: Nit: I think that 'unknown' would work as well and would avoid adding an 'any'

Golodhros:
> @Golodhros I addressed your feedback, removed cruft, and added typings. I also added some tricky fixes for some latent type errors -- asked a question about it.
> 
> One other question I have: this is a breaking change vs how google analytics was included before. I think that doing automatic compatibility bridging would not be worth the effort, because of the new requirement to install packages (it would not be great to just install GA package for everyone, even if not used, but it is certainly an option). What is the right way to communicate this change to the community?

Yes, it is a breaking change.
We had several conversations about how to go about these, but not sure what was the final recommendation. Let me reach out to the maintainer group and see what they think!

amundsen_application/static/js/utils/analytics.ts
dorianj: done

amundsen_application/static/js/ducks/bookmark/reducer.ts
dorianj: great!

feng-tao:
Given it is a breaking change, I think we should communicate through the #amundsen channel instead of the private one.

amundsen_application/static/js/interfaces/Analytics.ts
dorianj: Ah, good point, changed.

dorianj:
Message for #amundsen:

> Hi all. In the newest version of amundsenfrontendlib, there is a breaking change in that we have removed the `google` configuration parameter that was previously used to include Google Tag Manager. There is now pluggable infrastructure allowing you to install whichever analytics library you wish; please see the documentation page (link) with details of how to integrate GTM or other analytics libraries into Amundsen.

Golodhros:
Hey @dorianj, I will do a test run of this tomorrow, hopefully we can merge it this week!

amundsen_application/static/js/interfaces/Analytics.ts
```diff
@@ -1,4 +1,4 @@
 export interface AnalyticsEvent {
   name: string;
-  properties: { [prop: string]: unknown };
+  payload: { [prop: string]: unknown };
```
dorianj, line 3: curious why this change? `payload` already has meaning elsewhere, `properties` isn't great but it is the normal thing in analyitcs

Golodhros: Yeah, I changed it because it wasn't getting mapped on the analytics middleware, as we were expecting payload.

Here we are using the redux style of name/payload, so the meta.analytics is a redux action as well.
We kept the properties denomination on the last step (trackEvent)

dorianj: ahhh gotcha

dorianj:
@Golodhros From my POV, this is ready to go. If you agree, can you merge it? If not, what else do you think we need to do?

Golodhros:
> @Golodhros From my POV, this is ready to go. If you agree, can you merge it? If not, what else do you think we need to do?

Yes! Me too!
I am trying to test it first in our system, but I bumped on an unrelated issue and I haven't been able to deploy it. I will try to do it soon and merge this, sorry for the delay!

dorianj:
@Golodhros did you get a chance to do the testing? I'm happy to rebase this and fix the merge conflict if it lgty

Golodhros:
> @Golodhros did you get a chance to do the testing? I'm happy to rebase this and fix the merge conflict if it lgty

You won't believe it but I am still trying to test it. Having some deployment issues here.

Golodhros:
Done testing, looks great!
Could you rebase it? Thanks!

dorianj:
@Golodhros rebased :D

Golodhros:
> @Golodhros rebased :D

Sorry, could you try again @dorianj ?

dorianj:
@Golodhros ok, I rebased and removed the changes from all 3 edits to `package-lock.json`, then added a new commit that captures the run of `npm install --package-lock-only`, which I think is what we want

Golodhros (approved):
LGTM

dorianj:
🎉 thanks @Golodhros for the collab here!

PR #740, opened 2020-11-01
refactor: Made preview and announcements configurable through env

joshthoward:
### Summary of Changes

Added environment variables to configure preview and announcement clients. This will enable these clients to be configured through Helm rather than updating the setup.py script.

### Tests

There were no previous tests for the announcement client or preview client hooks. These were added. 

### Documentation

No documentation was added yet, if this looks like a direction that the core team wants to move in, I will update the docs in the parent repo.

### CheckList

- [-] PR title addresses the issue accurately and concisely. Example: "Updates the version of Flask to v1.0.2"
- [-] PR includes a summary of changes, including screenshots of any UI changes.
- [-] PR adds unit tests, updates existing unit tests, **OR** documents why no test additions or modifications are needed.
- [-] In case of new functionality, my PR adds documentation that describes how to use it.
- [-] PR passes all tests documented in the [developer guide](https://github.com/lyft/amundsenfrontendlibrary/blob/master/docs/developer_guide.md#testing)


joshthoward:
Also, the flake8 config in setup.cfg was modified to fix this [issue](https://gitlab.com/pycqa/flake8/-/issues/466). W503 was ignored to remain PEP8 compliant.

ttannis:
Just commenting as this is a work in progress, but the ideal thing to do here would be to *add* support for the clients to be configured through env variables, but *keep* existing support so that we don't force a breaking change on everyone else. 

This does get me thinking that perhaps we can start an initiative to consolidate our configurations (with input from the community), make backwards compatible changes one by one, and then finally deprecate the older configurations in a major release w/ official documentation. I will raise this in the community channel.

joshthoward:
@ttannis That's a fair point. Wasn't sure how sensitive the team was to breaking changes. This has been resolved. Otherwise, the PR is ready to go. It's labeled WIP due to the need for a documentation update. Happy to make that update if we want to keep the changes.

joshthoward:
@ttannis RE: consolidating configurations... 

I think the goal would be to load the LocalConfig class variables from the environment so that they can be configured via Helm. @verdan had the idea to replace some frontend configurations by pushing the enabled flags from the backend. Action logging would have to change due to the fact that the entrypoint is loaded outside of the application context.

joshthoward:
@ttannis Any further feedback on this? I think it's ready to review and possibly close.

feng-tao:
cc @Golodhros @allisonsuarez @dikshathakur3119 in case you have time. I could take a look later this week as well.

feng-tao:
I think setting through config than using entry point is better because most of people actually use FE docker image to deploy which they can't (hard) to change the entry point in this case.

If we could keep the entry point and add the config, we could slowly migrate over the config base approach to use announcement and preview client? @ttannis wdyt?

joshthoward:
@feng-tao, @ttannis

FYI: I updated the PR since @ttannis's last comment to make the changes backwards compatible. Existing configurations will work and will be overridden by the environment variables if set.

amundsen_application/api/preview/v0.py
```diff
@@ -29,17 +27,18 @@
...
     try:
-        if PREVIEW_CLIENT_INSTANCE is None and PREVIEW_CLIENT_CLASS is not None:
-            PREVIEW_CLIENT_INSTANCE = PREVIEW_CLIENT_CLASS()
-
         if PREVIEW_CLIENT_INSTANCE is None:
-            payload = jsonify({'previewData': {}, 'msg': 'A client for the preview feature must be configured'})
-            return make_response(payload, HTTPStatus.NOT_IMPLEMENTED)
+            if (app.config['PREVIEW_CLIENT_ENABLED']
+                    and app.config['PREVIEW_CLIENT'] is not None):
+                PREVIEW_CLIENT_CLASS = import_string(app.config['PREVIEW_CLIENT'])
```
feng-tao, line 36: may want to add an exception handle in case it doesn't work.

amundsen_application/api/preview/v0.py
```diff
@@ -29,17 +27,18 @@
...
+                PREVIEW_CLIENT_INSTANCE = PREVIEW_CLIENT_CLASS()
+            else:
+                payload = jsonify({'previewData': {}, 'msg': 'A client for the preview feature must be configured'})
```
feng-tao, line 39: may worth add a note/todo given there are two ways(entry point, config env) to do the same thing. Based on my current new experience with k8s, using config variable is much convenient. The only concern would be migration (e.g lyft case)

feng-tao (commented):
lgtm overall with a few nits, would be good to update any related doc (could be a different pr). @ttannis let me know if you have a chance to take a look.

amundsen_application/api/preview/v0.py
feng-tao: cc @allisonsuarez @ttannis

feng-tao (approved):

allisonsuarez:
@joshthoward can you take a look at the comments above before merging?

joshthoward:
I'll try to take a look tonight. I've been out of pocket this week.

amundsen_application/api/preview/v0.py
joshthoward: `import_string` raises an error, not an exception, if the module cannot be located. So no can do :(

amundsen_application/api/preview/v0.py
joshthoward: I will add a warning.

joshthoward:
@feng-tao, @allisonsuarez should be ready to go.

setup.cfg
```diff
@@ -3,7 +3,7 @@ format = pylint
 exclude = .svc,CVS,.bzr,.hg,.git,__pycache__,venv,venv3,node_modules
 max-complexity = 10
 max-line-length = 120
-ignore = NONE
+ignore = W503
```
feng-tao, line 6: just found this, why do we ignore this one?

joshthoward: Both W503 and W504 are enabled by default and they contradict one another. The error raised in W503 is planned to be best practice by Flake 8. Therefore we should ignore it. 
 
See these links
- https://gitlab.com/pycqa/flake8/-/issues/466
- https://stackoverflow.com/questions/57074300/what-is-the-recommended-way-to-break-long-if-statement-w504-line-break-after-b

feng-tao: i c

PR #748, opened 2020-11-02
fix: accept non int stats values

allisonsuarez:
### Summary of Changes

Changed logic so non numerical stats values will not be converted to int.
https://github.com/amundsen-io/amundsen/issues/778

### Tests

test cases where stats values are dates.

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/ColumnList/ColumnStats/index.tsx
+++ b/amundsen_application/static/js/components/ColumnList/ColumnStats/index.tsx
@@ -4,7 +4,7 @@
 import * as React from 'react';
 
 import { TableColumnStats } from 'interfaces/index';
-import { formatNumber } from 'utils/numberUtils';
+import { formatNumber, isNumber } from 'utils/numberUtils';
 import { getStatsInfoText } from '../utils';
 
 import { COLUMN_STATS_TITLE } from '../constants';
@@ -27,7 +27,9 @@ const ColumnStatRow: React.FC<ColumnStatRowProps> = ({
   return (
     <div className="column-stat-row">
       <div className="stat-name body-3">{stat_type.toUpperCase()}</div>
-      <div className="stat-value">{Number(stat_val) ? formatNumber(+stat_val) : stat_val}</div>
+      <div className="stat-value">
+        {isNumber(stat_val) ? formatNumber(+stat_val) : stat_val}
+      </div>
     </div>
   );
 };
--- a/amundsen_application/static/js/components/ColumnList/ColumnStats/index.tsx
+++ b/amundsen_application/static/js/components/ColumnList/ColumnStats/index.tsx
@@ -24,12 +24,10 @@ const ColumnStatRow: React.FC<ColumnStatRowProps> = ({
   stat_type,
   stat_val,
 }: ColumnStatRowProps) => {
-  const numberVal = +stat_val;
-
   return (
     <div className="column-stat-row">
       <div className="stat-name body-3">{stat_type.toUpperCase()}</div>
-      <div className="stat-value">{formatNumber(numberVal)}</div>
+      <div className="stat-value">{Number(stat_val) ? formatNumber(+stat_val) : stat_val}</div>
     </div>
   );
 };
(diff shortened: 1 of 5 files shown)


amundsen_application/static/js/components/ColumnList/ColumnStats/index.tsx
```diff
@@ -24,12 +24,10 @@ const ColumnStatRow: React.FC<ColumnStatRowProps> = ({
   stat_type,
   stat_val,
 }: ColumnStatRowProps) => {
-  const numberVal = +stat_val;
-
   return (
     <div className="column-stat-row">
       <div className="stat-name body-3">{stat_type.toUpperCase()}</div>
-      <div className="stat-value">{formatNumber(numberVal)}</div>
+      <div className="stat-value">{Number(stat_val) ? formatNumber(+stat_val) : stat_val}</div>
```
ttannis, line 30: The approaches mentioned here are more ideal/standard ways of checking this:
https://stackoverflow.com/questions/175739/built-in-way-in-javascript-to-check-if-a-string-is-a-valid-number

allisonsuarez: I am getting a type error when trying to use isNaN with Ts because it is expecting a number type

ttannis: Oh, here's another reference: https://stackoverflow.com/questions/42120046/typescript-isnan-only-accepts-a-number
```
if (typeof expectedValue === "string" && !Number.isNaN(Number(expectedValue))) {
    expectedValue = Number(expectedValue);
}
```

amundsen_application/static/js/components/ColumnList/ColumnStats/index.tsx
```diff
@@ -24,12 +24,12 @@ const ColumnStatRow: React.FC<ColumnStatRowProps> = ({
...
 }: ColumnStatRowProps) => {
-  const numberVal = +stat_val;
-
   return (
     <div className="column-stat-row">
       <div className="stat-name body-3">{stat_type.toUpperCase()}</div>
-      <div className="stat-value">{formatNumber(numberVal)}</div>
+      <div className="stat-value">
+        {isNumber(stat_val) ? formatNumber(+stat_val) : stat_val}
+      </div>
```
Golodhros, line 32: Could we extend the tests of ColumnStats to cover for this logic?

allisonsuarez: I tested the two functional individually wasn't sure if I should add tests to ColumsStats because I though those might be too specific to the logic rather than the presentational aspect of the component.

amundsen_application/static/js/utils/numberUtils.ts
```diff
@@ -12,3 +12,9 @@ export function formatNumber(value) {
   }
   return Intl.NumberFormat().format(value);
 }
+
+export function isNumber(value) {
+  if (typeof value == 'string') {
```
ttannis, line 17: We can type value as a string, and just have `return !Number.isNaN(Number(value));` We wouldn't need the check because code won't build if someone uses it incorrectly and passes something that isn't a string, and we would also want this method to always return something.

As is, someone can get `isNumber(100)` would return undefined

nit: we can also type the return type as a boolean.  `isNumber(value: string): boolean`

amundsen_application/static/js/components/ColumnList/ColumnStats/index.tsx
Golodhros: I see, however, if we ever change our logic to figure this out, this part would not be tested.

allisonsuarez: I see what you mean, so I should check if the correct text is displayed then?

Golodhros: Yes, please!

amundsen_application/static/js/components/ColumnList/ColumnStats/testDataBuilder.ts
```diff
@@ -135,6 +135,26 @@ function TestDataBuilder(config = {}) {
 
     return new this.Klass(attr);
   };
+  this.withNonNumericStats = () => {
```
Golodhros, line 138: Yay! Test Data Builders!

Golodhros (approved):
LGTM

Thanks for taking the time to add those tests!

PR #757, opened 2020-11-06
feat: Expose config for mTLS certificates.

davcamer:
### Summary of Changes

Provide Env Vars to configure a client cert for mTLS.

A client certificate can be specified to present with requests to
metadata and search services for mTLS. The configuration is given
as a pair of paths to the PEM encoded cert and key files. This is
described in more detail in the requests documentation:
https://requests.readthedocs.io/en/master/user/advanced/#client-side-certificates

This assumes that the frontend should present the same client
cert to both the search and metadata services, which is consistent
with the idea that a client cert should uniquely identify the
service that is presenting it.

If the environment variables are not set, then the clients are
created as normal without any expectation of mTLS.

This addresses https://github.com/amundsen-io/amundsen/issues/754

### Tests

I did not see any existing tests around the API utils, and I did not know what approach to take for them.

### Documentation

There are Python docstrings on the fields added to the config.

### CheckList

### DIFF
--- a/amundsen_application/api/utils/request_utils.py
+++ b/amundsen_application/api/utils/request_utils.py
@@ -105,7 +105,7 @@ def request_wrapper(method: str, url: str, client, headers, timeout_sec: int, da
         else:
             raise Exception('Method not allowed: {}'.format(method))
     else:
-        with requests.Session() as s:
+        with build_session() as s:
             if method == 'DELETE':
                 return s.delete(url, headers=headers, timeout=timeout_sec)
             elif method == 'GET':
@@ -116,3 +116,14 @@ def request_wrapper(method: str, url: str, client, headers, timeout_sec: int, da
                 return s.put(url, headers=headers, timeout=timeout_sec, data=data)
             else:
                 raise Exception('Method not allowed: {}'.format(method))
+
+
+def build_session() -> requests.Session:
+    session = requests.Session()
+
+    cert = app.config.get('MTLS_CLIENT_CERT')
+    key = app.config.get('MTLS_CLIENT_KEY')
+    if cert is not None and key is not None:
+        session.cert = (cert, key)
+
+    return session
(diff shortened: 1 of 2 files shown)


feng-tao:
will take a look

amundsen_application/api/utils/request_utils.py
```diff
@@ -116,3 +116,14 @@ def request_wrapper(method: str, url: str, client, headers, timeout_sec: int, da
...
                 raise Exception('Method not allowed: {}'.format(method))
+
+
+def build_session() -> requests.Session:
+    session = requests.Session()
+
+    cert = app.config.get('MTLS_CLIENT_CERT')
+    key = app.config.get('MTLS_CLIENT_KEY')
+    if cert is not None and key is not None:
+        session.cert = (cert, key)
```
feng-tao, line 127: so this assume private key and the certificate are two different file with different paths. Looking at the example link, it seems that it could be the case that single file has both info?

feng-tao:
lgtm mostly with one question

amundsen_application/api/utils/request_utils.py
feng-tao: I did some search which two files seem pretty common

davcamer: Yes, it could be either two files or a single file in a different format. In our deployment we use the two files format. Even in an organization that normally uses the single file format the key could be reformatted to two files.

I couldn't think of a clear way to allow for either one file or two, and since I only needed the two file format I only built for two files. I thought if the requirement came up later to handle the one file format as well, it could be added then with input from a real use case.

feng-tao (approved):

amundsen_application/api/utils/request_utils.py
feng-tao: let's go with two files for now.

feng-tao:
thanks @davcamer

PR #770, opened 2020-11-20
chore: build: upgrade webpack from 4 to 5

dorianj:
ref: https://github.com/amundsen-io/amundsen/issues/794

### Summary of Changes

Upgrade Webpack from 4 -> 5. Upgrade a few webpack plugin packages to fix deprecation warnings.

There is still one deprecation warning with a plain build:
```
(node:68621) [DEP_WEBPACK_WATCH_WITHOUT_CALLBACK] DeprecationWarning: A 'callback' argument need to be provided to the 'webpack(options, callbac
k)' function when the 'watch' option is set. There is no way to handle the 'watch' option without a callback.
```

This is [fixed in the newest version of `html-webpack-plugin`](https://github.com/jantimon/html-webpack-plugin/issues/1523), but it's in alpha and it's a major release, so I suggest just waiting until that comes out and upgrading separately.

There are other errors revealed by `node --trace-deprecation`... I believe that they're all caused by our dependencies, so I suggest we just ignore those for now and allow the dependencies to catch up, and we upgrade later.

`react-bootstrap` is particularly showing its age, however, it is likely even more difficult to do itself than this upgrade. I've tracked it here: https://github.com/amundsen-io/amundsen/issues/818, it should be done separately

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/amundsen_application/static/package-lock.json
+++ b/amundsen_application/static/package-lock.json
@@ -29001,6 +29001,12 @@
         "vm-browserify": "^1.0.1"
       },
       "dependencies": {
+        "path-browserify": {
+          "version": "0.0.1",
+          "resolved": "https://registry.npmjs.org/path-browserify/-/path-browserify-0.0.1.tgz",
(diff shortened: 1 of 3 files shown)


Golodhros (approved):
LGTM

Thanks for this Dorian!

dorianj:
@Golodhros can you please merge? Thank you!

PR #782, opened 2020-11-24
build: change url of title lint config file

dorianj:
ref: https://github.com/amundsen-io/amundsenfrontendlibrary/pull/606#issuecomment-732406016

### Summary of Changes

Title Lint app isn't picking up the config file. 

### Tests

This PR is prefixed with `build: ` which is in our config but not in the default of the Title Lint app, so if this passes CR, it worked


dorianj:
cc @Golodhros it didn't work :(

PR #785, opened 2020-11-24
fix: Revert "fix: (until they update urllib3) rolling back to v 0.10.15 of responses"

allisonsuarez:
Reverts amundsen-io/amundsenfrontendlibrary#784


Golodhros (approved):
LGTM

PR #788, opened 2020-11-25
chore: print human-readable error if index.html isn't found

dorianj:
### Summary of Changes

Issues like https://github.com/amundsen-io/amundsen/issues/822 seem like something is going very wrong with the install, but very commonly it's just a step was missing. In that case, print a human-readable more-helpful error line.

We don't have jinja2 in our requirements -- it's from Flask, but that seems stable enough to be safe

### Tests

n/a

### Documentation

n/a

### CheckList

### DIFF
--- a/amundsen_application/api/__init__.py
+++ b/amundsen_application/api/__init__.py
@@ -2,11 +2,15 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from typing import Any, Tuple
+import logging
 
 from flask import Flask, render_template
+import jinja2
 import os
 
+
 ENVIRONMENT = os.getenv('APPLICATION_ENV', 'development')
+LOGGER = logging.getLogger(__name__)
 
 
 def init_routes(app: Flask) -> None:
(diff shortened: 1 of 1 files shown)


amundsen_application/api/__init__.py
```diff
@@ -16,7 +18,11 @@ def init_routes(app: Flask) -> None:
 
 
 def index(path: str) -> Any:
-    return render_template("index.html", env=ENVIRONMENT)  # pragma: no cover
+    try:
+        return render_template("index.html", env=ENVIRONMENT)  # pragma: no cover
+    except jinja2.exceptions.TemplateNotFound as e:
+        logging.error("index.html template not found, have you built the front-end JS (npm run build in static/?")
```
feng-tao, line 24: shouldn't use logging directly but do sth like `LOGGER = logging.getLogger(` and use the Logger obj

ttannis: @dorianj here's an example of what Tao mentioned https://github.com/amundsen-io/amundsenfrontendlibrary/blob/master/amundsen_application/api/v0.py#L17. 

I know we do have some older examples that are still not doing this correctly that you likely followed so we'll have to clean those up, but here's some quick explanation: https://docs.python.org/3/library/logging.html#logger-objects

dorianj: Thanks y'all! Updated.

@ttannis ironically enough in that example, the actual log line later doesn't actually use the `LOGGER` object 😂  -- I will open a separate PR to fix these instances.

ttannis (approved):

feng-tao (approved):

dorianj:
Rebased to fix CI, could someone please merge?

feng-tao:
thanks

PR #802, opened 2020-12-02
fix: in search results, change header based on content

dorianj:
### Summary of Changes

When searching, `ResourceListItem` uses per-resource content classes to sub in the content columns. Thus the header columns need to vary based on the content as well.

Solve that by having the parent of `ResourceListHeader` pass in the types of the current content, and have `ResourceListHeader` swap out labels based on `ResourceType`

It's a bit smelly to me to have a fan-out based on `ResourceType` here. Perhaps ideally `Resource` would have a `columnLabel` method, and `DashboardResource` would override that to show the correct label (same for other subclasses of `Resources`). however AFAICT `DashboardResource` is just an interface, not an object, so I don't know that that's possible. Definitely looking for a better way to do this, this certainly works and isn't horrific, but it is a bit fragile. I've defaulted to just show BADGES (as before) so that if a new resource type is added, a label will be shown by default. If there are multiple different resource types passed in, we'll just display a blank column header.

### Tests

I don't know how to test this behavior, but would be happy to add test if pointed to a somewhat-similar example.

### Documentation

### CheckList

### DIFF
--- a/amundsen_application/static/js/components/common/ResourceList/ResourceListHeader/index.tsx
+++ b/amundsen_application/static/js/components/common/ResourceList/ResourceListHeader/index.tsx
@@ -5,13 +5,32 @@ import * as React from 'react';
 
 import './styles.scss';
 
+import { ResourceType } from 'interfaces';
 import {
   RESOURCE_HEADER_TITLE,
   SOURCE_HEADER_TITLE,
   BADGES_HEADER_TITLE,
+  LAST_UPDATED_HEADER_TITLE,
 } from './constants';
 
-const ResourceListHeader: React.FC = () => {
+export interface ResourceListHeaderProps {
+  resourceTypes: ResourceType[];
+}
+
+const contentHeaderTitle = (type: ResourceType): string => {
+  switch (type) {
+    case ResourceType.dashboard:
+      return LAST_UPDATED_HEADER_TITLE;
+
+    default:
+      return BADGES_HEADER_TITLE;
+  }
+};
+const ResourceListHeader: React.FC<ResourceListHeaderProps> = ({
+  resourceTypes,
+}: ResourceListHeaderProps) => {
+  const contentHeader =
+    resourceTypes.length === 1 ? contentHeaderTitle(resourceTypes[0]) : '';
   return (
     <div className="resource-list-header">
       <span className="resource">
@@ -19,7 +38,7 @@ const ResourceListHeader: React.FC = () => {
       </span>
       <span className="source">{SOURCE_HEADER_TITLE}</span>
       <span className="badges">
-        <span className="badges-text">{BADGES_HEADER_TITLE}</span>
+        <span className="badges-text">{contentHeader}</span>
       </span>
     </div>
   );
--- a/amundsen_application/static/js/pages/SearchPage/index.tsx
+++ b/amundsen_application/static/js/pages/SearchPage/index.tsx
@@ -145,9 +145,13 @@ export class SearchPage extends React.Component<SearchPageProps> {
       );
     }
 
+    const uniqueResourceTypes = [
+      ...new Set(results.results.map(({ type }) => type)),
+    ];
+
     return (
       <div className="search-list-container">
-        <ResourceListHeader />
+        <ResourceListHeader resourceTypes={uniqueResourceTypes} />
         <PaginatedApiResourceList
           activePage={page_index}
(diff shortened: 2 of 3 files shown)


allisonsuarez (approved):
Looks good to me, thanks Dorian!

amundsen_application/static/js/components/common/ResourceList/ResourceListHeader/index.tsx
```diff
@@ -5,21 +5,40 @@ import * as React from 'react';
...
+    resourceTypes.length === 1 ? contentHeaderTitle(resourceTypes[0]) : '';
   return (
     <div className="resource-list-header">
       <span className="resource">
         <span className="resource-text">{RESOURCE_HEADER_TITLE}</span>
       </span>
       <span className="source">{SOURCE_HEADER_TITLE}</span>
       <span className="badges">
-        <span className="badges-text">{BADGES_HEADER_TITLE}</span>
+        <span className="badges-text">{contentHeader}</span>
```
Golodhros, line 41: It could be that we don't have a badge text but we still render this element, right?

amundsen_application/static/js/components/common/ResourceList/ResourceListHeader/index.tsx
```diff
@@ -5,21 +5,40 @@ import * as React from 'react';
...
+
+const contentHeaderTitle = (type: ResourceType): string => {
+  switch (type) {
+    case ResourceType.dashboard:
+      return LAST_UPDATED_HEADER_TITLE;
+
+    default:
+      return BADGES_HEADER_TITLE;
+  }
+};
```
Golodhros, line 28: Nit: I wouldn't use a switch here, as a simple if would be enough.

amundsen_application/static/js/pages/SearchPage/index.tsx
```diff
@@ -145,9 +145,13 @@ export class SearchPage extends React.Component<SearchPageProps> {
       );
     }
 
+    const uniqueResourceTypes = [
+      ...new Set(results.results.map((resource) => resource.type)),
```
Golodhros, line 149: Nit: you could destructure {type}

Golodhros (commented):
Looks good! Some minor comments.

amundsen_application/static/js/components/common/ResourceList/ResourceListHeader/index.tsx
dorianj: Correct; I was worried that the layout might get messed up if we exclude the entire tree, but I can change to do that?

amundsen_application/static/js/components/common/ResourceList/ResourceListHeader/index.tsx
dorianj: My thinking is that it would make it clearer where to add code if folks added new resource types in the future -- wdyt?

Golodhros: Here I default to YAGNI, if that's the case, we could introduce the switch. (Although I would always prefer a mapped object instead)

amundsen_application/static/js/components/common/ResourceList/ResourceListHeader/index.tsx
Golodhros: Good call, I don't know what would happen. If you could test it, that would we awesome, thanks!

amundsen_application/static/js/components/common/ResourceList/ResourceListHeader/index.tsx
dorianj: yeah that's fair, updated

amundsen_application/static/js/components/common/ResourceList/ResourceListHeader/index.tsx
dorianj: ah, it makes the middle column float on over to the right.
Screen Shot 2020-12-02 at 10 33 13 AM

dorianj:
If LGTY'all, could someone merge? thanks!

amundsen_application/static/js/components/common/ResourceList/ResourceListHeader/index.tsx
Golodhros: Yeah, let's leave it empty for now!

PR #814, opened 2020-12-08
fix: make last run state in dashboard page optional

feng-tao:
Signed-off-by: feng-tao <fengtao04@gmail.com>

### Summary of Changes

For mode dashboard, we have last run state, for other dashboard type, we don’t have. So in https://github.com/amundsen-io/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/pages/DashboardPage/index.tsx#L323, it will become none. then once it enters https://github.com/amundsen-io/amundsenfrontendlibrary/blob/f5149bcbfbc90013ef169095ed1148d3dc96e016/amundsen_application/static/js/components/common/ResourceStatusMarker/index.tsx#L44 , it throws exception and fails the whole page.

Make last run state optional which fixed the issue.

### CheckList

### DIFF
--- a/amundsen_application/static/.betterer.results
+++ b/amundsen_application/static/.betterer.results
@@ -2239,7 +2239,7 @@ exports[`eslint`] = {
       [247, 14, 7, "\'wrapper\' is already declared in the upper scope.", "990908086"],
       [250, 10, 11, "Property name \`chart_names\` must match one of the following formats: camelCase", "219563138"]
     ],
-    "js/pages/DashboardPage/index.tsx:181364056": [
+    "js/pages/DashboardPage/index.tsx:2101761880": [
       [82, 20, 16, "Must use destructuring props assignment", "1899951550"],
       [88, 47, 19, "Must use destructuring props assignment", "1779396464"],
       [89, 20, 16, "Must use destructuring props assignment", "1899951550"],
--- a/amundsen_application/static/.betterer.results
+++ b/amundsen_application/static/.betterer.results
@@ -1,5 +1,5 @@
 // BETTERER RESULTS V2.
-exports[`no shadow`] = {
+exports[`eslint`] = {
   value: `{
     "js/components/ColumnList/ColumnStats/columnStats.story.tsx:850643706": [
       [9, 0, 48, "\`./testDataBuilder\` import should occur before import of \`.\`", "3767205268"]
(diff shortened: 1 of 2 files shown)


jornh:
Great fix, having a full failing page isn't fun! This is actually ONE step towards resolving https://github.com/amundsen-io/amundsen/issues/544 which has some discussion input in https://github.com/amundsen-io/amundsen/issues/544#issuecomment-656364931 we should consider. I'll discuss that later over there because that's more related to the metadata service API.

So, while at it why not tackle more (all?) of https://github.com/amundsen-io/amundsen/issues/544 - at the least for the "last run" related bits. Here's some reasoning:

If there is no `last run state` in the backend for a given dashboard it's also reasonable to assume there is no `last run` timestamp, therefore no `last successful run` time. This means for dashboard products that doesn't support the "last run" concept the whole section below "tags" is just useless UI noise as @jonhehir pointed out.

Following this it looks like we could just wrap both of the "last run" sections in the `hasLastRunState` condition and we're good.

Golodhros (approved):
LGTM

feng-tao:
> amundsen-io/amundsen#544



> Great fix, having a full failing page isn't fun! This is actually ONE step towards resolving [amundsen-io/amundsen#544](https://github.com/amundsen-io/amundsen/issues/544) which has some discussion input in [amundsen-io/amundsen#544 (comment)](https://github.com/amundsen-io/amundsen/issues/544#issuecomment-656364931) we should consider. I'll discuss that later over there because that's more related to the metadata service API.
> 
> So, while at it why not tackle more (all?) of [amundsen-io/amundsen#544](https://github.com/amundsen-io/amundsen/issues/544) - at the least for the "last run" related bits. Here's some reasoning:
> 
> If there is no `last run state` in the backend for a given dashboard it's also reasonable to assume there is no `last run` timestamp, therefore no `last successful run` time. This means for dashboard products that doesn't support the "last run" concept the whole section below "tags" is just useless UI noise as @jonhehir pointed out.
> 
> Following this it looks like we could just wrap both of the "last run" sections in the `hasLastRunState` condition and we're good.

thanks @jornh for the issue. Make sense. Let me take a look and update the pr.

feng-tao:
@Golodhros @allisonsuarez @jornh pr updated a bit

feng-tao:
@jornh the section is hidden now. thanks for the pointer!

PR #854, opened 2020-12-30
chore: add dorian to codeowners

dorianj:
I mostly want to approve dependabot PRs right right now.

I'll start reviewing other code as well but would like a second set of 👀 before I start merging alone


Golodhros (approved):
Thanks!

PR #861, opened 2021-01-07
fix: Set sender when sending feedback

JacobSMoller:
Signed-off-by: JacobSMoller <scherffenberg91@gmail.com>
### Summary of Changes

Sets the sender to current user, same style as its done for notifications. 

### Tests

Don't see any use in testing if the flask functionality works. 

### CheckList

### DIFF
--- a/amundsen_application/api/mail/v0.py
+++ b/amundsen_application/api/mail/v0.py
@@ -53,7 +53,10 @@ def feedback() -> Response:
             'form_data': data
         }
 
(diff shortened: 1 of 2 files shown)


JacobSMoller:
In my mind it makes the most sense if setting the current user email is set at this point, instead of having to be handled in individual clients. 
The individual clients can then handle `email_type: feedback` and set a recipient that makes sense for individual use cases.

JacobSMoller:
Maybe it doesn't make sense. 
I will think some more on how to handle this.

PR #865, opened 2021-01-13
chore: monthly release action

allisonsuarez:
### Summary of Changes

Added a workflow to create a new release every month.

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,3 +26,9 @@ disallow_untyped_defs = True
 ignore_missing_imports = True
 strict_optional = True
 warn_no_return = True
+
+[semantic_release]
+version_variable = "amundsenfrontendlibrary/setup.py:__version__",
+branch = "master"
+upload_to_pypi = true
+upload_to_release = true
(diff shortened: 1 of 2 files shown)


Golodhros (approved):
LGTM

setup.cfg
```diff
@@ -26,3 +26,9 @@ disallow_untyped_defs = True
 ignore_missing_imports = True
 strict_optional = True
 warn_no_return = True
+
+[semantic_release]
+version_variable = "amundsenfrontendlibrary/setup.py:__version__",
```
feng-tao, line 31: nice, qq: how does the version get updated?

feng-tao (commented):
could we do both auto release and manual release at the same time?

setup.cfg
Golodhros: the semantic-release library would do it for us.

Golodhros:
> could we do both auto release and manual release at the same time?

Yeah, we were planning to support the manual release as current, by adding a tagged commit to the repo.

We also want to automatically create pre-releases with all merged PRs, still WIP

mgorsk1:
does this take care of changelog generation as well or will be done in separate mr ?

PR #957, opened 2021-03-18
feat: Add Table Notices

Golodhros:
### Summary of Changes
Adds Table and Dashboards notifications in the detail page using the current JavaScript configuration system

Implements https://github.com/amundsen-io/rfcs/pull/29

redshift_dimension_rides_-_Amundsen_Table_Details

### Tests
Tested configuration helper

### Documentation
Not yet


mgorsk1:
Cool to see that one make it through ! We've been meaning to work on something similar next months.

Regarding questions I think I might have input for 1 and 2 - our use case assumed not only alerts but also warnings and info messages. Examples information we wanted to show were:
```
1. This is an official table, access it using view Y (info)
2. This table was not used in last 90 days and might be deprecated (warning)
3. To jest stara wersja tabeli. Nowa wersja dostepna pod adresem Y. (tak moglibysmy oznaczyc wzsystkie tabele ze starego ingestion) (alert - like in your example)
4. Other users has also viewed tables X, Y, Z (based on recommendation engine output)
```
so it definitely would be more scalable with more than just one type of notification in my opinion.

One nit - would it make sense to have alerts section `under` the description ?

Golodhros:
> Cool to see that one make it through ! We've been meaning to work on something similar next months.
> 
> Regarding questions I think I might have input for 1 and 2 - our use case assumed not only alerts but also warnings and info messages. Examples information we wanted to show were:
> 
> ```
> 1. This is an official table, access it using view X (info)
> 2. This table was not used in last 90 days and might be deprecated (warning)
> 3. There is a new version of this table named Y (alert - like in your example)
> 4. Other users has also viewed tables X, Y, Z (based on recommendation engine output) (info)
> ```
> 
> so it definitely would be more scalable with more than just one type of notification in my opinion (and possibility to define more than one notification per table)
> 
> One nit - would it make sense to have alerts section `under` the description ?

That's awesome, thanks for chiming in @mgorsk1!

I could easily add the three types of notification part, it think it would be something like:
```
'schema.name': {
  severity: NotificationSeverity.ALERT,
  message: 'Lorem ipsum....'
},
'schema.name2': {
  severity: NotificationSeverity.INFO/WARNING,
  message: ...
}]
```

Regarding the positioning, let me ask our designer to see what she thinks. I guess the placement was because they wanted to give it the most importance.

What do you think about the RFC, should I open one?

dorianj:
This is cool! I haven't looked closely, but very much want this functionality. 

Minor nit: is there something else we could name this? Notifications to me implies stuff getting pushed to people. Perhaps table notices, alerts, or something like that?

+1 on `three types of notification part` -- I think having severity classes is useful

I think that it would be reasonable to have multiple notifications per table?


regarding RFC, I think this does qualify, it is a new feature. The biggest discussion points I see are:
1. whether this should live in the graph, or configuration? (or possibly a mix -- might be reasonable to have a design where 
1. should the configuration, or graph, allow for wildcard associations? Some people might want to add notices for all tables in a certain namespace, for example
1. whether this should be associable with all object types, or just tables?

it is quite probable the initial implementation wouldn't do the full scope, but the initial impl details may change based on some of these answers

Golodhros:
> This is cool! I haven't looked closely, but very much want this functionality.
> 
> Minor nit: is there something else we could name this? Notifications to me implies stuff getting pushed to people. Perhaps table notices, alerts, or something like that?

Makes sense, changing it to "Notices"

> +1 on `three types of notification part` -- I think having severity classes is useful

Updated this PR with it

> I think that it would be reasonable to have multiple notifications per table?

This one is a bit more problematic, as we would need to limit them and maybe offer the option to close them?

> regarding RFC, I think this does qualify, it is a new feature. The biggest discussion points I see are:

I will create an RFC now.

> 1. whether this should live in the graph, or configuration? (or possibly a mix -- might be reasonable to have a design where
> 2. should the configuration, or graph, allow for wildcard associations? Some people might want to add notices for all tables in a certain namespace, for example
> 3. whether this should be associable with all object types, or just tables?
> 
> it is quite probable the initial implementation wouldn't do the full scope, but the initial impl details may change based on some of these answers

All legit questions.
I will add the association to Dashboards and People too. The others I think they are interesting though.

amundsen_application/static/js/pages/DashboardPage/index.tsx
```diff
@@ -121,52 +129,54 @@ export class DashboardPage extends React.Component<
...
   render() {
-    const { dashboard, isLoading } = this.props;
+    const { dashboard, isLoading, statusCode } = this.props;
     const hasDescription =
       dashboard.description && dashboard.description.length > 0;
     const hasLastRunState =
       dashboard.last_run_state && dashboard.last_run_state.length > 0;
+    const dashboardNotice = getResourceNotices(
+      ResourceType.dashboard,
+      `${dashboard.product}.${dashboard.name}`
```
danwom, line 173: I don't think the product and dashboard name are enough to uniquely identify a dashboard. Multiple dashboards in the same product are named some generic "payments dashboard" or something that would all match the same config value. The "uri" ("key" for tables, we should standardize to one or the other but that's a different issue) for dashboards is `{product}://{cluster}.{group}/{name}`.

docs/application_config.md
```diff
@@ -128,6 +131,51 @@ A table resource may have a source of table and column description attached to i
...
+    [ResourceType.table]: {
+      ... //Table Resource Configuration
+      notices: {
+          "<SCHEMANAME>.<TABLENAME>": {
+            severity: NoticeSeverity.ALERT,
+            message: `<span>This table is deprecated, please use <a href="<LINKTONEWTABLEDETAILPAGE>">this new table</a> instead.</span>`,
+          },
+      },
+    },
+  },
```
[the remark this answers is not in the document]
Golodhros: > A few notes about this configuration:
> 
> * I think this furthers the need to move our configs to an external config service. Updating these messages will require a rebuild and redeploy which is not ideal.

100% Agree, I think a combination of a feature flag system plus an admin UI would be ideal.

> * Using the `"<SCHEMANAME>.<TABLENAME>"` as the table identifier seems somewhat nonstandard. Everywhere we identify a table we always use the pattern `{database}://{cluster}.{schema}/{table_name}`. Is the intent here to match similar tables in different databases or clusters?

Oh, didn't think about this. I will try to include the rest.

> * nit: rename `message` to `messageHtml`. Also is it required to wrap the message in a span?

Good call, I that key name will be more explicit.
I will check about the span, I can't remember why it was there.

amundsen_application/static/js/components/Alert/index.tsx
```diff
@@ -20,6 +27,7 @@ export interface AlertProps {
 
 const Alert: React.FC<AlertProps> = ({
   message,
+  severity = NoticeSeverity.WARNING,
```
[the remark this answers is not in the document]
Golodhros: I actually thought about it all the way around. I put the default to the most important one, the warning. I think the others are nice to have, but the big deal is creating warnings, that's where most of the value of the feature comes from.

For now we haven't contemplated no icon. IMO that would remove the visual strength so we could do that kind of stuff with a programmatic description.

amundsen_application/static/js/pages/DashboardPage/index.tsx
Golodhros: Good call, I will try to add more stuff here.

danwom (commented):
Aside from changing the key/uri format, should be good to merge.

amundsen_application/static/js/pages/DashboardPage/index.tsx
```diff
@@ -121,52 +129,54 @@ export class DashboardPage extends React.Component<
...
   render() {
-    const { dashboard, isLoading } = this.props;
+    const { dashboard, isLoading, statusCode } = this.props;
     const hasDescription =
       dashboard.description && dashboard.description.length > 0;
     const hasLastRunState =
       dashboard.last_run_state && dashboard.last_run_state.length > 0;
+    const dashboardNotice = getResourceNotices(
+      ResourceType.dashboard,
+      `${dashboard.product}.${dashboard.cluster}.${dashboard.group_name}.${dashboard.name}`
```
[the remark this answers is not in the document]
Golodhros: I just didn't like the idea of putting a string with `//:` on it, it feels like using a URL and can be confusing.

danwom (approved):

amundsen_application/static/js/components/Alert/index.tsx
```diff
@@ -2,16 +2,23 @@
...
 import { IconSizes } from 'interfaces';
-import { AlertIcon } from '../SVGIcons';
+import { NoticeSeverity } from 'config/config-types';
+import { AlertIcon, InformationIcon } from '../SVGIcons';
 
 import './styles.scss';
 
-const STROKE_COLOR = '#b8072c'; // $red70
+const SEVERITY_TO_COLOR_MAP = {
+  [NoticeSeverity.INFO]: '#3a97d3', // cyan50
```
allisonsuarez, line 14: We don't have these colors on the _colors css file as variables?

allisonsuarez (approved):
LGTM with one small comment

amundsen_application/static/js/components/Alert/index.tsx
Golodhros: Yeah, that's the 'cyan50', but we cannot bring the scss variables into the JS files :(

PR #958, opened 2021-03-25
bug: remove duplicated env var fetch

dorianj:
In `wsgi.py`, we do:
```
application = create_app(
    config_module_class=os.getenv('FRONTEND_SVC_CONFIG_MODULE_CLASS') or 'amundsen_application.config.LocalConfig')
```

So this check of env vars inside `create_app` was duplicated. This is problematic if you use a custom `create_app` invocation, and then accidentally have a `FRONTEND_SVC_CONFIG_MODULE_CLASS` env var present. It might be reasonable to have `create_app` use the env var iff `config_module_class` isn't set (i'm happy to change PR to do that), which I think is more in line with expectations of default vars.

### Tests

Ran frontend with custom `FRONTEND_SVC_CONFIG_MODULE_CLASS` set, it's picked up.

(ty for @madison-ookla for noticing this irregularity)

### DIFF
--- a/amundsen_application/__init__.py
+++ b/amundsen_application/__init__.py
@@ -36,7 +36,7 @@ STATIC_ROOT = os.getenv('STATIC_ROOT', 'static')
 static_dir = os.path.join(PROJECT_ROOT, STATIC_ROOT)
 
 
-def create_app(config_module_class: str, template_folder: str = None) -> Flask:
+def create_app(config_module_class: str = None, template_folder: str = None) -> Flask:
     """ Support for importing arguments for a subclass of flask.Flask """
     args = ast.literal_eval(os.getenv('APP_WRAPPER_ARGS', '')) if os.getenv('APP_WRAPPER_ARGS') else {}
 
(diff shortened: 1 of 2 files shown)


amundsen_application/__init__.py
```diff
@@ -43,10 +43,7 @@ def create_app(config_module_class: str, template_folder: str = None) -> Flask:
     tmpl_dir = template_folder if template_folder else os.path.join(PROJECT_ROOT, static_dir, 'dist/templates')
     app = app_wrapper_class(__name__, static_folder=static_dir, template_folder=tmpl_dir, **args)
 
-    """ Support for importing a custom config class """
-    config_module_class = \
-        os.getenv('FRONTEND_SVC_CONFIG_MODULE_CLASS') or config_module_class
```
feng-tao, line 48: maybe just put `os.getenv('FRONTEND_SVC_CONFIG_MODULE_CLASS')` for the method's config_module_class default input value

dorianj: @feng-tao Updated to use this style. There are two caveats:
* Can't use default input values, since that's executed at definition time, not per-invocation
* The call-site still needs to check the env var, to pass in the fallback static default. (this could be moved into the function itself, but didn't feel right to have `amundsen_application.config.TestConfig` in `create_app`, open to feedback of course).

feng-tao (approved):

PR #960, opened 2021-03-29
feat: Allows for splitting stats' distinct values into a different element that shows in modal

Golodhros:
### Summary of Changes
This is part of the work around table stats
Splits distinct values from the regular stats table
Consolidates stats utils in one file
Adds a configuration option to add the distinct values stat name to the configuration

### Tests
Added tests for new components
Tested utils

## Screenshots
coco_dim_driver_-_Amundsen_Table_Details

coco_dim_driver_-_Amundsen_Table_Details

### Documentation
Soon

### CheckList


allisonsuarez:
Code looks good to me but referencing back to the design isn't the modal supposed to be less wide?

Golodhros:
> Code looks good to me but referencing back to the design isn't the modal supposed to be less wide?

Yeah, I changed it already, that's an old screenshot.

amundsen_application/static/js/features/ColumnList/ColumnStats/index.tsx
```diff
@@ -41,8 +42,9 @@ const ColumnStats: React.FC<ColumnStatsProps> = ({
   if (stats.length === 0) {
     return null;
   }
-  const startEpoch = Math.min(...stats.map(getStart));
-  const endEpoch = Math.max(...stats.map(getEnd));
+  const filteredStats = filterOutUniqueValues(stats);
```
danwom, line 45: Just a nit about readability: We're using the both distinct (`distinctStatTypeName`) and unique (`uniqueValues`) somewhat interchangeably. I think it would be better to use either `distinct` or `unique` consistently for both the configs and util functions.

amundsen_application/static/js/features/ColumnList/index.tsx
```diff
@@ -161,6 +163,7 @@ const ExpandedRowComponent: React.FC<ExpandedRowProps> = (
 
     return true;
   };
+  const uniqueValues = rowValue.stats && getUniqueValues(rowValue.stats);
```
danwom, line 166: I would suggest something like:

```
const stats = rowValue.stats;
const normalStats = filterOutUniqueValues(stats);
const uniqueValueStats = getUniqueValues(stats);
```

Then we can remove the changes in the `ColumnStats` component and send `normalStats` instead. This moves the logic of splitting out the stats that need special processing and the normal stats all in one location that makes it easier to follow.

danwom (commented):
Overall looks pretty good. Left some comments to help with readability.

amundsen_application/static/js/features/ColumnList/ColumnStats/index.tsx
Golodhros: I was using the unique values name until I found they call them distinct on the backend.

you are right, I will rename the configuration.

amundsen_application/static/js/features/ColumnList/index.tsx
Golodhros: That makes sense, thanks!

PR #963, opened 2021-03-30
chore: upgrade to marshmallow 3

dkunitsk:
This rolls frontend onto marshmallow 3. After this, only gremlin is left to upgrade.

Disclaimer: Marshmallow3 is more strict than its predecessor (e.g. by default does not accept unknown keys) and unless we override that globally - which I do not recommend - there may be API differences that go uncaught in testing. Unlike metadata/search, frontend makes more use of marshmallow so there is more potential for problems. 

### Testing
Have started up frontend and clicked around to confirm no obvious problems. 

### CheckList

### DIFF
--- a/amundsen_application/api/preview/v0.py
+++ b/amundsen_application/api/preview/v0.py
@@ -9,6 +9,7 @@ from http import HTTPStatus
 
 from flask import Response, jsonify, make_response, request, current_app as app
 from flask.blueprints import Blueprint
+from marshmallow import EXCLUDE, ValidationError
 from werkzeug.utils import import_string
 
 from amundsen_application.models.preview_data import PreviewDataSchema
@@ -49,11 +50,11 @@ def get_table_preview() -> Response:
         preview_data = json.loads(response.data).get('preview_data')
         if status_code == HTTPStatus.OK:
             # validate the returned table preview data
-            data, errors = PreviewDataSchema().load(preview_data)
-            if not errors:
+            try:
+                data = PreviewDataSchema().load(preview_data)
                 payload = jsonify({'previewData': data, 'msg': 'Success'})
-            else:
-                logging.error('Preview data dump returned errors: ' + str(errors))
+            except ValidationError as err:
+                logging.error('Preview data dump returned errors: ' + str(err.messages))
                 raise Exception('The preview client did not return a valid PreviewData object')
         else:
             message = 'Encountered error: Preview client request failed with code ' + str(status_code)
--- a/amundsen_application/api/preview/v0.py
+++ b/amundsen_application/api/preview/v0.py
@@ -9,7 +9,7 @@ from http import HTTPStatus
 
 from flask import Response, jsonify, make_response, request, current_app as app
 from flask.blueprints import Blueprint
-from marshmallow import EXCLUDE, ValidationError
+from marshmallow import ValidationError
 from werkzeug.utils import import_string
 
 from amundsen_application.models.preview_data import PreviewDataSchema
--- a/amundsen_application/api/utils/metadata_utils.py
+++ b/amundsen_application/api/utils/metadata_utils.py
@@ -5,6 +5,7 @@ import logging
 import urllib.parse
 
 from dataclasses import dataclass
+from marshmallow import EXCLUDE
 from typing import Any, Dict, List
 
 from amundsen_common.models.dashboard import DashboardSummary, DashboardSummarySchema
@@ -46,10 +47,9 @@ def marshall_table_partial(table_dict: Dict) -> Dict:
 
     TODO - Unify data format returned by search and metadata.
     """
-    schema = PopularTableSchema(strict=True)
-    # TODO: consider migrating to validate() instead of roundtripping
-    table: PopularTable = schema.load(table_dict).data
-    results = schema.dump(table).data
+    schema = PopularTableSchema()
+    table: PopularTable = schema.load(table_dict, unknown=EXCLUDE)
+    results = schema.dump(table)
     # TODO: fix popular tables to provide these? remove if we're not using them?
     # TODO: Add the 'key' or 'id' to the base PopularTableSchema
     results['key'] = f'{table.database}://{table.cluster}.{table.schema}/{table.name}'
@@ -104,10 +104,9 @@ def marshall_table_full(table_dict: Dict) -> Dict:
     :return: Table Dict with sanitized fields
     """
 
-    schema = TableSchema(strict=True)
-    # TODO: consider migrating to validate() instead of roundtripping
-    table: Table = schema.load(table_dict).data
-    results: Dict[str, Any] = schema.dump(table).data
+    schema = TableSchema()
+    table: Table = schema.load(table_dict)
+    results: Dict[str, Any] = schema.dump(table)
 
     is_editable = is_table_editable(results['schema'], results['name'])
     results['is_editable'] = is_editable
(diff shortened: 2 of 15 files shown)


amundsen_application/api/utils/metadata_utils.py
```diff
@@ -46,10 +47,9 @@ def marshall_table_partial(table_dict: Dict) -> Dict:
 
     TODO - Unify data format returned by search and metadata.
     """
-    schema = PopularTableSchema(strict=True)
-    # TODO: consider migrating to validate() instead of roundtripping
-    table: PopularTable = schema.load(table_dict).data
-    results = schema.dump(table).data
+    schema = PopularTableSchema()
+    table: PopularTable = schema.load(table_dict, unknown=EXCLUDE)
```
dkunitsk, line 51: the lines below insert `key` into the result, however some test cases have `key` as part of table_dict. Since `PopularTable` itself is anyway deprecated, making `load` more lenient made sense.

amundsen_application/api/utils/metadata_utils.py
```diff
@@ -149,9 +148,9 @@ def marshall_dashboard_partial(dashboard_dict: Dict) -> Dict:
     :param dashboard_dict: Dict of partial dashboard metadata
     :return: partial dashboard Dict
     """
-    schema = DashboardSummarySchema(strict=True)
-    dashboard: DashboardSummary = schema.load(dashboard_dict).data
-    results = schema.dump(dashboard).data
+    schema = DashboardSummarySchema(unknown=EXCLUDE)
```
dkunitsk, line 151: a test contains `total_usage` which is not on the schema. Should the schema (which is in `common`) be updated?

amundsen_application/base/base_announcement_client.py
```diff
@@ -31,20 +32,16 @@ def _create_error_response(message: str) -> Response:
             return make_response(payload, HTTPStatus.INTERNAL_SERVER_ERROR)
 
         try:
```
dkunitsk, line 34: i removed this outer try/except. There's nothing it catches which is not caught by the inner try/except.

amundsen_application/base/base_announcement_client.py
```diff
@@ -31,20 +32,16 @@ def _create_error_response(message: str) -> Response:
...
-                message = 'Announcement data dump returned errors: ' + str(errors)
-                return _create_error_response(message)
+            announcements = self.get_posts()
         except Exception as e:
-            message = 'Encountered exception: ' + str(e)
+            message = 'Encountered exception getting posts: ' + str(e)
+            return _create_error_response(message)
+
+        try:
+            data = AnnouncementsSchema().dump(announcements)
```
dkunitsk, line 41: this dump-then-load pattern is ugly, but I think it may be the best way to keep the functionality unchanged. See https://marshmallow.readthedocs.io/en/stable/upgrading.html#validation-does-not-occur-on-serialization. 

In general, marshmallow doesn't seem to understand the idea of validating an app object. `load` or `validate` only work on dictionaries. This seems odd to me. If anyone knows more than me, please LMK.

amundsen_application/models/announcements.py
```diff
@@ -29,10 +29,11 @@ class AnnouncementsSchema(Schema):
...
     @post_dump
-    def validate_data(self, data: Dict) -> None:
+    def validate_data(self, data: Dict, **kwargs) -> None:
         posts = data.get('posts', [])
         for post in posts:
             if post.get('date') is None:
                 raise ValidationError('All posts must have a date')
             if post.get('title') is None:
                 raise ValidationError('All posts must have a title')
+        return data
```
dkunitsk, line 39: https://marshmallow.readthedocs.io/en/stable/upgrading.html#pre-post-processors-must-return-modified-data

amundsen_application/models/preview_data.py
```diff
@@ -24,6 +24,6 @@ def __init__(self, columns: List = [], data: List = [], error_text: str = '') ->
 
 
 class PreviewDataSchema(Schema):
-    columns = fields.Nested(ColumnItemSchema, many=True)
+    columns = fields.Nested(ColumnItemSchema, many=True, unknown=EXCLUDE)
```
dkunitsk, line 27: `PreviewTest.test_good_client_response fails` because it passes many unknown fields (see https://github.com/amundsen-io/amundsenfrontendlibrary/blob/c6321020813e17fffef33a8b9a03d307233082a4/tests/unit/base/test_superset_preview_client.py#L18). 

Is this a test problem or should this schema handle that?

amundsen_application/models/user.py
```diff
@@ -28,16 +28,11 @@ def load_user(user_data: Dict) -> User:
...
 
 
 def dump_user(user: User) -> Dict:
     schema = UserSchema()
-    try:
-        data, errors = schema.dump(user)
-        return data
-    except ValidationError as err:
-        return err.messages
+    return schema.dump(user)
```
dkunitsk, line 38: dump would have never raised `ValidationError` previously. It would have returned errors. So the try/catch was not doing anything.

requirements.txt
```diff
@@ -56,12 +56,7 @@ requests==2.25.1
 # A lightweight library for converting complex datatypes to and from native Python datatypes.
 # License: MIT
 # Upstream url: https://github.com/marshmallow-code/marshmallow
-marshmallow>=2.15.3,<3.0
-
-# Allows declaring marshmallow schema through type annotations
-# License: MIT
-# Upstream url: https://github.com/justanr/marshmallow-annotations
-marshmallow-annotations>=2.4.0,<3.0
```
dkunitsk, line 64: we do not use marshmallow-annotations directly in this service

setup.py
```diff
@@ -37,11 +37,11 @@ def build_js() -> None:
 with open(requirements_path) as requirements_file:
     requirements = requirements_file.readlines()
 
-__version__ = '3.5.1'
+__version__ = '3.6.0'
 
 oicd = ['flaskoidc==0.1.1']
 pyarrrow = ['pyarrow==3.0.0']
-bigquery_preview = ['google-cloud-bigquery>=2.8.0,<3.0.0', 'flatten-dict==0.3.0']
+bigquery_preview = ['google-cloud-bigquery>=2.13.1,<3.0.0', 'flatten-dict==0.3.0']
```
dkunitsk, line 44: we import this library as `google.cloud`. The `google` package does not exist in v2.8.0. I don't know when exactly it was added but 2.13.1 (latest) has it.

tests/unit/base/test_announcement_client.py
```diff
@@ -75,7 +75,7 @@ def test_get_posts_success(self) -> None:
                 {
                     'title': 'Test Title',
                     'date': 'December 31, 1999',
-                    'info_list': ['Test announcement'],
+                    'html_content': 'content',
```
dkunitsk, line 78: this is a typo. There is no `info_list` anywhere. I think it passed before because even the old implementation of `dump` did not validate objects.

amundsen_application/api/utils/metadata_utils.py
allisonsuarez: it doesn't look like that field ever existed in common moel in the first place, weird

amundsen_application/api/preview/v0.py
```diff
@@ -9,6 +9,7 @@
 
 from flask import Response, jsonify, make_response, request, current_app as app
 from flask.blueprints import Blueprint
+from marshmallow import EXCLUDE, ValidationError
```
allisonsuarez, line 12: looks like EXCLUDE is never used here

dorianj (commented):
Don't have specific comments on the stickier bits where test cases were previously incorrectly passing, but the rest LGTM

amundsen_application/api/utils/metadata_utils.py
feng-tao: could you help to comment on `PopularTable itself is anyway deprecated` ? Do we plan to replace the popular table section with the full table object?

amundsen_application/api/utils/metadata_utils.py
feng-tao: maybe the model is outdated, cc @jinhyukchang in case Jin remembered

feng-tao (approved):
lgtm

mgorsk1:
nothing more to add in terms of the code - lgtm. just thinking out loud but some kind of integration test in amundsen repo would be useful for these kind of changes - we could easily spot issues with dependency issues, model incompatibilities, etc.

amundsen_application/api/utils/metadata_utils.py
dkunitsk: I only based this off the comment in https://github.com/amundsen-io/amundsencommon/blob/master/amundsen_common/models/popular_table.py.

PR #968, opened 2021-04-07
fix: Properly render html markdown for programmatic descriptions

slavierouse:
### Summary of Changes
Programmatic descriptions were rendering raw html instead of rendered html. This PR makes sure markdown with html tags like  renders the html, in this example the image. For example, at Lyft, we use programmatic descriptions with  to 
Before fix
image

After fix
image

@danwom and @allisonsuarez confirm there is no possibility of malicious injection through this component

### Tests
No tests modified - this aspect of programmatic descriptions was not covered in tests

### Documentation
None

### DIFF
--- a/amundsen_application/static/js/components/EditableText/index.tsx
+++ b/amundsen_application/static/js/components/EditableText/index.tsx
@@ -125,7 +125,7 @@ class EditableText extends React.Component<
       return (
         <div className="editable-text">
           <div className="markdown-wrapper">
-            <ReactMarkdown allowDangerousHtml={true}>{value}</ReactMarkdown>
+            <ReactMarkdown allowDangerousHtml>{value}</ReactMarkdown>
           </div>
           {editable && !value && (
             <a
--- a/amundsen_application/static/js/components/EditableText/index.tsx
+++ b/amundsen_application/static/js/components/EditableText/index.tsx
@@ -125,7 +125,7 @@ class EditableText extends React.Component<
       return (
         <div className="editable-text">
           <div className="markdown-wrapper">
-            <ReactMarkdown allowDangerousHtml={false}>{value}</ReactMarkdown>
+            <ReactMarkdown allowDangerousHtml={true}>{value}</ReactMarkdown>
           </div>
           {editable && !value && (
             <a


allisonsuarez (approved):
LGTM, the option name allowDangerousHtml makes me want to scream hahaha

slavierouse:
Force push to fix signoff

dorianj:
By definition there is possibility of malicious injection via this component.

For programmatic definitions, that's arguably OK because the integration controls what goes in. But I believe this covers all editable text fields, meaning an end user could edit a normal description and add HTML that gets rendered. I don't think that's desirable. WDYT?

For that particular example, Markdown supports images, so why not use Markdown images?

Golodhros:
> By definition there is possibility of malicious injection via this component.
> 
> For programmatic definitions, that's arguably OK because the integration controls what goes in. But I believe this covers all editable text fields, meaning an end user could edit a normal description and add HTML that gets rendered. I don't think that's desirable. WDYT?

Oops! You are right!

> For that particular example, Markdown supports images, so why not use Markdown images?
We could try that.

Another way would be to add a isSafe boolean prop to `EditableText` to cover this one.

Golodhros (changes requested):
It won't be safe, let's think it again.

allisonsuarez:
> It won't be safe, let's think it again.

Why wouldn't it be safe if there is no input for injecting HTML here? My understanding is that users have no way of actually altering the HTML so the only way in which this could be compromising is if the programatic description object from metadata is somehow malicious. Maybe I'm not interpreting what "safe" means in this context.

slavierouse:
@dorianj @Golodhros how about we sanitize the html before rendering?

dorianj:
I'm open to proposals, but in my experience sanitizing HTML securely is a very, very long and challenging road to go down. It only takes one miss to break it open, and the combos of html/browsers is very complicated (something that looks like completely invalid html might parse on some browsers). Perhaps there's a widely accepted library that does a good job of this, if that's the case, it could work.

slavierouse:
@dorianj @Golodhros this library is pretty accepted for sanitizing html: https://www.npmjs.com/package/sanitize-html there's a react wrapper around it which is only minorly different: https://www.npmjs.com/package/sanitize-html-react. I've used this before and it's pretty straightforward to implement. I think the bigger question is design: based on the docs sanitizing every user input before it's stored in the database on the backend is the best practice, sanitizing both on store and on serve from backend is like even more secure, sanitizing on the front end display only may be considered fickle.

Adding a "canBeUnsafe" prop to demark programmatic descriptions that can be assumed to be safe like Marcos suggested may be simpler, but if someone somewhere somehow implements programmatic descriptions in a way that a malicious actor can inject, it's probably not totally safe.

Also checking with the Lyft crew if we can just change our description

dorianj:
Seems like it has a fairly [good track record](https://snyk.io/vuln/npm:sanitize-html), all things considered... I'm down to use it

if we do just for programmatic descriptions, could use the sanitizer on render only?

if we do end up wanting to do it for all editables, we should sanitize in and out, as you mentioned

dorianj:
Please re-open against monorepo.

